PR #65, opened 2015-02-01
Added '-y' flag to the command 'add-apt-repository'. 

jon-whit:
I noticed this flag was missing from a few of the commands. Adding this argument will allow building without interactivity.


zvonimir:
Good catch!

Could you please double-check other scripts there for the same issues.
Thanks!


--
http://zvonimir.info
http://soarlab.org/

PR #75, opened 2015-03-06
Initialization Function Support

snoopsmsc:
Allows for initialization code to be inserted into main immediately after $static_init().

During function iteration, SmackModuleGenerator now looks for function definitions starting with "__SMACK_init_func_", and calls each such function from an additional generated function called $init_funcs().

After function iteration is finishes, the completed $init_funcs() is added to the boogie program.

The impetus for this change is to be able to add initialization code to main from c header files (specifically, for pthreads support, I need the ability to initialize an array that tracks the status of threads so they can't start in a "running" state).


snoopsmsc:
A small note, for the new test init_funcs_fail.c, boogie is returning "UNKNOWN" for all three memory models.  However, if you look at the output from "smackverify.py init_funcs_fail.c --verifier boogie", it seems like it is indicating failure.

I'm not totally up to speed on the new regtest infrastructure, so I thought I would mention this since they aren't coming out green.

lib/smack/SmackModuleGenerator.cpp
```diff
@@ -67,8 +67,13 @@ void SmackModuleGenerator::generateProgram(llvm::Module& m) {
...
 
       // First execute static initializers, in the main procedure.
-      if (naming.get(*func) == "main" && rep.hasStaticInits())
-        proc->insert(Stmt::call(SmackRep::STATIC_INIT));
+      if (naming.get(*func) == "main") {
+        proc->insert(Stmt::call(SmackRep::INIT_FUNCS));
+        if(rep.hasStaticInits())
+          proc->insert(Stmt::call(SmackRep::STATIC_INIT));
+      } else if (naming.get(*func).substr(0, 18)  == "__SMACK_init_func_")
+        // TODO? Check for no arguments?
```
zvonimir: I think you should add an assertion here that checks that init functions take no arguments and return no values. We should probably enforce this since I cannot imagine that arguments and return values would be needed.

test/basic/init_funcs.c
```diff
@@ -0,0 +1,16 @@
+#include "smack.h"
```
zvonimir: I think it would be good to have a slightly more realistic example. So maybe introduce two global variables, which get set to something in the initialization functions. Then check whether they were indeed set. I think that would make for a slightly more realistic example.

include/smack/SmackRep.h
```diff
@@ -184,8 +186,10 @@ class SmackRep {
   virtual vector<string> getModifies();
   unsigned numElements(const llvm::Constant* v);
   void addInit(unsigned region, const llvm::Value* addr, const llvm::Constant* val);
+  void addInitFunc(const Stmt* s);
```
zvonimir: I would change the signature of this so that it takes "llvm::Function*" as its parameter. I think that would be a slightly better encapsulation.

test/basic/init_funcs.c
```diff
@@ -0,0 +1,16 @@
...
+
+// @expect verified
+
+void __SMACK_init_func_test() {
```
zvonimir: Hey, here is an idea, and this might look nicer.
Could you introduce a macro called __SMACK_INIT in smack.h that prefixes whatever it is applied to with __SMACK_init_func_?
So then this function declaration would look like this:
void __SMACK_INIT(test)() {
....
}
You can play with this a little bit...
Another option would be to just have
__SMACK_INIT(test) {
...
}
and then expand it into
void __SMACK_init_func() {
...
}

test/regtest.py
```diff
@@ -78,7 +79,9 @@ def metadata(file):
 passed = failed = timeouts = unknowns = 0
 
 try:
-  for test in glob.glob("./**/*.c"):
+  #Allow for specific (or expanded wildcard) tests to be selected on command line
```
zvonimir: Please remove these changes to regtest.py for now....

test/basic/init_funcs.c
snoopsmsc: Wait, should:

void __SMACK_INIT(test)(){
...
}

be expanded to 

void __SMACK_init_func() {
...
}

or 

void __SMACK_init_funct_test() {
...
}

?

I assume you mean the latter?

zvonimir: The latter, yes...

include/smack/SmackRep.h
snoopsmsc: These are actually function calls, not function declaration/definitions.  Does an llvm::Function refer to a function call?

The backing data structure for this (vector<Stmt*> initFuncs) contains a list of call statements that get injected into the beginning of main().

Am I missing something here?

zvonimir: I was just saying that the signature of addInitFunc should be:
void addInitFunc(const llvm::Function* func);
Then this Stmt::call(naming.get(*func)) should be encapsulated in this function, as opposed to being done in SmackModuleGenerator.

snoopsmsc: Ah!  Gotcha.  Now it's making sense.

lib/smack/SmackModuleGenerator.cpp
```diff
@@ -67,8 +67,15 @@ void SmackModuleGenerator::generateProgram(llvm::Module& m) {
...
 
       // First execute static initializers, in the main procedure.
-      if (naming.get(*func) == "main" && rep.hasStaticInits())
-        proc->insert(Stmt::call(SmackRep::STATIC_INIT));
+      if (naming.get(*func) == "main") {
+        proc->insert(Stmt::call(SmackRep::INIT_FUNCS));
+        if(rep.hasStaticInits())
+          proc->insert(Stmt::call(SmackRep::STATIC_INIT));
+      } else if (naming.get(*func).substr(0, 18)  == "__SMACK_init_func_") {
+        assert(!func->doesNotReturn() && "Init functions cannot return a value");
```
zvonimir: You are using a wrong API call here...
It should be:
func->getReturnType()->isVoidTy()

zvonimir: Sorry, but one more thing - both of these assertions should go into addInitFunc.

share/smack/include/smack.h
```diff
@@ -17,6 +17,9 @@ void __SMACK_mod(const char *fmt, ...);
 void __SMACK_decl(const char *fmt, ...);
 void __SMACK_top_decl(const char *fmt, ...);
 
+// Sugar for __SMACK_init_func_XXX()
+#define __SMACK_INIT(x) __SMACK_init_func_##x()
```
zvonimir: I think you should add "void" here...

test/basic/init_funcs.c
```diff
@@ -0,0 +1,21 @@
+#include "smack.h"
+
+// @expect verified
+// @flag --unroll=1
+
+void __SMACK_INIT(defineProcessStates) {
+  __SMACK_top_decl("const unique $process_uninitialized: int;");
```
zvonimir: This is not an ideal usage of __SMACK_top_decl, since top_decl can be injected anywhere at it will still work. For that you don't really need __SMACK_INIT. Here is an example I had in mind:
int g = 10;

__SMACK_INIT(init1) {
  g = 11;
}

__SMACK_INIT(init2) {
  g = 12;
}

void main(void) {
  assert(g == 12);
}

Then in the _fail version, just change this assertion into:
  assert(g == 0 || g == 10 || g == 11);

I think that will be a better example...

snoopsmsc: I feel that this is slightly...  misleading?  I mean, these are all things you can do without __SMACK_INIT...

I guess I was trying to find something that couldn't be done without __SMACK_INIT...  But then again, I guess the whole purpose of it is to use from .h files, not from actual end-user files.

zvonimir: I see, fair enough, makes sense... Maybe you could create two examples - mine and yours :).
Also, please change your _fail version into something that resembles your not failing version. If you look at other regressions, that is how we typically do it. So maybe copy-paste more or less the same thing, and just change the final assertions into something that should fail.

test/basic/init_funcs.c
```diff
@@ -0,0 +1,18 @@
+#include "smack.h"
+
```
zvonimir: Please rename this file into init_funcs_global.c.

test/basic/init_funcs_fail.c
```diff
@@ -0,0 +1,18 @@
+#include "smack.h"
+
```
zvonimir: Please rename this file into init_funcs_global_fail.c.

test/basic/init_funcs_example.c
```diff
@@ -0,0 +1,38 @@
+#include "smack.h"
+
```
zvonimir: Please create a failing version of this and call it init_funcs_example_fail.c.
An assertion for that failing example could be:
__SMACK_code("assert($processStatus[@] != $process_uninitialized);", idx);

PR #85, opened 2015-03-31
Feature/threaded tests

jon-whit:
Tests can now be performed using multiple threads. Output information is now done with the logger class. I added two additional arguments to the tester that allow the user to choose the log level of the testing and an optional output log path.


README.md
```diff
@@ -1,3 +1,5 @@
+[![Build Status](http://kepler.cs.utah.edu:8080/buildStatus/icon?job=smack)](http://kepler.cs.utah.edu:8080/job/smack/)
```
zvonimir: Please remove these changes to README from this pull request since they are not related to threading.

jon-whit: Fixed. Sorry about that...

test/regtest.py
```diff
@@ -104,43 +104,142 @@ def metadata(file):
...
+
+    # start processing the tests. this blocks the main thread
+    async_results = []
+    for test in glob.glob("./**/*.c") if args.exhaustive else glob.glob("./basic/*.c"):
+        async_results.append(p.apply_async(process_test, args=(test, args,)))
+
+    try:
+        # keep the main thread active while there are active workers
+        count = 0
+        while True:
```
michael-emmi: Do we really need to do “polling” here, repeatedly checking the results of each async task?

It seems like it would be cleaner simply to tally the results once the tests have completed — either “naturally” or via an interrupt. In other words, after the call to `p.join()`.

test/regtest.py
```diff
@@ -104,43 +104,142 @@ def metadata(file):
...
-  
-except KeyboardInterrupt:
-  pass
+        str_result += '  [%.2fs]' % round(elapsed, 2)
 
-print
-print ' PASSED count:', passed
-print ' FAILED count:', failed
+        # log the info for this test
+        logging.info(str_result)
```
michael-emmi: This `logging.info(..)` call can be called concurrently from many threads. Is it thread safe?

test/regtest.py
```diff
@@ -104,43 +104,142 @@ def metadata(file):
...
 
-if timeouts > 0:
-  print 'TIMEOUT count:', timeouts
-
-if unknowns > 0:
-  print 'UNKNOWN count:', unknowns
+        # returns a status code based on the results of the test
+        if passed:
+            return 0
```
michael-emmi: Encoding results with integer literals leads to confusion. Better to use either strings, or integer constants.

test/regtest.py
michael-emmi: Looks like it’s OK, so long `multiprocessing.Pool` implements its pool with threads, not processes.

https://docs.python.org/2/howto/logging-cookbook.html#logging-to-a-single-file-from-multiple-processes

michael-emmi: Actually, looks like it’s not OK after all, since `multiprocessing.Pool` implements its “thread pool” with processes.

test/regtest.py
```diff
@@ -69,29 +73,25 @@ def metadata(file):
         m['expect'] = match.group(1).strip()
 
   if not m['skip'] and not 'expect' in m:
-    print red("WARNING: @expect MISSING IN %s" % file)
+    logging.warning("@expect MISSING IN %s" % file)
```
michael-emmi: This call to `logging.warning(..)` is not multi-process-safe — see discussion below.

test/regtest.py
```diff
@@ -104,43 +104,142 @@ def metadata(file):
...
+    logging.debug("Creating Pool with '%d' Workers" % args.n_threads)
+    p = multiprocessing.Pool(processes=args.n_threads, initializer=init_worker)
+
+    # start the tests
+    logging.info("Running regression tests...")
+    passed = failed = timeouts = unknowns = 0
+
+    # start processing the tests. this blocks the main thread
+    async_results = []
+    for test in glob.glob("./**/*.c") if args.exhaustive else glob.glob("./basic/*.c"):
```
michael-emmi: The inclusion of `./basic/*.c` should not be hardcoded. Instead, use the `meta['skip']` field for each individual test, as is done in the `develop` branch currently.

jon-whit: Thanks! I overlooked this. I was looking at the `master` branch when I copied this.

test/regtest.py
```diff
@@ -104,43 +114,149 @@ def metadata(file):
...
+    p = ThreadPool(processes=args.n_threads, initializer=init_worker)
+
+    # start the tests
+    logging.info("Running regression tests...")
+    passed = failed = timeouts = unknowns = 0
+
+    # start processing the tests. this blocks the main thread
+    async_results = []
+    for test in glob.glob("./**/*.c"):
+        async_results.append(p.apply_async(process_test, args=(test, args,)))
```
michael-emmi: The logic here seems a bit flawed, since only one task is being created per file, with a single return value. When the exhaustive flag is passed, then each file has several tests, corresponding to different memory models and verifiers — i.e., given in `meta['memory']` and `meta['verifiers']`.

The async task creation loop should thus look more like this:
````
for test in glob.glob(…):
  for memory in meta['memory'][…]:
    for verifier in meta['verifier'][…]:
      …
      p.apply_async(process_test, …)
````
Otherwise, the `--exhaustive` flag doesn’t really do anything.

File: test/regtest.py (hunk 2 above)
Comment (michael-emmi) [inline test/regtest.py]: A cleaner way to write this would be
````
while True:
  time.sleep(1)
````
and to do the tallying in a callback that gets passed to `apply_async` — the callbacks will be executed serially, on the main thread.

File: README.md [hunk 8]
```diff
@@ -54,7 +54,7 @@ While SMACK is designed to be a *modular* verifier, for our convenience, this
 whole process has also been wrapped into a single command in SMACK:
 
     smackverify.py simple.c -o simple.bpl
-
```
michael-emmi: This pull request does not concern `README.md`. Please remove this edit.

test/regtest.py
```diff
@@ -104,43 +114,149 @@ def metadata(file):
...
-print ' FAILED count:', failed
+        # log the info for this test
+        logging.info(str_result)
 
-if timeouts > 0:
-  print 'TIMEOUT count:', timeouts
-
-if unknowns > 0:
-  print 'UNKNOWN count:', unknowns
+        # returns a status code based on the results of the test
```
michael-emmi: Use 2 spaces of indentation, not 4.

test/regtest.py
jon-whit: Ok. Thanks. I will make this change.

jon-whit:
As you can tell, I am having issues with Git Diff between the `develop` and my branch. Even though the README on this branch is identical to the `develop` Git things they're different for some reason..

test/regtest.py
```diff
@@ -69,78 +79,171 @@ def metadata(file):
...
+    else:
+      return FAILED
+
+test_results = []
+def tally_result(result):
+    """
+    Called whenever a worker has finished its process. Only the main process
+    will modify test_results.
+    """
+    test_results.append(result)
```
michael-emmi: The indirection through `test_results` is unnecessary; `tally_result` can increment the `passed`, `failed`, `timeouts`, and `unknown` variables directly.

test/regtest.py
```diff
@@ -69,78 +79,171 @@ def metadata(file):
...
+      cmd += ['--time-limit', str(meta['time-limit'])]
+      cmd += ['-o', test + '.bpl']
+      cmd += meta['flags']
+
+      for memory in meta['memory'][:100 if args.exhaustive else 1]:
+        cmd += ['--mem-mod=' + memory]
+
+        for verifier in meta['verifiers'][:100 if args.exhaustive else 1]:
+          cmd += ['--verifier=' + verifier]
+          r = p.apply_async(process_test, args=(cmd, test, memory, verifier, meta,), callback=tally_result)
```
michael-emmi: Better to isolate use of `meta` to one function, i.e., here, and just pass the remaining value, `meta['expected']` to the `process_test` function as an `expected` parameter.

test/regtest.py
```diff
@@ -69,78 +79,171 @@ def metadata(file):
...
+      str_result += 'UNKNOWN'
+      unknown = True
+
+    else:
+      str_result += 'FAILED '
+
+    str_result += '  [%.2fs]' % round(elapsed, 2)
+
+    # log the info for this test
+    logging.info(str_result)
```
michael-emmi: Better to do logging in the `tally_result` callback, making it totally clear that the `logging.info(…)` calls are executed serially.

test/regtest.py
michael-emmi: The general idea is to isolate each concurrent task to non-shared data, and funnel the manipulation of shared data to the serially-executed callbacks.

test/regtest.py
```diff
@@ -69,78 +79,171 @@ def metadata(file):
...
+        cmd += ['--mem-mod=' + memory]
+
+        for verifier in meta['verifiers'][:100 if args.exhaustive else 1]:
+          cmd += ['--verifier=' + verifier]
+          r = p.apply_async(process_test, args=(cmd, test, memory, verifier, meta,), callback=tally_result)
+          results.append(r)
+
+    try:
+      # keep the main thread active while there are active workers
+      for r in results:
```
michael-emmi: Is this loop superior to the following?
````
while True:
    time.sleep(1)
````
in particular, will a keyboard interrupt work if the first result is not terminating any time soon?

File: test/regtest.py [hunk 14]
```diff
@@ -14,7 +19,7 @@
 
 def red(text):
   return '\033[0;31m' + text + '\033[0m'
-  
+
 def green(text):
   return '\033[0;32m' + text + '\033[0m'
```
jon-whit: Are these even needed anymore. They're not being used by the logging merely because the logging doesn't have a clean solution to do colored out. At least from what I've been reading, and the examples I've seen..

test/regtest.py
```diff
@@ -14,7 +19,7 @@
 
 def red(text):
```
jon-whit: Should we keep the `red` and `green` functions? They're no longer being used. I tried to utilize them, but the logging class doesn't have an elegant way of writing colored output to log files AND console.

test/regtest.py
```diff
@@ -69,81 +73,154 @@ def metadata(file):
...
   return m
-  
-parser = argparse.ArgumentParser()
-parser.add_argument("--exhaustive", help="be exhaustive", action="store_true")
-args = parser.parse_args()
 
-print "Running regression tests..."
-print
+# integer constants
+PASSED = 0; TIMEDOUT = 1; UNKNOWN = 2; FAILED = -1;
```
michael-emmi: These constants are no longer used.

test/regtest.py
```diff
@@ -69,81 +73,154 @@ def metadata(file):
...
+    """
+    str_result = "{0:>20}\n".format(test)
+    str_result += "{0:>20} {1:>10}    :".format(memory, verifier)
+
+    t0 = time.time()
+    p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    out, err  = p.communicate()
+    elapsed = time.time() - t0
+
+    str_result += '  [%.2fs]' % round(elapsed, 2)
```
michael-emmi: This is `str_result` puts things in the opposite order as before: should be
````
````

Comment (jon-whit) [inline test/regtest.py]: Sorry for that mix up. Thanks!

Comment (jon-whit) [inline test/regtest.py]: Mmm. Good catch! Thanks.

File: test/regtest.py [hunk 18]
```diff
@@ -69,81 +81,158 @@ def metadata(file):
...
+    elapsed = time.time() - t0
+
+
+
+    # get the test results
+    result = get_result(out+err)
+    if result == expect:
+      str_result += 'PASSED '
+      str_result += '  [%.2fs]' % round(elapsed, 2)
+      return green(str_result, log_file)
```
michael-emmi: Only the string `'PASSED'` was green — see how the existing regression script works.

test/regtest.py
```diff
@@ -69,81 +81,158 @@ def metadata(file):
...
+    p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    out, err  = p.communicate()
+    elapsed = time.time() - t0
+
+
+
+    # get the test results
+    result = get_result(out+err)
+    if result == expect:
+      str_result += 'PASSED '
```
michael-emmi:
````
str_result += green('PASSED ')
````

File: test/regtest.py [hunk 20]
```diff
@@ -69,81 +81,158 @@ def metadata(file):
...
+      str_result += '  [%.2fs]' % round(elapsed, 2)
+      return green(str_result, log_file)
+    elif result == 'timeout':
+      str_result += 'TIMEOUT'
```
michael-emmi:
````
str_result += red('TIMEOUT')
````

File: test/regtest.py [hunk 21]
```diff
@@ -69,81 +81,158 @@ def metadata(file):
...
+    elif result == 'unknown':
+      str_result += 'UNKNOWN'
+    else:
+      str_result += 'FAILED '
+
+    str_result += '  [%.2fs]' % round(elapsed, 2)
+    return red(str_result, log_file)
```
michael-emmi: See comments above.

PR #90, opened 2015-05-16
Fix Boogie and Corral wrapper scripts so that each argument passed to

delcypher:
the script is quoted. Previously if parameters with spaces in were passed to the script
then those parameters would be split into separate parameters.

E.g.

./boogie "my silly file name with spaces in it.bpl"

will expand to

/path/to/Boogie.exe "my" "silly" "file" "name" "with" "spaces" "in"
"it.bpl"

when invoked which would most likely fail.

### DIFF
--- a/bin/boogie
+++ b/bin/boogie
@@ -1,3 +1,3 @@
 #!/bin/bash
-$BOOGIE $@
+$BOOGIE "$@"
 
--- a/bin/corral
+++ b/bin/corral
@@ -1,3 +1,3 @@
 #!/bin/bash
-$CORRAL $@
+$CORRAL "$@"
 


MontyCarter:
Looks like we're on the right track, here.  Dan, have you checked this with other Boogie/Corral command line options?  I want to make sure that 
`boogie "space file.bpl" /loopUnroll:10`
doesn't get expanded to
`mono /path/to/boogie.exe "space file.bpl /loopUnroll:10"`
as this will give an unrecongnized arg output.

Also, this should go to the develop branch, rather than the master branch.  I don't see a way to change this, on my end...

MontyCarter:
@delcypher It occurred to me that this pull request may just be a courtesy, pointing us in the right direction...  If you don't have SMACK installed and running, we can test this on our end.

Let me know.

delcypher:
> Dan, have you checked this with other Boogie/Corral command line options? I want to make sure that 
boogie "space file.bpl" /loopUnroll:10
doesn't get expanded to
mono /path/to/boogie.exe "space file.bpl /loopUnroll:10"

Yes it expands as desired. This behaviour is described the bash man page

```
       @      
Expands  to  the  positional parameters, starting from one.  When the
expansion occurs within double quotes, each parameter expands to a separate word.  
That is, "$@" is equivalent to "$1" "$2" ... 
```

It is courtesy, I don't have SMACK built and installed right now but the wrapper script is easy to independently test from SMACK and seemed to work fine.

I can't change the branch this pull request was made against so I'm going to close this one and open #91

MontyCarter:
Great.  Thanks, Dan.  :)

It is much appreciated.

jon-whit:
Thanks for doing this Dan! Glad we got this one worked out.

PR #91, opened 2015-05-16
Fix Boogie and Corral wrapper scripts so that each argument passed to

delcypher:
the script is quoted. Previously if parameters with spaces in were passed to the script
then those parameters would be split into separate parameters.

E.g.

./boogie "my silly file name with spaces in it.bpl"

will expand to

/path/to/Boogie.exe "my" "silly" "file" "name" "with" "spaces" "in"
"it.bpl"

when invoked which would most likely fail.

### DIFF
--- a/bin/boogie
+++ b/bin/boogie
@@ -1,3 +1,3 @@
 #!/bin/bash
-$BOOGIE $@
+$BOOGIE "$@"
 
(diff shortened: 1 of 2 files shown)


delcypher:
@MontyCarter here's the new pull request.

PR #97, opened 2015-06-01
Small fixes to build.sh

MontyCarter:
* Delete /tmp/nuget/ after it gets created and used by 'mono nuget.exe' - if sudo user installs first and this folder isn't deleted, subsequent local, non-sudo installs can't write to this directory.
* Add --prefix INSTAlLDIR/bin to $PATH as first item, not last, so local installations get preference over system-wide installations
* Trim trailing '/' characters from --prefix argument values, to prevent seeing '//' in smack.environment

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -193,10 +193,10 @@ do
   case "$1" in
   --prefix)
     puts "Using install prefix $2"
-    INSTALL_PREFIX="$2"
+    INSTALL_PREFIX="${2%/}"
     CONFIGURE_INSTALL_PREFIX="--prefix=$2"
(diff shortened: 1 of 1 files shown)


bin/build.sh
```diff
@@ -331,6 +331,7 @@ then
   mozroots --import --sync
   ${WGET} https://nuget.org/nuget.exe
   mono ./nuget.exe restore Boogie.sln
+  rm /tmp/nuget/ -rf
```
zvonimir: Please update this into rm -rf /tmp/nuget/ to make it consistent with other rm commands.

MontyCarter:
New build.sh rm command is now updated to be consistent with other rm commands in build.sh.

Issue #99, opened 2015-06-08
Modularize smack.py


MontyCarter:
# Overview

I am proposing that we modularize `smack.py` into proper python modules.  As more and more features get added to `smack.py`, it is going to become large and unmanageable as a single file.  Creating a proper python module for smack will help remedy this.

Also, I feel that `smack.py` would greatly benefit from defining some good abstraction layers, such as input and output abstraction interfaces.  The move to `smack.py` was definitely an improvement, but to move things into submodules will require a cleaner abstraction, such as not relying on globals, for example.

This is rather long, so I've broken it in to sections.  The Background section isn't required to understand the proposal, but it gives some justification/rationale for the proposal.  I also have some subsections that give some additional suggestions/commentary that aren't centrally important to the main idea of modularizing `smack.py`.  These should be seen as what I, personally, had in mind, rather than requirements of the overall goal of modularizing `smack.py`.

Finally, I want to emphasize that this suggestion is NOT going back in the direction of having multiple scripts that get executed from the command line (as with `smackverify.py`, `smackgen.py`, and `llvm2bpl.py`).  Instead, the logic of `smack.py` would be entirely offloaded to a python module,  whose submodules don't get used by themselves.
# Background

To integrate support for svcomp, we decided to build the necessary options directly into the main toolchain executable (`smack.py`).  For example, adding the `--svcomp` flag includes a different header file, `svcomp-smack.h`, rather than `smack.h`.  This adds definitions and declarations for functions that tools competing in svcomp are required give special meaning/handling to.  Another example is the error trace format required by svcomp, whose generation will be enabled with `--error-witness`.

As more and more features like this get added, I'm concerned that `smack.py` is going grow in to a long, unwieldy file.

The extra code for the `--svcomp` flag is rather short, but the code for generating the error witness trace file is around 150 lines.

Currently, in the svcomp2016 branch, I have the svcomp error trace generation code in a pseudo-module called `toSVCOMPformat.py`, which is currently located in the `bin/` folder.  `smack.py` imports this, but because things aren't organized as formal python modules, `toSVCOMPformat.py` must be in the same folder as `smack.py`.

The result of this is that `toSVCOMPformat.py` must be installed to the machine/local bin folder.  As more features are added, this will either lead to a massive, messy `smack.py` file, or a `bin/` install folder with lots of cluttering python pseudo-modules that never get executed directly and so don't belong in the `bin/` folder.
# Motivation

[...]

PR #100, opened 2015-06-09
Integrate SVCOMP functionality

MontyCarter:
There are a number of things that smack needs to do differently to run on the SVCOMP benchmarks.  This pull request adds this functionality.

In addition to adding the svcomp functionality to smack, this pull request also includes "SMACKBench", which is basically a configuration script to set up a benchmarking framework very similar to what is used in the SVCOMP competition itself.  This allows us to estimate our performance for the SVCOMP competition, as well as provide a nice method for discovering bugs, performance issues, and testing various combinations of CLI parameters to see how they perform on a large group of benchmarks.

Finally, this pull request adds a file, test-bpl.sh, which is used to generate bpl files for all svcomp benchmarks.  Note that this is not currently working because smack.py's `--no-verify` flag is not currently validating generated .bpl files to see if they are valid boogie programs (this is because `boogie /noVerify` is not generating a non-zero exit code when parsing, name resolution or type checking fail).

Please note that I expect there to be a lot of issues with this pull request.  A fair amount of changes in this pull request were carried over from last year's svcomp support, and so we just kind of propagated the things that were working.  Many of these, I suspect, wont fit within conventions/organizational expectations for the smack project, but I didn't know what the preferred choices would be, so I left them as is, with the expectation that they would be fleshed out during the pull request.


CMakeLists.txt
```diff
@@ -184,6 +184,9 @@ INSTALL(TARGETS smack
 INSTALL(FILES ${CMAKE_CURRENT_SOURCE_DIR}/bin/boogie
   ${CMAKE_CURRENT_SOURCE_DIR}/bin/corral
   ${CMAKE_CURRENT_SOURCE_DIR}/bin/smack.py
+  ${CMAKE_CURRENT_SOURCE_DIR}/bin/tokenizer
+  ${CMAKE_CURRENT_SOURCE_DIR}/bin/token_replace.py
```
MontyCarter: Please see #99  for a proposal on how to handle this clutter in the `bin/` installation directory.

CMakeLists.txt
```diff
@@ -184,6 +184,9 @@ INSTALL(TARGETS smack
 INSTALL(FILES ${CMAKE_CURRENT_SOURCE_DIR}/bin/boogie
   ${CMAKE_CURRENT_SOURCE_DIR}/bin/corral
   ${CMAKE_CURRENT_SOURCE_DIR}/bin/smack.py
+  ${CMAKE_CURRENT_SOURCE_DIR}/bin/tokenizer
```
MontyCarter: This is a binary provided by SVCOMP for converting C files to have one token per line.  I don't know of a better place to put this than bin, since it is not really part of smack...

I suppose we could put it in the `share/` installation directory, and reference it using relative paths...

bin/smack.py
```diff
@@ -140,11 +198,22 @@ def arguments():
   verifier_group.add_argument('--smackd', action="store_true", default=False,
     help='generate JSON-format output for SMACKd')
 
+  svcomp_group = parser.add_argument_group('svcomp options')
+
+  svcomp_group.add_argument('--svcomp', action="store_true", default=False,
+    help='enter svcomp mode')
+
+  svcomp_group.add_argument('--syntax-check', action="store_true", default=False,
```
MontyCarter: Shaobo put this in - I am not sure it does anything yet...

However, I kind of had a different idea.  Instead of having an explicit `--syntax-check` flag that validates generated .bpl files, I think it makes sense to always validate bpl files to see if they are valid boogie programs by passing them through boogie by calling `boogie /noVerify ...`.  When the `--no-verify` flag is not set, either boogie or corral validate the bpl files for parse, name resolution, and typechecking errors.  It seems reasonable that we should still validate bpl files even when they aren't being verified.

We haven't integrated this validation functionality yet because boogie does not set a non-zero exit code when one of these parsing related issues is raised.  There is currently a proposed patch to fix this (boogie-org/boogie#13).  We'll add this functionality when the patch is completed and accepted.

share/smack/include/smack.h
```diff
@@ -24,6 +24,10 @@ void __SMACK_top_decl(const char *fmt, ...);
 // with an integer argument (DSA gets confused otherwise)
 __attribute__((always_inline)) void __SMACK_dummy(int v);
 
+#ifdef SVCOMP
```
zvonimir: I don't particularly like this solution. Ideally, SVCOMP and SMACK benchmarks should be compatible. I emailed SVCOMP mailing list to ask about this.

bin/smack.py
```diff
@@ -15,13 +15,71 @@
 import sys
 import tempfile
 from threading import Timer
+from toSVCOMPformat import *
+from token_replace import *
 
 VERSION = '1.5.1'
 temporary_files = []
 
+def rewriteIExtensionToC(args):
```
zvonimir: Is it possible to force clang to do preprocessing? Maybe using some command line switch?

bin/smack.py
```diff
@@ -140,11 +198,22 @@ def arguments():
...
+
+  svcomp_group.add_argument('--error-witness', metavar='FILE', default=None, type=str, 
+    help='save error witness to FILE')
+
   args = parser.parse_args()
 
   if not args.bc_file:
     args.bc_file = temporary_file('a', '.bc', args)
-
+
```
zvonimir: Remove these spaces on this empty line.

bin/smack.py
zvonimir: --syntax-check should I think be under "verifier options".

bin/smack.py
```diff
@@ -222,18 +301,23 @@ def clang_frontend(args):
...
   smack_headers = os.path.join(smack_root, 'share', 'smack', 'include')
-  smack_lib = os.path.join(smack_root, 'share', 'smack', 'lib', 'smack.c')
+  smack_lib = os.path.join(smack_root, 'share', 'smack', 'lib')
   smack_bc = temporary_file('smack', '.bc', args)
+  smack_svcomp_bc = temporary_file('smack-svcomp', '.bc', args)
 
   compile_command = ['clang', '-c', '-emit-llvm', '-O0', '-g', '-gcolumn-info']
   compile_command += args.clang_options.split()
-  compile_command += ['-I' + smack_headers, '-include' + 'smack.h']
+  compile_command += ['-I' + smack_headers, '-include' + ('smack-svcomp.h' if args.svcomp else 'smack.h')]
```
zvonimir: I would like to remove smack-svcomp.h if at all possible.

bin/smack.py
```diff
@@ -379,6 +469,7 @@ def smackdOutput(corralOutput):
     traces = []
     for traceLine in corralOutput.splitlines(True):
       traceMatch = re.match('(' + FILENAME + ')\((\d+),(\d+)\): Trace: Thread=(\d+)  (\((.*)\))?$', traceLine)
+      traceAssumeMatch = re.match('(' + FILENAME + ')\((\d+),(\d+)\): Trace: Thread=(\d+)  (\((\W*\w+\W*=\W*\w+\W*)\))$', traceLine)
```
zvonimir: Why did you change this line?

bin/smack.py
```diff
@@ -386,7 +477,15 @@ def smackdOutput(corralOutput):
         colno = int(traceMatch.group(3))
         threadid = int(traceMatch.group(4))
         desc = str(traceMatch.group(6))
-        trace = { 'threadid': threadid, 'file': filename, 'line': lineno, 'column': colno, 'description': '' if desc == 'None' else desc }
+        assm = ''
```
zvonimir: Why did you change this line?

bin/smack.py
MontyCarter: We thought about this, but then apparently forgot to look in to this.

I'll take a look now.

svcomp/bench/toInstallDir/benchexec/benchexec/tools/smack.py
```diff
@@ -0,0 +1,87 @@
+import os
```
zvonimir: We should rename this file if at all possible. We should not have two smack.py files. Please comment.

svcomp/bench/toInstallDir/data/index.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
```
zvonimir: Please add comments explaining what this file does. Also, is python3 now required?

svcomp/bench/toInstallDir/data/results.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python3
```
zvonimir: Please add comments explaining briefly what this file does.

svcomp/bench/toInstallDir/data/lib.py
```diff
@@ -0,0 +1,82 @@
+import cgi
```
zvonimir: Please add comments explaining briefly what this file does.

svcomp/bench/toInstallDir/runSMACKBench.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
zvonimir: Not sure where to leave this comment. I don't particularly like the name of this folder (toInstallDir). Why did you call it like that?
If nothing else, then I would at least probably prefer to-install-dir.

bin/smack.py
MontyCarter: This extracts a variable symbol and its assumed value, and makes it available from the match groups.  This is then used to add assume statements to the witness file to tell the witness checker what values were assumed for variables at various points.

bin/smack.py
MontyCarter: Ok, so I talked to Shaobo, and I found out more about this...  

So we are including smack.h to benchmarks not by injecting `#include <smack.h>` into the benchmark, but rather by using the clang switch `-Ismack.h` (or currently, `-Ismack-svcomp.h`).  This is preprocessor functionality.

When clang sees the file is a `.i` file, it disables the preprocessor, which means that even files included with the `-I` switch don't get included.

A cursory search turned up ways to *disable* the preprocessor, but I haven't found a way to force it on.

At this point, I think it might be worth spending a little more time to look for a way to force the preprocessor on, but I haven't found a solution yet.  I just wanted to add this comment as a status update with some additional info.

bin/smack.py
MontyCarter: For each trace line that we parse, the witness file needs to know if there is an assumption made by the tool regarding a specific variable, allowing the witness checker to make the same assumption and thus prune the search space.

With smack, we don't see a variable assumption on each trace line.  Clearly, then for those without, we can't add a variable value assumption to the witness file.  For such trace lines, this change adds the assumption as an empty string (this makes it easier when reading the json later - we know the `'assumption'` child node will always exist, so we can just check to see if it has a value).  For trace lines that DO have a variable value assumption, the symbol and value are extracted from the `traceAssumeMatch` groups, and added as the `'assumption'` child to the json `'trace'` node.

I split the assignment of `trace` on to multiple lines because it was becoming unwieldy with the addition of the `'assumption'` child node.

svcomp/bench/toInstallDir/benchexec/benchexec/tools/smack.py
MontyCarter: Done

svcomp/bench/toInstallDir/data/index.py
MontyCarter: Done.  Also, the python3 requirement has been removed.

(I did this because I've just been used to programming in python3...  There were only a few print calls that had to be modified.)

svcomp/bench/toInstallDir/data/results.py
MontyCarter: Done

svcomp/bench/toInstallDir/data/lib.py
MontyCarter: Done.

svcomp/bench/toInstallDir/runSMACKBench.sh
MontyCarter: The name of this folder has been changed to `src`.  I'm not sure if this is ideal, since nothing in this folder gets compiled.  However, I couldn't really think of a better name - this folder contains some scripts, some library files for the scripts, and some "overlay" files (which get copied on top of an installation of BenchExec, for example).  With this said, however, everything in the now-named `src` folder only works when they've been copied to an "install" directory, where BenchExec, cpachecker, and sv-comp benchmarks have been installed.

MontyCarter:
I think I've addressed a majority of the issues you've raised.  Here are the things that are outstanding:
* `smack.py` python modularization
    * This might be better to leave as a separate issue, since we might not reach consensus and implementation in a timely manner?
*  `smack-svcomp.h` elimination
    * Waiting for finalization of conversation with Dirk, and determination on how to proceed. See [google groups conversation](https://groups.google.com/forum/#!topic/sv-comp/8aXOFR6_LaQ).
* Acceptance of folder rename from `toInstallDir` to `src`.

bin/test-bpl.sh
```diff
@@ -0,0 +1,47 @@
+GOOD=""
```
zvonimir: Please add comments about what this script does and how to invoke it.

MontyCarter: This file is a script to run the bpl translation verification (for that other repo).

I didn't do any commenting because we decided to do that functionality (bpl translation + validation, but no verification) as a separate pull request.

I left this in there because it is fully functional, and didn't know where else to store it until we do the second pull request.  The other things regarding bpl translation+validation were hacky and incomplete, so I just removed them (from `smack.py`, for example).

In any event, it'll need comments in the long run, so I'll comment it and commit/push again.

MontyCarter: This file has been updated with comments, usage, and error handling for CLI arguments.

I wonder if this file should be moved to a different location, however, I'm not sure where else to put it currently.  I propose we address this when the bpl validation is reimplemented.

share/smack/lib/smack-svcomp.c
```diff
@@ -4,11 +4,11 @@
 #include "smack-svcomp.h"
 
 void __VERIFIER_error(void) {
-  __VERIFIER_assert(0);
+  assert(0);
```
michael-emmi: Both `__VERIFIER_assert` and `assert` are identical. `__VERIFIER_assert` is preferred here.

share/smack/include/smack.h
michael-emmi: `assume` and `assert` should be defined by `__VERIFIER_{assume,assert}`.

share/smack/lib/smack-svcomp.c
michael-emmi: Please merge these definitions into `smack.c`.

bin/smack.py
michael-emmi: Agreed. `smack.h` and `smack-svcomp.h` should be merged.

bin/smack.py
```diff
@@ -15,13 +15,71 @@
...
+    if proc: os.killpg(os.getpgid(proc.pid), signal.SIGKILL)
+
+  temporary_files.append(beforeTokenizedName)
+  args.input_file = tokenizedName
+  return args
+
 def frontends():
   """A dictionary of front-ends per file extension."""
   return {
+    '.i': clang_frontend,
```
michael-emmi: Can we assume that all SVCOMP inputs have the `.i` extension? In that case, we should remove the `--svcomp` flag, and switch to a custom front end here.

bin/smack.py
```diff
@@ -140,6 +198,14 @@ def arguments():
   verifier_group.add_argument('--smackd', action="store_true", default=False,
     help='generate JSON-format output for SMACKd')
 
+  svcomp_group = parser.add_argument_group('svcomp options')
+
+  svcomp_group.add_argument('--svcomp', action="store_true", default=False,
```
michael-emmi: Can this flag be removed by assuming that `.i` files will be processed by a special SVCOMP front-end?

bin/smack.py
```diff
@@ -140,6 +198,14 @@ def arguments():
...
+    help='enter svcomp mode')
+
+  svcomp_group.add_argument('--error-witness', metavar='FILE', default=None, type=str,
```
michael-emmi: This flag should be merged with the `--trace-file` flag because they both have a common meaning: file to store witness of violation. In the case that we are in SVCOMP mode, e.g., because the input file had the `.i` extension, then we can output the error witness instead of a regular trace. Same flag though.

bin/smack.py
```diff
@@ -155,6 +221,16 @@ def arguments():
   #     if m:
   #       return args = parser.parse_args(m.group(1).split() + sys.argv[1:])
 
+  if args.svcomp:
```
michael-emmi: Please replace this code and the `replacer` function by a `svcomp_frontend` function which is invoked when the input file extension is `.i`.

bin/smack.py
michael-emmi: Please merge {`smack.h`, `smack-svcomp.h`} and {`smack.c`, `smack-svcomp.c`} and revert this code.

bin/smack.py
```diff
@@ -222,18 +298,23 @@ def clang_frontend(args):
...
+  smack_lib = os.path.join(smack_root, 'share', 'smack', 'lib')
   smack_bc = temporary_file('smack', '.bc', args)
+  smack_svcomp_bc = temporary_file('smack-svcomp', '.bc', args)
 
   compile_command = ['clang', '-c', '-emit-llvm', '-O0', '-g', '-gcolumn-info']
   compile_command += args.clang_options.split()
-  compile_command += ['-I' + smack_headers, '-include' + 'smack.h']
+  compile_command += ['-I' + smack_headers, '-include' + ('smack-svcomp.h' if args.svcomp else 'smack.h')]
   compile_command += ['-DMEMORY_MODEL_' + args.mem_mod.upper().replace('-','_')]
+  if args.svcomp:
```
michael-emmi: This should be contained to the `svcomp_frontend` function.

bin/smack.py
```diff
@@ -322,7 +403,14 @@ def verify_bpl(args):
   if args.smackd:
     print smackdOutput(verifier_output)
 
+
+
   else:
+    if args.error_witness and result == 'error':
```
michael-emmi: Please unify this with the `trace_file` stuff below — e.g., by case splitting between `smackJsonToXmlGraph` and `error_trace`.

bin/smack.py
MontyCarter: No.  *Most* of the svcomp benchmarks are .c, not .i.  The .i files are used when benchmarks require several includes for typedefs - rather than requiring competing tools to handle includes, the preprocessor alone is executed by the competition organizers, who then deliver a .i file to competitors, where the includes have already been included.

bin/smack.py
MontyCarter: I suppose we could switch to a custom front end front end for `.i` files, however this wouldn't eliminate the need for `--svcomp`, as indicated in the above comment.

Also, I feel this wouldn't be the cleanest solution, anyway.  A `.i` extension indicates that the file has already been preprocessed, not that it is an svcomp benchmark.  It is conceivable that a user may want to verify a preprocessed file outside of the context of svcomp.

With that said, it *would* be nice to have some sort of detection feature to switch on svcomp mode...

bin/smack.py
MontyCarter: I like this idea because I can't really imagine a scenario when we would need both a `.txt` trace output and a `.graphML` error witness trace.

With that said, however, I dislike this idea because the purpose of `--error-witness` is clear from context, and it is self documenting.  If we get rid of `--error-witness` in lieu of `--trace-file` whose output format is dependent on whether the `--svcomp` mode is set, then users will have to be aware that to get an error witness, you must be in `--svcomp` mode, and you must use the `--trace-file` flag.

Getting rid of `--error-witness` would also mean that there is no way to get a `.txt` error trace file when using `--svcomp` mode.  This is probably fine, since you can always pipe output to file, but it should at least be considered when making a decision here.

Finally, I think it is conceivable that we might want to use error-witness checking outside of the context of svcomp, to 'verify' our verifications.  For example, if we have a non-svcomp program that is giving an error, but we believe it should be bug free, we could use error witness checking to confirm or dispel our suspicions.

I really am fine, either way - I don't have a strong opinion about this.  However, I wanted to point out the few considerations that came to mind on this.

Can we come to a consensus on this, and I'll be happy to implement whichever option we converge on?

zvonimir: I think that choosing an error trace format should be orthogonal to the input file, if at all possible. Anyhow, I think we should merge this pull request, and then refine this implementation, and also hopefully split it into modules. So let me do this merge first.

PR #108, opened 2015-07-08
Pthreads feature integration

MontyCarter:
This pull request adds support for verification of programs using the pthread library.

The benchmarks added to the `examples/pthread` folder probably need to be cleaned up, but I'm not sure which ones we should keep, and whether some of them belong in the regressions test folder.

The benchmarks in the `test/pthread_extras` folder were some older regressions that I decided not to use as real regression tests, but might be useful when working on performance tuning of pthreads, since they are currently taking a LONG time for being relatively simple benchmarks.


bin/smack.py
```diff
@@ -233,6 +239,11 @@ def clang_frontend(args):
   smack_headers = os.path.join(smack_root, 'share', 'smack', 'include')
   smack_lib = os.path.join(smack_root, 'share', 'smack', 'lib', 'smack.c')
   smack_bc = temporary_file('smack', '.bc', args)
+  if args.pthread:
```
MontyCarter: These things also need to happen in the svcomp frontend, however, I wonder if we should break this out into a function, rather than copy/pasting this.

bin/smack.py
```diff
@@ -254,7 +260,16 @@ def clang_frontend(args):
   smack_bc = temporary_file('smack', '.bc', args)
   try_command(compile_command + [smack_lib(), '-o', smack_bc])
   try_command(compile_command + [args.input_file, '-o', args.bc_file])
-  try_command(['llvm-link', args.bc_file, smack_bc, '-o', args.bc_file])
+  link_targets = [args.bc_file, smack_bc]
```
MontyCarter: Mike (and Z...), as we add more features to SMACK, we'll probably end up needing to support additional SMACK provided libraries.  We might want to come up with a less hard coded way of compiling and linking SMACK provided libraries like these.

bin/smack.py
```diff
@@ -264,13 +279,23 @@ def json_compilation_database_frontend(args):
...
 
   with open(args.input_file) as f:
     for cc in json.load(f):
       if 'objects' in cc:
         # TODO what to do when there are multiple linkings?
         bit_codes = map(lambda f: re.sub('[.]o$','.bc',f), cc['objects'])
-        try_command(['llvm-link', '-o', args.bc_file] + bit_codes + [smack_bc])
+        try_command(['llvm-link', '-o', args.bc_file] + bit_codes + smack_lib_bcs)
 
       else:
```
MontyCarter: Mike, will you double check this?  I'm not really sure what is going on here in this `else` branch...  It doesn't seem like it deals with SMACK provided libraries at all (and thus no need to support SMACK's pthread.c/spinlock.c libraries), but I'd like to make sure everything is good here.

michael-emmi: Let’s ignore this particular `json_compilation_database_frontend` method for now: it is still under development.

To answer your question though: the `if` branch is supposed to simulate linking commands from a compilation database, while the `else` branch should handle compilation commands. Rightly so, the compilation commands need not deal with linking libraries, since that is only a concern for linking.

bin/smack.py
zvonimir: Yeah, good point. But I am fine with your current solution for now.

bin/smack.py
```diff
@@ -376,6 +401,8 @@ def verify_bpl(args):
     command += ["/cex:%s" % args.max_violations]
     command += ["/maxStaticLoopBound:%d" % args.loop_limit]
     command += ["/recursionBound:%d" % args.unroll]
+    if args.context_switches:
```
zvonimir: Monty, this part is wrong. If you look several lines above this, you will see that we are already passing /k:1 to Corral. The reason I had to do this is because Corral by default in fact sets k to 2. So your context_switches parameter should actually be set to 1 by default, and then slightly rewrite this part so that k is always set, meaning that you don't need the ``if args.context_switches`` part.

bin/smack.py
```diff
@@ -141,6 +144,9 @@ def arguments():
   verifier_group.add_argument('--loop-limit', metavar='N', default='1', type=int,
     help='upper bound on minimum loop iterations [default: %(default)s]')
 
+  verifier_group.add_argument('--context-switches', metavar='k', default='2', type=int,
```
zvonimir: This should be 1 by default. See my comment below.

test/pthread/cond.c
```diff
@@ -0,0 +1,41 @@
+// Generally tests pthread_cond_wait, pthread_cond_signal,
+//  pthread_cond_init
```
zvonimir: You have one extra space here I think.

test/.gitignore
```diff
@@ -2,3 +2,5 @@
 *.bc
 *.o
 *.bpl
+corral_out_trace.txt
```
zvonimir: Do we need these? The way we are invoking Corral now, I think that no output files should be generated by Corral. Am I wrong about that?

test/pthread/.gitignore
```diff
@@ -0,0 +1 @@
+*~
```
zvonimir: Why do we need this?

test/pthread/cond_fail.c
```diff
@@ -0,0 +1,41 @@
+// Generally tests pthread_cond_wait, pthread_cond_signal,
+//  pthread_cond_init
```
zvonimir: Same here :). Why this extra space before pthread_cond_init?

test/pthread/config.yml
```diff
@@ -0,0 +1,3 @@
+flags: ['--pthread']
```
zvonimir: You have to set context switches to something more than 1 here.

test/pthread/exit.c
```diff
@@ -0,0 +1,29 @@
+
+// @expect verified
+
+#include <pthread.h>
+
+int x = 1;
+
+void *t1(void *arg)
+{
```
zvonimir: These curly braces are not according to our coding standard (see your main procedure below).

test/pthread/join.c
```diff
@@ -0,0 +1,24 @@
+// Shows pthread_join effectively blocks until child thread done
+
+// @expect verified
+
+#include <pthread.h>
+
+int x = 1;
+
+void *t1(void *arg)
+{
```
zvonimir: Again, curly braces issue.

test/pthread/join_fail.c
```diff
@@ -0,0 +1,24 @@
+// Shows join.c fails when parent doesn't wait for child before x++ call
+
+// @expect error
+
+#include <pthread.h>
+
+int x = 1;
+
+void *t1(void *arg)
+{
```
zvonimir: Again, curly braces issue.

test/pthread/join_return.c
```diff
@@ -0,0 +1,37 @@
...
+
+int x = 1;
+
+typedef struct pair{
+  int x;
+  int y;
+} pair;
+
+void *t1(void *arg)
+{
```
zvonimir: Again, curly braces issue. You get the point I am sure :).

test/pthread/join_return_fail.c
```diff
@@ -0,0 +1,37 @@
+// Tests join & return in general - would fail if join doesn't block.
+//  Also tests that exit and join successfully store/load a return value
```
zvonimir: Hmmm, again, why these extra spaces in the beginning of lines? Please align those properly.

test/pthread_extras/fib3_false.c
```diff
@@ -0,0 +1,29 @@
+
```
zvonimir: What is this ``pthread_extras`` folder??

examples/pthreads/CIVL-0.17_2046/27_Boop_simple_vf_false.c
```diff
@@ -0,0 +1,83 @@
+#include <pthread.h>
```
zvonimir: So in general we want to use the examples folder for nice illustrative examples. For instance, it would be good to place in the examples folder that standard bank account example. So please do that. I think Dekker would also be nice to keep in the examples folder maybe. I think that everything else should be moved to our regressions folder (test). So please restructure your files in such a way. Maybe in your ``test/pthread`` folder it would be good to prefix CIVL examples with ``civl-`` and svcomp example with `svcomp-`` or something like that.

examples/pthreads/svcomp2015/unseen/scull_safe.h
```diff
@@ -0,0 +1,72 @@
+#define inode int // the lock
```
zvonimir: What are unseen examples?

share/smack/include/pthread.h
```diff
@@ -0,0 +1,71 @@
+//
```
zvonimir: We should go over these key files in person once you are back. And also we should merge before we go over them.

bin/smack.py
MontyCarter: Ah.  I see.  This makes sense.

So do you want me to remove these lines from `json_compilation_database_frontend()`, then, before we merge this pull request, or should I just leave it?

bin/smack.py
MontyCarter: So this is fixed in my next commit, however, now, I'm wondering why it should be a problem that Corral would have `k` set to 2 by default.  For single-threaded input programs, is there any difference when `k == 1` vs `k == 2`, since there are no opportunities for context switches, anyway?

bin/smack.py
MontyCarter: This is fixed in my next commit.

test/pthread/cond.c
MontyCarter: Sometimes (for no real reason - and not consistently) I indent my comments when they go past a single line.

This is fixed in my next commit.

test/.gitignore
MontyCarter: No, I think you are correct.

This is taken out in my next commit.

test/pthread/.gitignore
MontyCarter: Quick answer, we don't.  It is taken out in my next commit.

Long answer is that emacs creates a file like `editedFile.ext~` if you modify and save `editedFile.ext`.  This gives you a chance to revert your changes to the original if you screw things up.  I just added this to take them out of my `git status` view.  I usually add files manually to a commit, not as a group, so it isn't really a problem to get rid of this.

test/pthread/cond_fail.c
MontyCarter: Fixed in next commit.

test/pthread/config.yml
MontyCarter: Fixed in next commit.

test/pthread/exit.c
MontyCarter: Sorry...  I didn't know there was one regarding curly braces.

This is fixed in next commit.

test/pthread/join.c
MontyCarter: Fixed in next commit.

test/pthread/join_fail.c
MontyCarter: Fixed in next commit.

test/pthread/join_return.c
MontyCarter: Fixed in next commit.

test/pthread/join_return_fail.c
MontyCarter: I went through all of the files in the `test/pthread/` folder, and aligned all comments, and looked for non-conforming curly braces.

These will all appear in my next commit.

test/pthread_extras/fib3_false.c
MontyCarter: These are basically a bunch of tests that I wrote that I decided to exclude from the main folder that exercises the pthread library.

Most of these simply take a LONG time to run, and so didn't really seem appropriate for our general `test/pthread/` folder, but I didn't want to get rid of them straight away because they contain interesting performance bugs that should be worked out.

This certainly doesn't need to be here for any particular reason, but I didn't really know what to do with these examples.

Do you have any ideas on what we should do with these?

examples/pthreads/CIVL-0.17_2046/27_Boop_simple_vf_false.c
MontyCarter: Let's talk about how to organize these when we meet to review the meat of pthreads pull request.

I'm not sure that all of these examples belong in our regression tests - I'll look into these examples more before we meet, so I know what is in them, what works, what doesn't, etc.

examples/pthreads/svcomp2015/unseen/scull_safe.h
MontyCarter: I put all of the benchmarks I wanted to evaluate into this unseen folder, and then as I assessed them for suitability for benchmarking, I moved them into the other folders.

I don't remember why this scull example never got moved into one of the other folders.  As I mentioned, I'll go through all the examples so I know why each of them is categorized as they are.

share/smack/include/pthread.h
MontyCarter: I'll go through the examples today and tomorrow, and make comments for each of the files regarding suitability, etc.

Let me know when you want to sit down.

MontyCarter:
All examples have been moved to regression tests, and they have been properly configured with `./regtest.py`'s yml configuration.

All long running tests are now in `pthread_extras` folder, and will only run when the `--all-examples` flag is set for `regtest.py`.

There are two subfolders in `pthread_extras` which never get run.  One of these folders contains tests that are not currently compatible with SMACK (svcomp `__VERIFIER_atomic_*()` functions are not currently supported).  The other folder has some tests that take a long time to run and would be good for tuning the performance of our pthread support.  These should be integrated into the main pthread regression folders when they become supported.

There is still one failing regression test, `pthread/exit.c`, which is failing because we have not yet decided on an implementation for `exit()` functionality.

bin/smack.py
```diff
@@ -141,6 +144,9 @@ def arguments():
   verifier_group.add_argument('--loop-limit', metavar='N', default='1', type=int,
     help='upper bound on minimum loop iterations [default: %(default)s]')
 
+  verifier_group.add_argument('--context-switches', metavar='k', default='1', type=int,
```
michael-emmi: This flag is improperly named. Corral’s `/k:int` flag specifies the number of contexts to allow during a single execution. The number of “context switches” is one less than the number of contexts. For example, in an execution with one context, there are zero context switches.

bin/smack.py
```diff
@@ -141,6 +144,9 @@ def arguments():
...
+    help='switch thread contexts in Corral (only) up to k number of times')
```
michael-emmi: We want this flag to make sense independently of the back-end verifier, so better not to mention Corral. (Since this number is a *bound*, and all tools support at least one context, this is not complete nonsense.) Also, this help message should be semi grammatical and natural. How about “bound the number of thread contexts” ?

share/smack/include/spinlock.h
```diff
@@ -0,0 +1,31 @@
+//
+// Copyright (c) 2013 Zvonimir Rakamaric (zvonimir@cs.utah.edu),
+//                    Michael Emmi (michael.emmi@gmail.com)
+// This file is distributed under the MIT License. See LICENSE for details.
+//
+#ifndef SPINLOCK_H
+#define SPINLOCK_H
+
+#include "smack.h"
```
michael-emmi: Why is `smack.h` included here?

share/smack/lib/pthread.c
```diff
@@ -0,0 +1,253 @@
+#include "pthread.h"
```
michael-emmi: `smack.h` should be included here.

share/smack/lib/spinlock.c
```diff
@@ -0,0 +1,20 @@
+#include "spinlock.h"
```
michael-emmi: `smack.h` should be included here.

share/smack/include/pthread.h
```diff
@@ -0,0 +1,71 @@
...
+
+#define UNLOCKED 0
+#define UNINITIALIZED 0
+#define INITIALIZED 1
+//  !!!!! ERROR HERE? - if mutex already initialized, behavior
+//        of initialization should be undefined
+#define PTHREAD_MUTEX_INITIALIZER {UNLOCKED, INITIALIZED}
+
+//Declare corral primitive procedures
+__SMACK_INIT(corral_primitives);
```
michael-emmi: Is it possible to do this in the implementation (`.c`) file instead?

share/smack/include/pthread.h
```diff
@@ -0,0 +1,71 @@
...
+
+// Initialize thread status tracking variables (OS level behavior)
+__SMACK_INIT(thread);
```
michael-emmi: Is it possible to do this in the implementation (`.c`) file instead?

share/smack/include/pthread.h
```diff
@@ -0,0 +1,71 @@
...
+// Copyright (c) 2013 Zvonimir Rakamaric (zvonimir@cs.utah.edu),
+//                    Michael Emmi (michael.emmi@gmail.com)
+// This file is distributed under the MIT License. See LICENSE for details.
+//
+#ifndef PTHREAD_H
+#define PTHREAD_H
+
+// Include pthreadtypes.h
+#include "pthreadtypes.h"
+#include "smack.h"
```
michael-emmi: Better not to include `smack.h` here, if possible — see below.

bin/smack.py
MontyCarter: So there is some ambiguity for me, on this point...

Consider the following program:
```
#include <pthread.h>

int z = 1;

void *t1(void *arg) {
  pthread_mutex_t* lock = arg;
  pthread_mutex_lock(lock);
  z++;
  pthread_mutex_unlock(lock);
}

void *t2(void *arg) {
  if(z==3)
    assert(0);
}

int main() {
  pthread_mutex_t lock;
  pthread_mutex_init(&lock, 0);

  pthread_t tid1, tid2;

  pthread_create(&tid1, 0, t1, &lock);
  pthread_create(&tid2, 0, t2, &lock);
  pthread_mutex_lock(&lock);
  z++;
  pthread_mutex_unlock(&lock);
  pthread_join(tid1, 0);
  pthread_join(tid2, 0);
}
```

The only way for the assertion in `t2` to fail is if the main calling thread, and `tid1` have already incremented `z`.  With the current definition of `--context-switches`, if `--context-switches=2` (i.e., `/k:2`), it finds the bug.  However, this obviously requires considering statements in all three threads, which requires 3 contexts and 2 switches between them.

The bug is not found when `--context-switches=1`.

My original interpretation was the same as yours -- that `/k` was the number of contexts to consider, however, that isn't really the behavior that I see.

Thoughts on this?  I am fine to change this, if you want to be consistent with the way `/k` is described by Corral.  (Though, perhaps other backends may set limits in terms of context switches, rather than contexts to be considered.  Maybe it would be a good idea to take a parameter like `--context-switches`, and add one to this value before passing it to `/k` for corral...)

michael-emmi: I think there are many possible hypotheses to be made here, depending, e.g., on what is Corral’s default schedule of threads (is `main` first, or last?), e.g., on whether `main` must complete its execution (and thus have joined both threads), e.g., on whether a switch is counted when a thread completes, etc. My hunch, mainly driven by how Corral describes its own flag, is that the parameter is the number of execution contexts *per thread*. So for instance, supposing that `main` must complete its execution (and also depending on other factors, like your implementation of `pthreads`) the `main` thread must be given two contexts — the first to create the other threads, and the second to join them. But there are many moving parts here, and I have not reviewed thoroughly your `pthreads` implementation, nor the inner workings of Corral.

Above all though, it cannot be right to call this parameter “context switches” when the default value of `1` is used for sequential programs — which do not do context switches.

So perhaps it’s best that we don’t make a firm call on the precise meaning of this parameter, at least not just yet. The safest thing would be to remain vague, but not too vague. Maybe `--context-bound` would be a good choice?

bin/smack.py
michael-emmi: Yes, please remove these lines for now. We will deal with `pthreads` and compilation databases at a later time.

share/smack/include/pthread.h
michael-emmi: Question: why are we providing header files at all?

I would imagine that the key use case here would be that an application uses `pthreads` and that we want to provide our false implementation of `pthreads` in place of the real one. In that case, there ought to be no change to the existing `pthreads` header files. The only change would be against which `pthreads` implementation we link. Right?

I think it would be best not to include these header files in SMACK.

bin/smack.py
MontyCarter: `--context-bound` sounds like a fine solution - works for me.  Let's go with this.

I agree with you that `--context-switches` is mislabeled if its default value for sequential programs is 1.

I'll fix this in my next commit.

bin/smack.py
MontyCarter: I agree - see above.

I'll patch this up with my next commit.

share/smack/include/pthread.h
MontyCarter: Mike, to fill you in, Zvonimir and I talked a bit about this, and we didn't really come up with a clear winner...  

The pthread specification only specifies an API and behavior for the API functions.  The internal workings and data structures are left up to the individual implementation.  This means, for example, that struct fields on `pthread_t` objects can be named differently across different implementations.  As a result, our implementation in `pthread.c` can't assume a particular layout or field names for the pthread structs needed to track the status of things.

Further, the system provided header files are more complex, both in terms of struct sizes/complexity and in terms of nested headers/dependencies.  Some of these nested includes can bring in dozens of other header files that wouldn't be used by our implementation.  I don't see this item as being inherently problematic, but rather just unnecessary (and possibly performance degrading).

With that said, our workaround does have some issues. We are preempting the system's `pthread.h` file by passing the path to our version of `pthread.h` on the clang command line.  However, `pthreadtypes.h` is not consistently named across various implementations.  (Sometimes the statement is `#include <pthreadtypes.h>, while other systems may use `#include <bits/pthreadtypes.h>`, while still others may use different folder structures/naming conventions).  As a result, we either need to have a copy of our pthread typedefs in the include paths expected by each implementation, or we need to "preempt" the inclusion of the systems pthread typedefs.

I used include guards to prevent the pthread structs from being defined twice (see https://github.com/smackers/smack/blob/94617b24f5c8579ae8b58e5969bc4e7db9f7d5ae/share/smack/include/pthreadtypes.h#L2 and https://github.com/smackers/smack/blob/94617b24f5c8579ae8b58e5969bc4e7db9f7d5ae/share/smack/include/pthreadtypes.h#L3).

These include guards work, but only if `pthread.h` is included in the user's program *before* any other header that recursively includes the pthread typedefs.  If header `foo.h` is included *before* `pthread.h` in the user's program, and `foo.h` recursively includes `pthreadtypes.h` somehow, then the data structures defined in the system's version of `pthreadtypes.h` will be used, which breaks our implementation in `pthread.c`.

Moving to your solution would do away with this problem, however, we would have to address the issue of unpredictable struct fields (and possibly somehow mitigate the inclusion of unneeded header files).

I am definitely open to any suggestions to address this problem - we just tucked it under the rug so I could make progress on the implementation itself.

I hope this has made sense.  It gets a little bit cumbersome to describe this problem.

share/smack/lib/pthread.c
zvonimir: Why Mike?
I find it strange to include anything but the respective header file in .c files. At least that is the convention that I am used to. Everything else should be included from the header file, no?

michael-emmi: I don’t think header files (`.h` files) should be considered as specifications of what *resources* implementation files (`.cpp` files) use. I think header files should be considered as *interface* specifications: they should specify what the implementations of that interface provide. The implementation files themselves ought to include any additional required resources.

Consider a module `B.cpp` which includes `A.h` whose methods are implemented in `A.cpp`. The `B.cpp` module needs to know about how it can interact with `A`. It does not need to know how `A.cpp` was implemented, e.g., whether it used the C library `time.h`. This is an implementation detail, that has no business being mentioned in `A.h` — unless, of course, knowledge of the interface provided by `time.h` is required for using the interface provided by `A.h`, e.g., that some of `A`’s methods have parameters with types provided by `time.h`.

Besides the conceptual idea of modularity, this difference also impacts compile-time performance: including the header files used by other modules (e.g., by `A.cpp`) when compiling any given module (e.g., `B.cpp`) increases the total number of headers included for each module, and results in slower compilation time.

It is true that we have been abusing this all along while developing SMACK. I think that was a mistake which should be corrected as we go along.

I do not know where your convention of ensuring that each implementation includes a single header file comes from, but I think it is a bad convention, and one that is certainly not followed by LLVM. Just look at any LLVM implementation file, e.g,.

  http://llvm.org/docs/doxygen/html/Function_8cpp_source.html

Further, my points are directly mentioned in the LLVM coding standards:

http://llvm.org/docs/CodingStandards.html#a-public-header-file-is-a-module
http://llvm.org/docs/CodingStandards.html#include-as-little-as-possible

Touchée! ;-)

share/smack/include/pthread.h
michael-emmi: Allow me to try and summarize what you are saying:

1. There are many implementations of the pthread API, so we cannot depend on any particular implementation details, e.g., regarding the layout of the `pthread_mutex_t` structs.

2. The pthread API headers tend to include many other headers.

Please correct me if I am wrong.

For the first issue, I don’t see what the problem is. Why couldn’t we reuse existing pthread API headers, and then link in our own implementation of the pthread API? Why should we care how any particular system defines, e.g., `pthread_mutex_t`? We could treat it as an opaque pointer if we wanted to, or we could provide our own layout — like you are doing — no? Is there any reason why linking in our own implementation of the pthread API against the standard pthread headers would not succeed?

The only problem I can imagine is if pthread API clients were to break the API abstraction, e.g., by accessing platform-specific struct layouts. Does this happen often? My advice would be to suppose it does not really happen, and say that your implementation would be unsound if it did.

For the second issue, is it really a problem? As far as I am aware, including additional *code* — i.e., procedure implementations — can be costly at analysis time. What would go wrong if we were to include a whole bunch of additional *headers* though?

I think that sticking to original pthread headers is more practical and user friendly. If we were to write programs / projects from scratch, there is no problem to include your custom pthread API headers. However, I am guessing we want to analyze programs that are already written, and compile against some system headers. In that case compiling against your custom headers might fail, as you have already noticed via altering the include order.

I think the right thing to do is to try and avoid providing custom pthread API headers, avoid modifying the compilation process, and simply link our own pthread API implementation instead of the system’s.

share/smack/lib/pthread.c
zvonimir: Good enough :). I see your point.

share/smack/include/pthread.h
zvonimir: Let me jump in here.
Here is a simple example taken from `pthread.c` that Monty wrote:
```
pthread_t pthread_self(void) {
  int tmp_tid = __VERIFIER_nondet_int();
  __SMACK_code("call @ := corral_getThreadID();", tmp_tid);
  // Print actual tid to SMACK traces
  int actual_tid = tmp_tid;
  return actual_tid;
}
```
So Monty uses `pthread_t` as a very convenient place to store a thread id, and hence he defined this type to be an integer. Now, how `pthread_t` is actually defined is implementation-dependent, and for some pthread implementations it is in fact a struct for example. If you wanna be really crazy you can probably happily define it to be a float for example. If such headers would be included, then Monty's code above would not work, right?

Now, this is actually related to a major design decision we had to make early on - should we write models in Boogie or in C. I thought that having models be written in C as much as possible is the way to go, and I still do. Are you of a different opinion?

michael-emmi: I still do not understand where the fundamental conflict arises. Here ares some concrete questions; please be mindful of the perhaps subtle distinction I am trying to make between the pthread API *headers* and *implementations* — including the possibility that the line between the two is blurred:

1. Do the *headers* of the pthreads API determine the layout of the `pthread_t` type? Or is that determined by pthreads API *implementations*? Or are the *headers* dependent on the *implementation*? In other words, is the layout of `pthread_t` type necessarily fixed when modules are compiled against the pthread API headers? Are there no “generic” pthread API headers to compile against? (On asking this, I realize that the pthread API headers themselves must be implementation dependent, but I want to be sure we are clear on these points.)

2. Now, even supposing we are forced to determine the layout of pthread API types, e.g., `pthread_t`, at compile time, does anything prevent us from linking a module which implements the pthread API with possibly conflicting pthread API types?

For example, suppose the pthread API headers we compile our program against on a given system say that `pthread_t` is 20 bytes long. Then suppose Monty’s implementation — using custom pthread API headers — says that `pthread_t` is 4 bytes long. Let’s suppose for the moment that Monty’s `pthread_t` is smaller than or equal to in size than the program’s. If we were then to link Monty’s implementation with that program, what goes wrong? Perhaps we don’t want to execute it, but from the standpoint of our Boogie encoding, is there any problem with it?

I have tried such type-incompatible linking out just to see what happens, and so far I don’t see the obstacle. Consider the following code:
````
// a.h
struct S {
  int a1;
  int a2;
};

// b.h
struct S {
  int b1;
};

// a.c
#include "a.h"
#include "stdio.h"
void foo(struct S);
int main(void) {
  struct S s;
  foo(s);
  printf("World!\n");
}

// b.c
#include "b.h"
#include "stdio.h"
void foo(struct S s) {
  printf("Hello!\n");
  s.b1 = 1;
}
````
This all happily compiles, links, and executes.
````
$ clang -c -emit-llvm a.c
$ clang -c -emit-llvm b.c
$ llvm-link a.bc b.bc -o c.bc
$ clang c.bc 
$ ./a.out 
Hello!
World!
````

So why are we insisting on fixing the pthread API headers that the program compiles against?

(I am not necessarily suggesting, yet, that there is not reason. It is just that I do not yet see the reason.)

zvonimir: Your question 1: Ah, I see now where the mix-up is coming from. Vendors provide implementations of pthreads API spec. The spec does not prescribe what e.g. `pthread_t` should be mapped to. So vendors' implementations consist of both header files and the respective C implementations. Vendors define all types as they wish.

Your question 2: I honestly had no idea that kind of linking would work at all without throwing any errors or at least warnings. I have to say it looks a bit worrisome, but maybe it would work, not sure. At the very least it might confuse DSA if the types do not match and things will get collapsed. I have to think a bit more about this. In the meantime, Monty could also test it on an actual pthreads example.

michael-emmi: My guess would be that DSA nodes — and only those corresponding to pthread API data — would collapse only if the client program didn't treat those types opaquely, e.g., trying to access into `pthread_t` with its own (possibly incorrect) knowledge of the types. As far as I am aware, DSA attempts to infer node structure from the actual program accesses. And I am supposing that client programs do treat those types like `pthread_t` opaquely.

Anyways, I guess it shouldn’t be too hard to try this out. And if it allowed us to avoid that delicate front-end header stuff, perhaps it is worth it, no?

zvonimir: Yep, we should try it out. It looks like a promising solution, but maybe there are potential issues we are not anticipating right now. I'll tell Monty to give it a try. Thanks Mike.

michael-emmi: After further reflection, I suspect that there is yet a cleaner solution, which is a variation of the original solution. (Or perhaps it was even the original solution, but I did not understand it correctly.)

Monty says: “We are preempting the system's pthread.h file by passing the path to our version of pthread.h on the clang command line,” and that “include guards work, but only if pthread.h is included in the user's program before any other header that recursively includes the pthread typedefs.”

First of all, I don’t see in this pull request where the “[preemption of] the system’s pthread.h” is happening. Where is that actually happening?

However, in principle, shouldn’t it be possible to simply ensure that:

1. our pthread API headers appear in the include paths of all compilation commands, before default system headers, and

2. our pthread API implementation module is linked in?

I would expect the first to appear as a modification to our standard clang compilation command, with some additional flags to specify the path to our headers. If this were done, why would we have to worry about recursive includes and the possibility that default system pthread API headers might be pulled in? Clang’s include path would ensure that our pthread API headers are pulled in, wouldn’t it? Or is it not possible to preempt default system headers in this way?

After all, I think this method would be superior to the type-incompatible linking solution, if you can get rid of the caveat of include frailty.

michael-emmi: Could it be that Monty is in fact including the system `pthread.h` headers when he writes `#include <pthread.h>` — and not his own custom `pthread.h` header? I suspect so. That would also explain the order dependency on `#include "pthreadtypes.h"`. If this is the case, then his implementation actually depends on a pthread API implementation which has a file called `pthreadtypes.h` with certain properties. This seems to be very platform dependent.

Note: this is what happens on OS X:
````
$ find /usr/include -name "pthread.h"
/usr/include/pthread/pthread.h
/usr/include/pthread.h
$ find /usr/include -name "pthreadtypes.h"
````
Which means that the current solution is not portable.

If we were simply to ensure that `#include <pthread.h>` pulled in Monty’s version of `pthread.h`, then we would have a clean, portable solution. And no need for this non-portable `pthreadtypes.h` stuff.

zvonimir: We have this line:
https://github.com/smackers/smack/blob/pthreadsFeature/bin/smack.py#L252
Using `-I` appends the folder to the beginning of the compiler's search list, and so Monty's `pthread.h` should be picked up before the system's `pthread.h`. This part makes sense to me and I don't see why it would not work. We should test it be absolutely sure though.
Now, I am confused about the `pthreadtypes.h` part. Do user files ever include `pthreadtypes.h` or is it included just from `pthread.h'? It seems that Monty hinted that: (a) user files do include it and on top of that (b) they include it using implementation dependent paths/names. Is that right? Monty, could you maybe point out some example?

michael-emmi: Are we targeting portable code that uses the pthread API — i.e., does not depend on platform specific pthread API headers? (I suppose we are.)

Is `pthreadtypes.h` part of the pthread API specification? (I suppose it is not.)

If our `pthread.h` header were being picked up by Clang, instead of the default system `pthread.h` header, then how would it be possible for “other header[s] [to] recursively include the [system] pthread typedefs?”

Something does not make sense here.

michael-emmi: Ok, I think I finally see what the issue is: that other system headers (i.e., other than `pthread.h`) may provide conflicting type definitions for pthread API types, e.g., in `#include <sys/types.h>`. (I guess that is what was originally meant by “recursively [including] pthread typedefs.”)

But could there be a portable way to handle this? I doubt the current solution of hijacking `pthreadtypes.h` will work across platforms, since `pthreadtypes.h` is a very platform dependent entity. I could be wrong though.

So it seems that fixing pthread API types to our liking in a robust manner is hard. Hijacking system definitions does not seem portable, nor universally feasible. For example the `<sys/types.h>` header on OS X defines `typedef __darwin_pthread_t pthread_t` guarded by the `_PTHREAD_T` and `__POSIX_LIB__` macros. Perhaps defining the `_PTHREAD_T` and `__POSIX_LIB__` macro would prevent this definition, but would it prevent all possible definitions of `pthread_t` in all possible OS X system headers? I wonder if there could be a set of macros, e.g., including `_PTHREAD_T` for OS X and `PTHREADTYPES_H` on Linux, that would ensure that no system header defines `pthread_t`?

zvonimir: Ok, I think I now as well finally understand what the issue is :): non-standard `pthreadtypes.h` occasionally gets included from "weird" places. I am really not sure what a portable solution would be.
To me it seems that we can either:
1. Use system-wide pthreads type definitions and just perform linking of our models (i.e., C files).
2. Stick more or less with what we have now and try to figure out a reasonable set of "guard macros".
I am somehow a bit worried that option (1) would create problems that we have not even envisioned, in addition to the ones mentioned above. Not sure though.
Which one should we go with? Opinions?
Note that we are also maybe over-analyzing this a bit since our current pthreads benchmarks are rather smallish. If we ever get into checking large pthreads projects, we can revisit this discussion. At that point we would also have a concrete project to go after to drive these decisions.

michael-emmi: I also think that Option 1 is risky, even if it should happen to work :-)

But we still have a bit of investigation to do in order to determine whether we could make Option 2 portable. So far we have identified a small set of potential guard macros for which this might work on whatever version of Linux you are using, and my version of OS X (Yosemite). But is there any way of knowing whether those same guard macros would also work with older (or newer) versions of Linux and OS X? How about other systems, like Cygwin, Free BSD, etc. If we could at least determine that those headers are fairly stable between system versions, then perhaps maintenance would be feasible. Otherwise things could get tedious.

My key concern at this point is to avoid baking in platform dependencies to SMACK.

zvonimir: I was just looking at Cygwin headers related to this. I can kind of see which macros guard them, but it is totally not clear that the same macros are not used at some other key locations since they do not seem to be pthreads-specific.
Here is another proposition - Monty, who most likely will not like this idea :), can try to rewrite our model so that nothing gets stored in C structures. Basically, all state would be tracked using custom Boogie maps. I think that should be doable, and I also think that it should be portable since we would not have to redefine C types - I think. There are two drawbacks that I can think of: rewriting everything is probably a good amount of work and creating all these maps within Boogie might potentially prevent some opportunities for scalarization since DSA would not kick in.
Other than that. I am out of ideas. Thoughts?

zvonimir: Actually, I realized last night that putting everything into Boogie as I suggested above potentially has its own issues that would be nice to avoid. The issues I envision are related to manipulating pthreads structures outside of the API invocations. For example, can `pthread_mutex_t` structure be copied using `memcpy`? Seems crazy, but I guess it can be done, right? Things like this could potentially break a purely Boogie model since we would lose track of such relationships. I will meet with Monty this week to discuss this.

PR #116, opened 2015-07-30
Modified the build and test scripts so that test failures are reported.

jon-whit:
Before these changes Jenkins reported builds as successful even if there were tests
that timed out/failed. This fixes that issue by setting the shell return status
based on the results of the regressions tests. Jenkins will now report a build as
failed if the shell return status is anything other than 0.

### DIFF
--- a/test/regtest.py
+++ b/test/regtest.py
@@ -239,5 +239,10 @@ def main():
     logging.info(' TIMEOUT count: %d' % timeouts)
     logging.info(' UNKNOWN count: %d' % unknowns)
 
+    # if there are any failed tests or tests that timed out, set the system
+    # exit code to a failure status
+    if timeouts > 0 or failed > 0:
+        sys.exit(1)
+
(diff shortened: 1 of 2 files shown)


test/regtest.py
```diff
@@ -239,5 +239,10 @@ def main():
     logging.info(' TIMEOUT count: %d' % timeouts)
     logging.info(' UNKNOWN count: %d' % unknowns)
 
+    # if there are any failed tests or tests that timed out, set the system
+    # exit code to a failure status
+    if timeouts > 0 or failed > 0:
+        sys.exit(1)
```
zvonimir: Indentation seems to be off by 1 here - it should be 2 spaces instead of 3. Please make this minor fix.

jon-whit: Fixed. Thanks for pointing that out!

jon-whit:
Fixed. Thanks for pointing that out!

PR #119, opened 2015-08-07
Added a test showing string translation issue. Also fixed path in Makefile

alexandrugyori:
Added small test showing that current handling of strings causes problems in the translated bpl.

The assertion  assert(s1!=retStr()); can be proved although the two string literals are the same.

Using a hash of the string literal when generating the boogie variable would address the issue I think.

### DIFF
--- a/examples/simple-project/Makefile
+++ b/examples/simple-project/Makefile
@@ -1,5 +1,5 @@
 # Set this variable to point to folder share of your SMACK installation
-INSTALL_SHARE = ../../../install/share
+INSTALL_SHARE = ../../share
 
 CC = clang
 LD = llvm-link
--- a/examples/simple-project/incr.c
+++ b/examples/simple-project/incr.c
@@ -9,3 +9,7 @@ int incr(int x) {
   return x + y;
 }
 
+char* retStr() {
+  char* s1 = "1string";
+  return s1;
+}
--- a/examples/simple-project/simple.c
+++ b/examples/simple-project/simple.c
@@ -10,10 +10,17 @@ void test(int a) {
   assert(a > b);
 }
 
+void test2() {
+  char* s1 = "1string";
+  assert(s1!=retStr());
+
+}
+
 int main(void) {
   int a = __VERIFIER_nondet();
   assume(a < 10000); // prevents overflow when bit-precise
   test(a);
+  test2();
   return 0;
 }
 


michael-emmi:
The `simple-project` code is not to be modified — though we will accept your `simple-project/Makefile` fix, in a subsequent pull request. Please submit broken examples as a regression test, e.g., in the `test/simple` directory, or simply by typing the code into a new Github issue.

We are unable to reproduce the supposed string translation issue on the following example.
````
char* str() {
  char* s = "1string";
  return s;
}

int main(void) {
  char *s = "1string";
  assert(s == str());
  return 0;
}
````
The assertion passes, as expected, and our translation gives the same reference to both pointers `s` and the return value of `str`.

**Comment by alexandrugyori** [comment]
Yes, your example will be ok if both functions are in the same file, because clang will be nice and will give you the same symbol. But if the two functions are in different files, clang will not be so nice, as I mentioned in an email to Zvonimir. Sorry, I should have mentioned  in the pull request that putting the functions in different files is critical to reproduce the bug.

I did not add it as a test mainly because it seemed that all tests are single file tests, but maybe I missed something. Could you try to reproduce the issue on an example with multiple files and let me know if you are unable to do so?

**Comment by michael-emmi** [comment]
Oh now I see. The issue is that a normal linker (`ld`) seems to identify two identical string constants as one, whereas `llvm-link` does not. Will consider fixes.

**Comment by michael-emmi** [comment]
Actually, does C semantics specify that pointer comparison between two identical string literals should hold?

**Comment by michael-emmi** [comment]
Just to be clear here, what is the target semantics here? Aren’t we in the space of undefined semantics? I am guessing that C language specifications would not insist that the pointers to two identical string constants be equal. Therefore I am hesitant to insist that equality in our encoding.

Is your point that they should not necessarily be unequal, either?

**Comment by alexandrugyori** [comment]
I am not sure about the C semantics, but I would expect that I cannot prove them unequal. While not being able to prove equality may be ok, proving the unequal seems wrong to me.

I believe an easy-ish fix would be to consider the value of the string literal in the translation to boogie. That would help me a lot, since we're using smack for differential program verification, where we build two boogie programs independently from two versions of a program. In that case, if the value of the string literal changed between the two versions of a program I will not see any change in the resulting bpl. If the value of the literal would be considered when producing the bpl, then I will see the difference in the boogie programs.

**Comment by michael-emmi** [comment]
Would it help you if we were to simply output the value of the string literal(s) in the Boogie translation? e.g.,
````
const {:string "1string"} .str: ref;
````

Considering that distinct instances of identical string literals are stored at different addresses does not particularly bother me. This semantics is probably unspecified anyways, and strings are not supposed to be compared by their addresses. This is compiler-specific territory. There are plenty of places where SMACK makes assumptions that are consistent with C semantics, yet might be inconsistent with those of some C compilers. I don’t see any good reason to follow any particular compiler’s choice on this one.

**Comment by michael-emmi** [comment]
To respond more directly to your question though: there is a reason we are hesitant to include the static initializers for strings. Our experience is that verifier scalability suffers with a large number of static initializers. And that verification queries typically do not rely on the values of string literals.

Thus if the simple {:string "1string"} annotation will help you, this may be a better solution.

If that does not help, we can attempt to include static initializers for strings, possibly as an option which is disabled by default in case scalability is still an issue.

Let us know :-)

**Comment by shuvendu-lahiri** [comment]
I think it is fine for now. 

The issue for us in SymDiff was that the constant address created for a string remains the same across two versions of a program, even if the string contents are different. So, we can prove the following programs equivalent (if they start with the same value of $M.xx maps), if I am not wrong:

```
foo(){return "a";}; //procedure foo(){ret := $ref.sub(0,4);}
foo'(){return "b";}//procedure foo'(){ret := $ref.sub(0,4);}
```
Not a very pressing issue for now.

Issue #125, opened 2015-09-10
Allow project building to recover from errors (Vagrant)


cogumbreiro:
Currently, building Smack through Vagrant is hit and miss experience as any error means that the build  process must be done from scratch.

It would be nice to let  `build.sh` be more resilient to errors: let it skip steps performed correctly so that it can resume building from where it last encountered an error.

zvonimir:
We implemented this.

PR #126, opened 2015-09-10
Made `build.sh` more resiliant, letting it resume after error.

cogumbreiro:
This is an attempt at handling issue #125.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -219,14 +219,18 @@ then
     ;;
 
   linux-ubuntu-14*)
-    sudo add-apt-repository "deb http://llvm.org/apt/trusty/ llvm-toolchain-trusty-3.5 main"
-    ${WGET} -O - http://llvm.org/apt/llvm-snapshot.gpg.key | sudo apt-key add -
-    sudo apt-get update
-    sudo apt-get install -y ${DEPENDENCIES}
-    sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-3.5 20
-    sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-3.5 20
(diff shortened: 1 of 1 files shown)


zvonimir:
Sorry about the delay. I've been super busy with an NSF large proposal (deadline is today).
I like your improvement, and I think this should probably be the default behavior. I certainly often manually turn off installation of a bunch of things using flags.
However, it would be great to be able to turn it off if and when needed. Could you maybe introduce a flag, maybe something like `--force-install` or something like that, which would overwrite the default behavior?
Thanks and it is great to have you (somewhat :)) on board!

cogumbreiro:
Don't worry about the delay and thanks for accepting the push request.

Sure, I can introduce that flag that ignores the checks.

cogumbreiro:
What do you think about calling the switch `--force-build` or `--skip-build-cache`?

My motivation for this is that the calling it `--force-install` might make sure the dependencies are installed no matter what, where what it actually does disabling some sort of build-cache.

Let me know which name you prefer `--force-install`.

zvonimir:
I think that `--force-build` sounds good to me. You will have to resolve conflicts, but that should be relatively easy I think. Thanks a lot!
Side note: for some reason I occasionally do not get email notifications when people post comments to issues and pull requests on github. Hence I completely missed your comment. I still have to get to the bottom of this.

michael-emmi:
Any reason this pull request is still hanging?

zvonimir:
Well, the `force-build` flag we discussed above was not implemented, and so this never got merged. I still like the addition though since I often manually turn off installation of dependencies that I know I installed previously. However, silently skipping dependencies is probably not a good idea either.
I think that the right thing to do would be to issue a warning/note if installation of a dependency is skipped since it is already present (instead of implementing the `--force-build` flag).

zvonimir:
We implemented a version of this.

PR #135, opened 2015-10-15
Add a script to classify Floats set

Guoanshisb:
- A script to classify Floats category. 
- Works for all categories except for Termination-crafted.set
- Based on grep, AST is not required.


zvonimir:
Please rename the file into `svcomp_filters.py`. I think it would be good to put both floats and bit-vectors filters into the same file.

zvonimir:
Man, it seems you did not rename it, but rather you added another file. You should use `git mv`. Did you use that?

Guoanshisb:
I just realized that. Already done.

2015-10-14 21:53 GMT-07:00 Zvonimir <notifications@github.com>:

PR #136, opened 2015-10-17
Add bv-vector classifier

Guoanshisb:
Add bit vector classifier.

### DIFF
--- a/bin/smack.py
+++ b/bin/smack.py
@@ -329,8 +329,11 @@ def svcomp_frontend(args):
   if len(args.input_files) > 1:
     raise RuntimeError("Expected a single SVCOMP input file.")
 
-  # test float benchmarks
-  if float_filter(args.input_files[0]):
+  # test float\bv benchmarks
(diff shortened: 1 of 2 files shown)


bin/smack.py
```diff
@@ -329,8 +329,11 @@ def svcomp_frontend(args):
   if len(args.input_files) > 1:
     raise RuntimeError("Expected a single SVCOMP input file.")
 
-  # test float benchmarks
-  if float_filter(args.input_files[0]):
+  # test float\bv benchmarks
+  filter_status = svcomp_filter(args.input_files[0])
+  if filter_status == 1:
```
zvonimir: Shaobo, instead of returning integers 1 or 2, please return something more descriptive than that. This is Python and not C :). You can even return 'float' and 'bitvector' strings or something like that. Please fix this and pull again.

Guoanshisb: I have fixed it. Please take a look at it.

PR #138, opened 2015-10-21
Fix link error by renaming defined exit functions

Guoanshisb:
- Fix issues with run http://pc352.emulab.net/results.py?category=HeapManipulation&--verifier=svcomp&Submit=Submit
- Need discussion on creating temporary files

### DIFF
--- a/bin/smack.py
+++ b/bin/smack.py
@@ -325,6 +325,20 @@ def json_compilation_database_frontend(args):
 
   llvm_to_bpl(args)
 
+def svcomp_process_file(args, name, ext):
+  # create a temp file (TODO: current dir or example dir)
+  tmp_file = temporary_file(name, ext, args)
+  # replace exit definition with exit_
+  with open(args.input_files[0], 'r') as fi:
+    s = fi.read()
+  s = re.sub(r'''void\s+exit\s*\(int s\)''',
+        r'''void exit_(int s)''',
+        s)
+  with open(tmp_file, 'w') as fo:
+    fo.write(s)
+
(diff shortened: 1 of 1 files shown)


bin/smack.py
```diff
@@ -325,6 +325,24 @@ def json_compilation_database_frontend(args):
 
   llvm_to_bpl(args)
 
+def svcomp_process_file(args, name, ext):
+  # create a temp file (TODO: current dir or example dir)
+  tmp_file = temporary_file(name, ext, args)
```
zvonimir: Please move this down to where your temp file is actually being open.

bin/smack.py
```diff
@@ -325,6 +325,24 @@ def json_compilation_database_frontend(args):
...
+  # replace exit definition with exit_
+  with open(args.input_files[0], 'r') as fi:
+    s = fi.read()
+  s = re.sub(r'''void\s+exit\s*\(int s\)''',
```
zvonimir: Why do you have 3 quotation marks around strings?

bin/smack.py
```diff
@@ -325,6 +325,24 @@ def json_compilation_database_frontend(args):
...
+        r'''void exit_(int s)''',
+        s)
+  # replace all occurrences of 100000 with 10
+  s = re.sub(r'''100000''',
```
zvonimir: This can all be on one line of code.

bin/smack.py
zvonimir: Put this in one line of code. It won't be too long.

PR #153, opened 2016-02-19
Added memory safety pass

jiten-thakkar:
Added LLVM pass which calls smack memory check function before every Load and Store instruction calls. Smack memory check function inserts boogie predicates to check if memory being accessed is allocated and there is no buffer overflow.


CMakeLists.txt
```diff
@@ -158,6 +158,7 @@ add_library(smackTranslator STATIC
   include/smack/SimplifyLibCalls.h
   include/smack/SmackRep.h
   include/smack/SmackRepFlatMem.h
+  include/smack/MemoryAllocationChecker.h
```
zvonimir: Rename this into `MemorySafetyChecker`

lib/smack/MemoryAllocationChecker.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include "llvm/IR/Module.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/Support/Debug.h"
+
+namespace smack {
+  using namespace llvm;
+
+  const char* VERIFIER_FUNCTION = "__SMACK_check_memory_access";
```
zvonimir: Move this into `Naming.h` and `Naming.cpp`. Also, use std::string instead of char.

lib/smack/MemoryAllocationChecker.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+  bool MemoryAllocationChecker::runOnModule(Module& m) {
+    DataLayout* dataLayout = new DataLayout(&m);
+    Function* verifierFunction;
+    for(auto& f:m) {
```
zvonimir: Module has a `getFunction` function. So you can probably use that instead of this loop. You should report an assertion violation if this function is not found.

lib/smack/MemoryAllocationChecker.cpp
```diff
@@ -0,0 +1,54 @@
...
+      if(f.getName() == VERIFIER_FUNCTION) {
+        verifierFunction = &f;
+        break;
+      }
+    }
+    for (auto& F : m) {
+      if(!Naming::isSmackName(F.getName())) {
+        for (auto& B : F) {
```
zvonimir: Could you get rid of the loop that goes over basic blocks by iterating directly over instructions?
See: http://llvm.org/docs/ProgrammersManual.html#iterating-over-the-instruction-in-a-function

share/smack/include/smack.h
```diff
@@ -35,6 +35,8 @@ smack_value_t __SMACK_return_value(void);
 // Sugar for __SMACK_init_func_XXX()
 #define __SMACK_INIT(x) void __SMACK_init_func_##x()
 
+void __SMACK_check_memory_access(void*,long int);
```
zvonimir: Please add a short comment before this function saying what its purpose is.

share/smack/lib/smack.c
```diff
@@ -1006,6 +1008,19 @@ void __SMACK_decls() {
 #undef D
 }
 
+void __SMACK_check_memory_access(void* pointer, long int size) {
+  #if MEMORY_MODEL_NO_REUSE_IMPLS
+  #elif MEMORY_MODEL_REUSE // can reuse previously-allocated and freed addresses
+    __SMACK_code("assert $slt.ref.bool(@,$0.ref) || $Alloc[$base(@)] == true;", pointer, pointer);
+    __SMACK_code("assert $slt.ref.bool(@,$0.ref) || $base(@) <= @;", pointer, pointer, pointer);
+    __SMACK_code("assert $slt.ref.bool(@,$0.ref) || @+@ <= $base(@) + $Size[$base(@)];", pointer, pointer, (size+7)/8, pointer, pointer);
```
zvonimir: Why do you have here (size+7)/8?

tools/llvm2bpl/llvm2bpl.cpp
```diff
@@ -122,6 +123,7 @@ int main(int argc, char **argv) {
   pass_manager.add(new llvm::MergeArrayGEP());
   // pass_manager.add(new smack::SimplifyLibCalls());
   pass_manager.add(new llvm::Devirtualize());
+  pass_manager.add(new smack::MemoryAllocationChecker());
```
zvonimir: This should not be enabled by default. You should introduce a command line option that enables it. See how this is done with `StaticUnroll` flag. Call this option e.g. `MemorySafety`. You should also modify the SMACK python script to include this option.

lib/smack/MemoryAllocationChecker.cpp
```diff
@@ -0,0 +1,54 @@
...
+          for (auto& I : B) {
+            if(LoadInst* li = dyn_cast<LoadInst>(&I)) {
+              Value* pointer = li->getPointerOperand();
+              Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64), dataLayout->getTypeSizeInBits(dyn_cast<PointerType>(li->getPointerOperand()->getType())->getPointerElementType()), false);
+              Type *PtrTy = PointerType::getUnqual(IntegerType::getInt8Ty(F.getContext()));
```
zvonimir: Rename this variable into e.g. `ptrType`.

share/smack/lib/smack.c
jiten-thakkar: The size parameter represents number of **bits** that are being accessed. Whereas when memory is allocated in, malloc function in boogies is passed number of bites. In order to make it a multiple of 8, I am adding 7 and then dividing it by 8. For ex. suppose size = 61. In this case number of bytes would be (61+7) / 8 = 8 which is correct. Now suppose size = 64. In this case also (64+7)/8 = 8. 
The other option is that I multiply 8 with $Size[$base(@)].

tools/llvm2bpl/llvm2bpl.cpp
zvonimir: This should not be enabled by default. You should introduce a command line option that enables it. See `StaticUnroll` for an example of how to do that. You should also add a command like option into the SMACK python script. Maybe call it `--memory-safety`.

share/smack/include/smack.h
zvonimir: Add a comment just before this briefly saying what this function does, e.g., it is used for generating memory safety checks.

lib/smack/MemoryAllocationChecker.cpp
zvonimir: Rename this variable into e.g. `ptrType`.

tools/llvm2bpl/llvm2bpl.cpp
jiten-thakkar: @zvonimir  Should I choose reuse or no-reuse as memory model when memory safety is enabled? if no memory model is chosen then the default one in impl which doesn't check memory safety

jiten-thakkar:
Closing this request to issue a new pull request.

PR #154, opened 2016-03-08
Memory Safety Checker

jiten-thakkar:


jiten-thakkar:
@Guoanshisb hey shaobo, I need to merge these changes. Any reason for closing the pull request?

include/smack/MemorySafetyChecker.h
```diff
@@ -0,0 +1,18 @@
+//
+// This file is distributed under the MIT License. See LICENSE for details.
+//
+#ifndef MEMORYSAFETYCHECKER_H
+#define MEMORYSAFETYCHECKER_H
+
+#include "llvm/Pass.h"
+
+namespace smack {
+  class MemorySafetyChecker: public llvm::ModulePass {
```
zvonimir: Indentation is wrong here. See our other header files and fix this.

include/smack/MemorySafetyChecker.h
```diff
@@ -0,0 +1,18 @@
...
+    public:
```
zvonimir: Wrong indentation again.

include/smack/Naming.h
```diff
@@ -78,6 +78,7 @@ class Naming {
   static const std::string PTR_VAR;
   static const std::string UNDEF_SYM;
   static const std::string CONTRACT_EXPR;
+  static const std::string VERIFIER_FUNCTION;
```
zvonimir: Make this variable name more descriptive. Maybe `MEMORY_CHECK_FUNCTION`.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,46 @@
+//
+// This file is distributed under the MIT License. See LICENSE for details.
+//
+#include "smack/MemorySafetyChecker.h"
+#include "smack/Naming.h"
+
```
zvonimir: No need for this empty line.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,46 @@
...
+  bool MemorySafetyChecker::runOnModule(Module& m) {
+    DataLayout* dataLayout = new DataLayout(&m);
+    Function* verifierFunction = m.getFunction(Naming::VERIFIER_FUNCTION);
+    assert(verifierFunction != NULL && "Couldn't find memory safety function");
+    for (auto& F : m) {
+      if(!Naming::isSmackName(F.getName())) {
+        for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
+          if(LoadInst* li = dyn_cast<LoadInst>(&*I)) {
+            Value* pointer = li->getPointerOperand();
+            Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64), dataLayout->getTypeSizeInBits(dyn_cast<PointerType>(li->getPointerOperand()->getType())->getPointerElementType()), false);
```
zvonimir: Break up this line.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,46 @@
...
+            Type *PtrTy = PointerType::getUnqual(IntegerType::getInt8Ty(F.getContext()));
+            CastInst* castPointer = CastInst::Create(Instruction::BitCast, pointer, PtrTy, "", &*I);
+            Value* args[] = {castPointer, size};
+            CallInst::Create(verifierFunction, ArrayRef<Value*>(args, 2), "", &*I);
+          } else if (StoreInst* si = dyn_cast<StoreInst>(&*I)) {
+            Value* pointer = si->getPointerOperand();
+            Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64), dataLayout->getTypeSizeInBits(si->getValueOperand()->getType()), false);
```
zvonimir: Break up this long line.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "llvm/IR/InstIterator.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/Support/Debug.h"
+
+namespace smack {
+  using namespace llvm;
+
+  bool MemorySafetyChecker::runOnModule(Module& m) {
+    DataLayout* dataLayout = new DataLayout(&m);
+    Function* verifierFunction = m.getFunction(Naming::VERIFIER_FUNCTION);
```
zvonimir: Rename the variable `verifierFunction`.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,46 @@
...
+    assert(verifierFunction != NULL && "Couldn't find memory safety function");
+    for (auto& F : m) {
+      if(!Naming::isSmackName(F.getName())) {
+        for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
+          if(LoadInst* li = dyn_cast<LoadInst>(&*I)) {
```
zvonimir: See our other code for some coding conventions. Put space between `if` and `(`.

share/smack/lib/smack.c
```diff
@@ -1006,6 +1008,25 @@ void __SMACK_decls() {
 #undef D
 }
 
+//The size parameter represents number of bits that are being accessed.
```
zvonimir: Add space after `//`.

share/smack/lib/smack.c
```diff
@@ -982,6 +982,7 @@ void __SMACK_decls() {
 
 #else // NO_REUSE does not reuse previously-allocated addresses
   D("var $Alloc: [ref] bool;");
+  D("function $Size(ref) returns (ref);");
```
zvonimir: Wait, you did not add `$Size` to `NO_REUSE_IMPLS`. That is our main used memory model, and so please do that. You will have to introduce the appropriate constraints on base there too!

share/smack/lib/smack.c
zvonimir: Add space after `//`.

zvonimir: Why does it represent the number of bits???

share/smack/lib/smack.c
```diff
@@ -1006,6 +1008,25 @@ void __SMACK_decls() {
...
 }
 
+//The size parameter represents number of bits that are being accessed.
+//Whereas when memory is allocated, malloc function in boogie is passed 
+//number of bites. In order to make it a multiple of 8, I am adding 7 
+//and then dividing it by 8. For ex. suppose size = 61. In this case
+//number of bytes would be (61+7) / 8 = 8 which is correct. Now suppose
+//size = 64. In this case also (64+7)/8 = 8. 
+void __SMACK_check_memory_access(void* pointer, long int size) {
+  #if MEMORY_MODEL_NO_REUSE_IMPLS
```
zvonimir: Add the checks for this memory model too.

share/smack/lib/smack.c
```diff
@@ -1006,6 +1008,25 @@ void __SMACK_decls() {
...
+  #elif MEMORY_MODEL_REUSE // can reuse previously-allocated and freed addresses
+    __SMACK_code("assert $slt.ref.bool(@,$0.ref) || $Alloc[$base(@)] == true;", pointer, pointer);
+    __SMACK_code("assert $slt.ref.bool(@,$0.ref) || $base(@) <= @;", pointer, pointer, pointer);
+    __SMACK_code("assert $slt.ref.bool(@,$0.ref) || @+@ <= $base(@) + $Size[$base(@)];", pointer, pointer, (size+7)/8, pointer, pointer);
+  #else
```
zvonimir: Wait, the checks look the same in both memory models. Why do you have this if-elif-else split?
Are they not all going to be the same?

tools/llvm2bpl/llvm2bpl.cpp
```diff
@@ -122,6 +127,10 @@ int main(int argc, char **argv) {
   pass_manager.add(new llvm::MergeArrayGEP());
   // pass_manager.add(new smack::SimplifyLibCalls());
   pass_manager.add(new llvm::Devirtualize());
+   
+  if(MemorySafety) {
```
zvonimir: Add space after `if`.

share/smack/top.py
```diff
@@ -135,6 +135,9 @@ def arguments():
   translate_group.add_argument('--entry-points', metavar='PROC', nargs='+',
     default=['main'], help='specify top-level procedures [default: %(default)s]')
 
+  translate_group.add_argument('--memory-safety', action='store_true', default=False,
+    help='enable memory safety checks. Only reuse and no-reuse memory models are supported')
```
zvonimir: I forgot why these checks are not supported for all memory models.

share/smack/lib/smack.c
```diff
@@ -1006,6 +1008,25 @@ void __SMACK_decls() {
...
+//Whereas when memory is allocated, malloc function in boogie is passed 
+//number of bites. In order to make it a multiple of 8, I am adding 7 
+//and then dividing it by 8. For ex. suppose size = 61. In this case
+//number of bytes would be (61+7) / 8 = 8 which is correct. Now suppose
+//size = 64. In this case also (64+7)/8 = 8. 
+void __SMACK_check_memory_access(void* pointer, long int size) {
+  #if MEMORY_MODEL_NO_REUSE_IMPLS
+  #elif MEMORY_MODEL_REUSE // can reuse previously-allocated and freed addresses
+    __SMACK_code("assert $slt.ref.bool(@,$0.ref) || $Alloc[$base(@)] == true;", pointer, pointer);
+    __SMACK_code("assert $slt.ref.bool(@,$0.ref) || $base(@) <= @;", pointer, pointer, pointer);
```
zvonimir: Use `$slt.ref.bool` for the comparisons with `$base` too.

share/smack/lib/smack.c
```diff
@@ -1006,6 +1008,25 @@ void __SMACK_decls() {
...
+    __SMACK_code("assert $slt.ref.bool(@,$0.ref) || @+@ <= $base(@) + $Size[$base(@)];", pointer, pointer, (size+7)/8, pointer, pointer);
```
zvonimir: Do not use `+`. Instead use `$add.ref`.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,46 @@
...
+            Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64), dataLayout->getTypeSizeInBits(si->getValueOperand()->getType()), false);
+            Type *ptrTy = PointerType::getUnqual(IntegerType::getInt8Ty(F.getContext()));
+            CastInst* castPointer = CastInst::Create(Instruction::BitCast, pointer, ptrTy, "", &*I);
+            Value* args[] = {castPointer, size};
+            CallInst::Create(verifierFunction, ArrayRef<Value*>(args, 2), "", &*I);
+          }
+        }
+      }
+    }
+    return false;
```
zvonimir: You should return `true` since you pass updates the module.

share/smack/lib/smack.c
jiten-thakkar: @zvonimir the function getTypeSizeInBits in class DataLayout only provides size in bits for any type.

share/smack/lib/smack.c
jiten-thakkar: @zvonimir for memory reuse, $Size is an array and for memory no-reuse $Size if a function. That's the only difference in both of them.

zvonimir: Ah, I see. Ok, that makes sense.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "llvm/Support/Debug.h"
+
+namespace smack {
+
+using namespace llvm;
+
+bool MemorySafetyChecker::runOnModule(Module& m) {
+  DataLayout* dataLayout = new DataLayout(&m);
+  Function* memorySafetyFunction = m.getFunction(Naming::MEMORY_SAFETY_FUNCTION);
+  assert(verifierFunction != NULL && "Couldn't find memory safety function");
```
zvonimir: I guess you are not compiling under debug mode, since this assertion would not compile in this case :). In particular, the variable `verifierFunction` does not exist any more.

share/smack/lib/smack.c
```diff
@@ -937,10 +937,13 @@ void __SMACK_decls() {
 
 #if MEMORY_MODEL_NO_REUSE_IMPLS
   D("var $Alloc: [ref] bool;");
+  D("function $Size(ref) returns (ref);");
   D("var $CurrAddr:ref;");
 
   D("procedure $alloc(n: ref) returns (p: ref)\n"
     "modifies $CurrAddr, $Alloc;\n"
+    "ensures $Size(p) == n;\n"
```
zvonimir: We do not need these ensures, we only need the assumes below. So please remove them.

share/smack/top.py
```diff
@@ -135,6 +135,9 @@ def arguments():
   translate_group.add_argument('--entry-points', metavar='PROC', nargs='+',
     default=['main'], help='specify top-level procedures [default: %(default)s]')
 
+  translate_group.add_argument('--memory-safety', action='store_true', default=False,
+    help='enable memory safety checks.')
```
zvonimir: Make help message consistent with others. Meaning, remove the period in the end of the message. Other messages do not have it.

share/smack/lib/smack.c
```diff
@@ -1006,6 +1013,24 @@ void __SMACK_decls() {
 #undef D
 }
 
+// The size parameter represents number of bits that are being accessed.
```
zvonimir: Why is size parameter the number of bits?
Can it not be the number of bytes?
It is awkward to have this division by 8 in Boogie, and this might cause problems.

jiten-thakkar: I am afraid I couldn't find any other function in LLVM which would give me the size of a datatype other than the function getTypeSizeInBits in class DataLayout. This function provides only bits. Moreover, the division happens in C not in boogie.

lib/smack/MemorySafetyChecker.cpp
jiten-thakkar: sorry missed that one.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,49 @@
...
+  for (auto& F : m) {
+    if (!Naming::isSmackName(F.getName())) {
+      for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
+        if (LoadInst* li = dyn_cast<LoadInst>(&*I)) {
+          Value* pointer = li->getPointerOperand();
+          Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64), 
+                        dataLayout->getTypeSizeInBits(dyn_cast<PointerType>(li->getPointerOperand()
```
zvonimir: I think you should use the following LLVM function here to get the number of bytes:
http://llvm.org/docs/doxygen/html/classllvm_1_1DataLayout.html#a4b979cd93852fd44c48851c94ee00758
So than `size` will be the number of bytes (not bits), and you can remove that division from Boogie code.

share/smack/lib/smack.c
```diff
@@ -1006,6 +1011,24 @@ void __SMACK_decls() {
...
+// Whereas when memory is allocated, malloc function in boogie is passed
```
zvonimir: Update this comment based on your latest implementation. I guess you can completely remove it.

share/smack/include/smack.h
```diff
@@ -35,6 +35,10 @@ smack_value_t __SMACK_return_value(void);
 // Sugar for __SMACK_init_func_XXX()
 #define __SMACK_INIT(x) void __SMACK_init_func_##x()
 
+// Inserts memory access checks in form of assert to check null pointer access
+// and buffer overflow errors
+void __SMACK_check_memory_safety(void*,long int);
```
zvonimir: This should be `unsigned long` instead of `long int`.

share/smack/lib/smack.c
```diff
@@ -1006,6 +1011,19 @@ void __SMACK_decls() {
 #undef D
 }
 
+// The size parameter represents number of bytes that are being accessed.
+void __SMACK_check_memory_safety(void* pointer, long int size) {
```
zvonimir: This should be `unsigned long` instead of `long int`.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,49 @@
...
+  Function* memorySafetyFunction = m.getFunction(Naming::MEMORY_SAFETY_FUNCTION);
+  assert(memorySafetyFunction != NULL && "Couldn't find memory safety function");
+  for (auto& F : m) {
+    if (!Naming::isSmackName(F.getName())) {
+      for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
+        if (LoadInst* li = dyn_cast<LoadInst>(&*I)) {
+          Value* pointer = li->getPointerOperand();
+          Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64), 
+                        dataLayout->getTypeStoreSize(dyn_cast<PointerType>(li->getPointerOperand()
+                        ->getType())->getPointerElementType()), false);
```
zvonimir: Remove `false` as the argument. That is the default argument and hence you can omit it.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,49 @@
...
+          Type *PtrTy = PointerType::getUnqual(IntegerType::getInt8Ty(F.getContext()));
+          CastInst* castPointer = CastInst::Create(Instruction::BitCast, pointer, PtrTy, "", &*I);
+          Value* args[] = {castPointer, size};
+          CallInst::Create(memorySafetyFunction, ArrayRef<Value*>(args, 2), "", &*I);
+        } else if (StoreInst* si = dyn_cast<StoreInst>(&*I)) {
+          Value* pointer = si->getPointerOperand();
+          Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64), 
+                        dataLayout->getTypeStoreSize(si->getValueOperand()->getType()), false);
```
zvonimir: Remove `false` as the argument. That is the default argument and hence you can omit it.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,49 @@
...
+bool MemorySafetyChecker::runOnModule(Module& m) {
+  DataLayout* dataLayout = new DataLayout(&m);
+  Function* memorySafetyFunction = m.getFunction(Naming::MEMORY_SAFETY_FUNCTION);
+  assert(memorySafetyFunction != NULL && "Couldn't find memory safety function");
+  for (auto& F : m) {
+    if (!Naming::isSmackName(F.getName())) {
+      for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
+        if (LoadInst* li = dyn_cast<LoadInst>(&*I)) {
+          Value* pointer = li->getPointerOperand();
+          Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64),
```
zvonimir: Introduce an additional variable here to break up this very long expression.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,49 @@
...
+          Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64), 
+                        dataLayout->getTypeStoreSize(dyn_cast<PointerType>(li->getPointerOperand()
+                        ->getType())->getPointerElementType()), false);
+          Type *PtrTy = PointerType::getUnqual(IntegerType::getInt8Ty(F.getContext()));
+          CastInst* castPointer = CastInst::Create(Instruction::BitCast, pointer, PtrTy, "", &*I);
+          Value* args[] = {castPointer, size};
+          CallInst::Create(memorySafetyFunction, ArrayRef<Value*>(args, 2), "", &*I);
+        } else if (StoreInst* si = dyn_cast<StoreInst>(&*I)) {
+          Value* pointer = si->getPointerOperand();
+          Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64),
```
zvonimir: Introduce an additional variable here to break up this very long expression.

lib/smack/MemorySafetyChecker.cpp
zvonimir: It is weird to break up a line just before `->`.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -0,0 +1,49 @@
...
+  DataLayout* dataLayout = new DataLayout(&m);
+  Function* memorySafetyFunction = m.getFunction(Naming::MEMORY_SAFETY_FUNCTION);
+  assert(memorySafetyFunction != NULL && "Couldn't find memory safety function");
+  for (auto& F : m) {
+    if (!Naming::isSmackName(F.getName())) {
+      for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
+        if (LoadInst* li = dyn_cast<LoadInst>(&*I)) {
+          Value* pointer = li->getPointerOperand();
+          Value* size = ConstantInt::get(IntegerType::get(F.getContext(), 64), 
+                        dataLayout->getTypeStoreSize(dyn_cast<PointerType>(li->getPointerOperand()
```
zvonimir: Replace `li->getPointerOperand()` with your `pointer` variable!

lib/smack/MemorySafetyChecker.cpp
zvonimir: Is it safe to have 64 hard-coded here? Will this work even on 32-bit architectures?
My hunch is that you should get this number by looking at the type of the second argument of your `memorySafetyFunction`.

jiten-thakkar: Can't I just use sizeof function?

zvonimir: No, I don't think you can. The reason is quite simple - SMACK and the target application we want to verify might not be compiled on the same platform. For example, SMACK would likely be compiled on a 64-bit platform, while the target application might be compiled with `-m32` since it is supposed to be running on a 32-bit platform.
Hence, you have to get this size from the target application bitcode file, for example by looking at the size of the second argument of the memory safety function.

jiten-thakkar: We are creating this integer in the pass to save an integer value which represents the size of the data type. So it won't matter what target application is built on as long as smack is run the same platform that it was compiled on. In that case sizeof should work.

zvonimir: Is it not true that the size (i.e., bit-width) of the data type you are creating here has to match the size of the second argument of `__SMACK_check_memory_safety`?

jiten-thakkar: Yes that is true.

zvonimir: Good, good. So then does it make sense that invoking `sizeof` in `smack.c` and in `MemorySafetyChecker.cpp` on the same type might not result in the same number? Meaning, `unsigned long` in `smack.c` might be of different size than `unsigned long` in ``MemorySafetyChecker.cpp`.
This can happen since the two files might not get compiled for the same platforms.

jiten-thakkar: Ok, now I get it. So I will get that number from the second variable of memory safety function using the methods in Function class of llvm.

zvonimir: Yes, I think that is the way to go! Good job.
Btw, there are some other comments you still have not addressed. Do let me know once you are done and I'll take a look at everything.

PR #155, opened 2016-03-21
Add Symbooglix as a backend verifier

Guoanshisb:
**Symbooglix** is a symbolic execution engine for Boogie programs. So we would like to add it to our backend verifier arsenal. This pull request implements support to Symbooglix,
- Modify build script to fetch Symbooglix repo, build it, and update the environment.
- Create a binary called `symbooglix` which is similar to `boogie` and `corral`.
- Modify `top.py` to invoke Symbooglix.

The following issues are worth discussion,
- Symbooglix uses return code to indicate verification result and exceptions/errors occurred as opposed to Boogie/Corral. Details of exit codes can be found at https://github.com/symbooglix/symbooglix/blob/be450d80e744aae664e581ded9d3136e0a36b828/src/SymbooglixDriver/Driver.cs#L237. I didn't change `try_command` to process return code. Instead I did it in `verification_result` which searches for return codes while does not raise exceptions when something goes wrong for Symbooglix.
- The behavior of `--max-depth` is different from `/loopUnroll` of Boogie and `/recursionBound` of Corral. Note that Symbooglix has a special return code called `NO_ERRORS_NO_TIMEOUT_BUT_HIT_BOUND` which indicates a "unrolling" bound is reached. For some examples (e.g. `globals_fail.c`), unrolling bound `2` suffices to expose the bug for Boogie or Corral while Symbooglix returns the code mentioned above.
- The verification result when Symbooglix returns `NO_ERRORS_NO_TIMEOUT_BUT_HIT_BOUND` and `NO_ERRORS_NO_TIMEOUT_BUT_FOUND_SPECULATIVE_PATHS` is `unknown` even though the program being verified may not contain any bugs.

### DIFF
--- a/share/smack/top.py
+++ b/share/smack/top.py
@@ -235,7 +235,7 @@ def try_command(cmd, cwd=None, console=False, timeout=None):
     proc = None
     if timeout and timed_out[0]:
       return output + ("\n%s timed out." % cmd[0])
-    elif rc:
+    elif rc and args.verifier != "symbooglix":
       raise RuntimeError("%s returned non-zero." % cmd[0])
     else:
       return output
@@ -385,11 +385,11 @@ def annotate_bpl(args):
     f.write(bpl)
 
 def verification_result(verifier_output):
-  if re.search(r'[1-9]\d* time out|Z3 ran out of resources|timed out', verifier_output):
+  if re.search(r'[1-9]\d* time out|Z3 ran out of resources|timed out|ERRORS_TIMEOUT', verifier_output):
     return 'timeout'
-  elif re.search(r'[1-9]\d* verified, 0 errors?|no bugs', verifier_output):
+  elif re.search(r'[1-9]\d* verified, 0 errors?|no bugs|NO_ERRORS_NO_TIMEOUT$', verifier_output):
     return 'verified'
-  elif re.search(r'\d* verified, [1-9]\d* errors?|can fail', verifier_output):
+  elif re.search(r'\d* verified, [1-9]\d* errors?|can fail|ERRORS_NO_TIMEOUT$', verifier_output):
     return 'error'
   else:
     return 'unknown'
--- a/share/smack/top.py
+++ b/share/smack/top.py
@@ -433,7 +433,7 @@ def verify_bpl(args):
     command += ["/tryCTrace", "/noTraceOnDisk", "/useDuality", "/oldStratifiedInlining"]
     command += ["/recursionBound:1073741824", "/k:1"]
 
-  if args.bit_precise:
+  if args.bit_precise and args.verifier != "symbooglix":
     x = "bopt:" if args.verifier != 'boogie' else ""
     command += ["/%sproverOpt:OPTIMIZE_FOR_BV=true" % x]
     command += ["/%sz3opt:smt.relevancy=0" % x]
--- a/share/smack/top.py
+++ b/share/smack/top.py
@@ -138,7 +138,7 @@ def arguments():
   verifier_group = parser.add_argument_group('verifier options')
 
   verifier_group.add_argument('--verifier',
-    choices=['boogie', 'corral', 'duality', 'svcomp'],
+    choices=['boogie', 'corral', 'duality', 'svcomp', 'symbooglix'],
     help='back-end verification engine')
 
   verifier_group.add_argument('--unroll', metavar='N', default='1',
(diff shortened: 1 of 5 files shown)


delcypher:
@Guoanshisb Note ``--max-depth`` is branch depth and is not a loop bound.

It is also probably not a good idea to rely on the exitCode from symbooglix in the long term as that may change. Symbooglix when it completes emits a YAML file containing counts of the different terminated state types, this is more stable so it would probably be better to rely on that.

share/smack/top.py
```diff
@@ -420,13 +420,21 @@ def verify_bpl(args):
     command += ["/maxStaticLoopBound:%d" % args.loop_limit]
     command += ["/recursionBound:%d" % args.unroll]
 
+  elif args.verifier == 'symbooglix':
+    command = ['symbooglix']
+    command += [args.bpl_file]
+    command += ["--file-logging=0"]
+    command += ["--entry-points=%s" % ",".join(args.entry_points)]
+    command += ["--timeout=%d" % args.time_limit]
+    command += ["--max-depth=%d" % args.unroll]
```
zvonimir: As Dan said, `max-depth` is not the same as loop unroll count. Does symbooglix have something like loop unroll count???

zvonimir:
@delcypher Does symbooglix have something similar to loop unroll bound?

Guoanshisb:
@delcypher  Thanks for your suggestion. I guess that you were referring to `termination_counters.yml`, right? If so, I was wondering if Symbooglix can only generate this file rather than `symbooglix-*` directory.

share/smack/top.py
delcypher: @zvonimir Current there is no loop bound option. One could be implemented but it would require some thought. I'd want a per execution state loop bound and I'd like this to be implemented using support for execution traces which Symbooglix really could do with having.

delcypher:
@Guoanshisb 

> @delcypher Thanks for your suggestion. I guess that you were referring to termination_counters.yml, right? If so, I was wondering if Symbooglix can only generate this file rather than symbooglix-* directory

No files generated by Symbooglix always have to go in a directory because it can generate multiple files that shouldn't clobber any other instance of Symbooglix. You can control the destination directory using ``--output-dir`` option.

You can use

* ``--skip-log-success-states``
* ``--skip-log-unsat-assume-states``
* ``--log-terminated-state-info=0``
* ``--log-non-terminated-state-info=0``
* ``--write-smt2=0``

to reduce the amount of logging files produced.

You should also consider using

* ``--check-entry-requires=0``
* ``--check-entry-axioms=0``
* ``--check-unique-vars=0``

to prevent Symbooglix from trying to check that the initial assumptions are consistent. Checking the assumptions may take some time and in some cases the solver might timeout meaning Symbooglix will make no progress on a boogie program. For my experiments in the Symbooglix ICST2016 paper I checked that the Boogie programs had consistent initial assumptions beforehand and when running Symbooglix I used the above options to disable checking assumptions.

You should also use ``--stop-at-failure=1`` so that Symbooglix stops exploring other paths if it finds a bug. If you've found a bug the Boogie program is incorrect and there isn't a need to check other paths.

share/smack/top.py
zvonimir: Got it. Yeah, most other Boogie checkers (if not all) define loop/recursion bound as the parameter for bounding executions. Ultimately, we would like to run Symbooglix on all SVCOMP benchmarks and do a comparison with e.g. Corral. Currently, it would be hard/unfair to compare them given that we would have to unsure that we are bounding the exploration in the same way (or at least very similar).
Now, once we start running Sumbooglix on SVCOMP, is there a chance you would be interested in implementing support for specifying loop bounds?

delcypher: @zvonimir I would be interested in implementing it. Having thought about it some more there might be away to implement it that doesn't require all the tracing stuff to be implemented. I might give that approach a try as it will be a much faster to implement. Whilst I'm at it, do you need a procedure call bound as well?

zvonimir: Awesome! Well, we would need to be able to specify a bound on the number of loop unrollings and a bound on the number of recursion unrollings. Some tools threat those as one single bound, but it is fine if you would like to threat them as two separate bounds. Thanks!

delcypher: @zvonimir How would want nested loops to be treated? Effectively what I'm thinking of doing is having Symbooglix keep track of what loops it is in whilst executing and incrementing a per path loop head counter every time the loop head basic block is visited and resetting the counter every time that loop is escaped. If the visit count for any basic block for the head of a loop exceeds the bound then the path will be terminated. For nested loops this means the outer and inner loop are treated completely independently. Is this the behaviour you expect?

zvonimir: Yes, this is the behavior I would expect.
Now, I remember that either Boogie or Corral have a different (and less intuitive) behavior when it comes to nested loops. But what you suggested makes sense.

PR #189, opened 2016-11-29
Signed overflow check

Guoanshisb:
Implement a simple signed overflow check pass.


share/smack/top.py
```diff
@@ -141,6 +142,9 @@ def arguments():
   translate_group.add_argument('--memory-safety', action='store_true', default=False,
     help='enable memory safety checks')
 
+  translate_group.add_argument('--signed-integer-overflow', action='store_true', default=False,
+    help='enable signed integer overflow')
```
zvonimir: `enable signed integer overflow checks`

tools/llvm2bpl/llvm2bpl.cpp
```diff
@@ -57,6 +58,10 @@ static llvm::cl::opt<bool>
 MemorySafety("memory-safety", llvm::cl::desc("Enable memory safety checks"),
   llvm::cl::init(false));
 
+static llvm::cl::opt<bool>
+SignedIntegerOverflow("signed-integer-overflow", llvm::cl::desc("Enable signed-integer-overflow checks"),
```
zvonimir: `Enable signed integer overflow checks`

PR #209, opened 2017-03-21
Updating CMakeFiles to compile on Ubuntu 16

dddejan:
Basically, just prefer llvm-config-3.7 to llvm-config (which could be different version).


zvonimir:
Hi Dejan. Thanks for this update!
It would be great if you could create another pull request (with the same commit) onto our `develop` branch (instead of master). That is the branch in which active development happens (see http://nvie.com/posts/a-successful-git-branching-model/), and I don't think I can make that change on my end once a pull request is created.

PR #211, opened 2017-03-22
Updating SMACK to LLVM 3.8

zvonimir:
I updated the source code, build scripts, and documentation to LLVM 3.8.


michael-emmi:
So far the main comment that I have is that the following commits have nothing to do with updating to LLVM 3.8.
- 912bbe2
- 60db558
- 694a631
- 208b79e (except for the one-line change to `CMakeLists.txt`)
If we are being pedantic, then those should just be committed to develop, either before or after this pull request.

I am also not sure whether it is a good idea to change the license on DSAAliasAnalysis. Even if we gut and rename the module to DSAWrapper, legally the original license should probably still hold. (Though I admit that I do not understand the technicalities of software licensing rules.)

zvonimir:
Btw, do you know where the original LLVM's DSAAliasAnalysis source file resides? I can't find it.

michael-emmi:
Not sure how to search for “the original” but the following GitHub search reveals a `DataStructureAA.cpp` file from which our `DSAAliasAnalysis.cpp` was derived at some point.

https://github.com/search?utf8=✓&q=DataStructureAA+extension%3A.cpp&type=Code&ref=advsearch&l=&l=

zvonimir:
Ok, we basically ended up completely rewriting this file, but just in case I did put the original license back on. Doesn't really matter to me either way, and I am not an expert on this either.
I don't know how to separate the commits at this point, and so I will just merge it. In the future, I will be more careful to only work on functionality that belongs to a particular branch.
Thanks for your help!

PR #222, opened 2017-04-18
Symbooglix support

liammachado:
SMACK with Symbooglix as the backend verifier now passes most of the regression tests.


bin/build.sh
```diff
@@ -362,6 +365,24 @@ then
   puts "Built Corral"
 fi
 
+if [ ${BUILD_SYMBOOGLIX} -eq 1 ]
+then
+  puts "Building Symbooglix"
+
+  cd ${ROOT}
+  git clone https://github.com/symbooglix/symbooglix.git ${SYMBOOGLIX_DIR}
```
zvonimir: I think that `git clone --recursive` will clone all the submodules right away, and so you can avoid the two additional git commands below. Please try this out.

bin/build.sh
```diff
@@ -388,25 +409,26 @@ then
...
   puts "Configuring shell environment"
   echo export BOOGIE=\"mono ${BOOGIE_DIR}/Binaries/Boogie.exe\" >> ${SMACKENV}
-  echo export CORRAL=\"mono ${CORRAL_DIR}/bin/Release/corral.exe\" >> ${SMACKENV}
-  echo export LOCKPWN=\"mono ${LOCKPWN_DIR}/Binaries/lockpwn.exe\" >> ${SMACKENV}
-  source ${SMACKENV}
-  puts "The required environment variables have been set in ${SMACKENV}"
-  puts "You should source ${SMACKENV} in your .bashrc"
-
-  puts "Built SMACK"
+	echo export CORRAL=\"mono ${CORRAL_DIR}/bin/Release/corral.exe\" >> ${SMACKENV}
```
zvonimir: It seems that indentation got messed up here. Maybe you are using tabs instead of spaces? Not sure, but please take a look.

bin/build.sh
```diff
@@ -388,25 +409,26 @@ then
...
+	puts "You should source ${SMACKENV} in your .bashrc"
+
+	puts "Built SMACK"
 fi
 
 
 if [ ${TEST_SMACK} -eq 1 ]
 then
-  puts "Running SMACK regression tests"
+	puts "Running SMACK regression tests"
```
zvonimir: Messed up indentation.

bin/build.sh
```diff
@@ -388,25 +409,26 @@ then
...
 
-  cd ${SMACK_DIR}/test
-  ./regtest.py ${TRAVIS_ENV}
-  res=$?
+	cd ${SMACK_DIR}/test
```
zvonimir: Messed up indentation.

bin/build.sh
```diff
@@ -388,25 +409,26 @@ then
...
+	./regtest.py ${TRAVIS_ENV}
+	res=$?
 
-  puts "Regression tests complete"
+	puts "Regression tests complete"
```
zvonimir: Messed up indentation.

test/basic/array2.c
```diff
@@ -3,7 +3,7 @@
 #include "smack.h"
 
 // @flag --loop-limit=11
-// @flag --unroll=11
+// @flag --unroll=100
```
zvonimir: Ah, I see what is going on here. I would prefer a different temporary solution. You should leave these unrolls as they were. Then in the `top.py` file you should add a hacky solution where you multiply the unroll factor by e.g. 10 when being invoked with Symbooglix. I hope this makes sense.

test/config.yml
```diff
@@ -1,6 +1,6 @@
-verifiers: [corral, boogie]
+verifiers: [symbooglix]
```
zvonimir: We should run all 3 of them.

zvonimir:
One more thing - please do not add a period after your commit message title. See https://wiki.openstack.org/wiki/GitCommitMessages

bin/build.sh
liammachado: I tried making this change, but the build failed, printing "Cannot locate a solution file." Something having to do with nuget.exe I think.

liammachado:
I've made all the changes you recommended.

PR #226, opened 2017-05-05
Removing function calls internal to SMACK from DSA

zvonimir:
Removing calls to functions such as `__SMACK_code` and `__SMACK_check_memory_safety` from being analyzed by DSA. These functions are internal to SMACK and are noops as far as DSA is considered. Removing them prevents merging of nodes and makes DSA more precise.

### DIFF
--- a/lib/DSA/StdLibPass.cpp
+++ b/lib/DSA/StdLibPass.cpp
@@ -442,7 +442,7 @@ const struct {
 
 // SMACK functions that are excluded from DSA
 Regex SMACK_PROC_IGNORE("^("
-  "__SMACK_code|__SMACK_mod|__SMACK_decl|__SMACK_top_decl|"
+  "__SMACK_mod|__SMACK_decl|__SMACK_top_decl|"
   "__SMACK_values|__SMACK_check_memory_safety"
 ")$");
 
--- a/lib/DSA/StdLibPass.cpp
+++ b/lib/DSA/StdLibPass.cpp
@@ -24,6 +24,7 @@
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/FormattedStream.h"
+#include "llvm/Support/Regex.h"
 #include "llvm/Support/Timer.h"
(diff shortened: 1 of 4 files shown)


michael-emmi:
Just wondering: if DSA does not analyze those functions, then does that mean that it will not construct DSA nodes for the those functions’ arguments, e.g., when they are not used elsewhere? That could cause problems at some point, e.g., when we try to assign a region to the pointer argument to the memory safety check function.

zvonimir:
I see, so if a pointer passed to one of these functions is not used anywhere else, then a region might not be assigned to it. You might be right. On the other hand, none of our regressions crash because of this change. Having said that, one pthreads benchmarks does return a false bug when `__SMACK_code` is excluded in this way. I did not get to the bottom of this as of yet.
Taking a step back, I am not sure which solution is better - this one or the one you implemented. I have to think more about this.

zvonimir:
Ok, the root cause of the problem I saw was not related to this pull request. The particular regression worked only by chance because of the way that regions got merged. Now I implemented a hacky for this problem and everything should now work.
Coming back to your question/concern, my understanding is that we need regions only when processing load/store instructions. So we don't really ask for a region of a pointer argument to a function. I guess that is why this works in the first place.

michael-emmi:
Not sure whether regions are only needed for load and store instructions. At the very least, I remember them being necessary for various annotations, e.g., in contracts, e.g., in security annotations.

zvonimir:
I see, so I guess we would have to be careful with such functions. One thing I observed while debugging the mentioned pthreads regression is that even `__SMACK_code` causes nodes to be merged when pointers are passed into this function. With this patch we might get more memory splitting in some cases. I will run it on the whole of SVCOMP to see if it has any effect.

zvonimir:
I'll close this pull request. First of all, it hacks into DSA, and I am not totally sure that the changes I made there are fine - since I don't have an excellent understanding of DSA. Second, I think that @michael-emmi at some point improved memory safety checking so that it results in less collapsing of DSA nodes.

PR #227, opened 2017-05-08
Emit timing information into metadata

danielsn:
This is a new pass which records data useful for future passes that can use timing analysis data (e.g. ctverif).

1. Annotate the metadata of each LLVM instruction with the timing cost in the LLVM architectural model
2. Annotate the metadata of each LLVM instruction with information about the name and type of the instruction.  This is useful for future boogie passes that might want to branch on that information, which will have otherwise been lost in translation.
3. Add a generic mechanism to emit any metadata that starts with ```smack.``` to boogie annotations, making it easy to use in later passes such as boogieman.


include/smack/AddTiming.h
```diff
@@ -0,0 +1,49 @@
...
+    unsigned getInstructionCost(const Instruction *I) const;
+
+  private:
+    void addTimingMetadata(Instruction *Inst, const std::string &name, const std::string& value) const;
+    void addTimingMetadata(Instruction *Inst, const std::string &name, unsigned cost) const;
+    void addTimingMetadata(Instruction *Inst, unsigned cost) const;
+    void getAnalysisUsage(AnalysisUsage &AU) const override;
+    bool runOnFunction(Function &F) override;
+    void print(raw_ostream &OS, const Module*) const override;
+    std::string nameInstruction(Instruction *Inst) const;
```
michael-emmi: This function is not defined nor used.

michael-emmi:
Hey @danielsn, any rebuttal to the review? are the suggested changes acceptable? we can move fast on merging once all issues are addressed.

danielsn:
Hi.  I'm doing the changes now - just need to get them approved by legal,
and then I'll ship you the new PR.

DSN

michael-emmi:
Hey @danielsn, thanks for the updates. Three basic issues remain:

1. This currently breaks on load instructions which return struct-type values. The breakage occurs [here](http://llvm.org/docs/doxygen/html/ValueTypes_8cpp_source.html#l00276) since struct is not a type this code is willing to handle. We need to dig deeper to figure out how to best deal with this.

2. Is it possible to move all/most of the additions to `llvm2bpl.cpp` into the AddTiming pass itself? The AddTiming pass should be given the relevant module anyways, so the target information should be accessible there. And normally you would include required passes as dependencies, rather than adding them to the PassManager explicitly. We should strive for containment of as much of this code as possible into the AddTiming module.

3. Lastly, we will want to enable the AddTiming pass and related annotations conditionally, e.g., with a flag, possibly named `--timing-annotations`. You can follow the examples from the SmackOptions C++ module, along with the `top.py` Python module (to pass flags from the `smack` top-level script to the `llvm2bpl` executable).

danielsn:
HI @michael-emmi 

1.  I wonder if this is because the llvm code is not expecting to see structs at this point - perhaps they should have been lowered? 

2. I will move as much of the code as I can.  Much of the code that I added to llvm2bpl was based on code in the llvm opt package, so they considered it generic.

3.  Willdo.

michael-emmi:
Ok I think I understand what is happening.

We are using a pass (StructRet) which removes dependence on struct-value returns where possible, by adding a pointer argument to which the struct-value is mem-copied before the function returns. While this pass can make the return value irrelevant, it does not actually delete the return instruction, nor the corresponding struct-value load instruction that takes place just before. In principle, both instructions can probably be removed without harm by this pass.

Currently, Smack incorrectly translates struct-value load instructions as if they were pointer-value load instructions. So far this has done no harm, apparently because the only struct-value load instructions we have come across were those preceding those struct-value return instructions which are made irrelevant by the StructRet pass. So even though the translation of struct-value load instructions is wrong, it does not harm the overall translation.

Going back to your issue, I guess there are two possible approaches. First, we could have StructRet “finish the job“ and just remove those irrelevant struct-value load and return instructions (and change the function signature to void, long with the corresponding callers). However, there could be some reason why those functionally-irrelevant instructions are not removed, and furthermore, there are cases where the StructRet transformation is not applied, e.g., if the function in question has its address taken at some point.

The other thing we might try is attempting to compute your own cost, or just ignore the cost, for struct-value load instructions in `AddTiming::getInstructionCost`. I guess the right thing here depends on what code would be generated for these irrelevant struct-value load and returns; perhaps they will somehow be pruned before machine code generation?

michael-emmi:
@zvonimir btw do you have any idea why the StructRet pass might be leaving in the struct-value load and return instructions that its transformation makes irrelevant?

zvonimir:
As usual, my only wild guess would be that it has something to do with DSA :). Otherwise, I cannot come up with any good reasons for leaving these instructions.
Btw, I think that @Guoanshisb has improved this pass in the past month, but that still did not get merged. @Guoanshisb can you comment on this (and on your improvements to `StructRet` pass if there are any)?

Guoanshisb:
@zvonimir @michael-emmi I wrote a new pass that split load/store to structs into those to their scalar fields rather than improved that pass. I guess that the reason load/return instructions are not removed is that it is not necessary.

zvonimir:
So your new pass does not subsume the `StructRet` pass? Meaning, we need both of them?

Guoanshisb:
Yes.

include/smack/AddTiming.h
```diff
@@ -0,0 +1,54 @@
...
+namespace smack {
+  using namespace llvm;
+  
+  class AddTiming : public FunctionPass {
+
+    enum Flags { NO_TIMING_INFO = -1};
+    static const std::string INT_TIMING_COST_METADATA;
+    static const std::string INSTRUCTION_NAME_METADATA;
+    
+    Naming naming;
```
michael-emmi: This field is not used.

include/smack/AddTiming.h
```diff
@@ -0,0 +1,54 @@
...
+// This file is distributed under the MIT License. See LICENSE for details.
+//
+#ifndef ADDTIMING_H
+#define ADDTIMING_H
+
+#include "llvm/Pass.h"
+#include "llvm/Analysis/Passes.h"
+#include "llvm/Analysis/TargetTransformInfo.h"
+#include <string>
+#include "smack/Naming.h"
```
michael-emmi: This include is unnecessary since `Naming` is not used here.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -73,15 +73,42 @@ void SmackInstGenerator::annotate(llvm::Instruction& I, Block* B) {
       return;
     }
   }
-
+  
   if (SmackOptions::SourceLocSymbols && I.getMetadata("dbg")) {
     const DebugLoc DL = I.getDebugLoc();
     auto *scope = cast<DIScope>(DL.getScope());
     B->addStmt(Stmt::annot(Attr::attr("sourceloc", scope->getFilename().str(),
-      DL.getLine(), DL.getCol())));
```
michael-emmi: Whitespace should not be changed, especially against formatting conventions.

tools/llvm2bpl/llvm2bpl.cpp
```diff
@@ -80,17 +86,50 @@ namespace {
...
+			  Options,
+			  Reloc::Static, /* was getRelocModel(),*/
+			  CodeModel::Default, /* was CMModel,*/
+			  CodeGenOpt::None /*GetCodeGenOptLevel())*/
+			  );
+
+  }
 }
 
+
```
michael-emmi: Please remove whitespace-only changes.

tools/llvm2bpl/llvm2bpl.cpp
```diff
@@ -80,17 +86,50 @@ namespace {
...
+
 int main(int argc, char **argv) {
   llvm::llvm_shutdown_obj shutdown;  // calls llvm_shutdown() on exit
   llvm::cl::ParseCommandLineOptions(argc, argv, "llvm2bpl - LLVM bitcode to Boogie transformation\n");
 
-  llvm::sys::PrintStackTraceOnErrorSignal();
+    llvm::sys::PrintStackTraceOnErrorSignal();
```
michael-emmi: Please remove whitespace-only changes.

tools/llvm2bpl/llvm2bpl.cpp
```diff
@@ -99,6 +138,7 @@ int main(int argc, char **argv) {
   if (L.empty())
     module.get()->setDataLayout(DefaultDataLayout);
 
+
```
michael-emmi: Please remove whitespace-only changes.

michael-emmi:
In principle this can be merged now. I think the only remaining concern is how to protect this feature from breakage due to future updates. Ideally we could include a small number of regression tests to check that timing annotations are present and not obviously wrong. It is not clear that such regression tests fit into our existing regression test framework in `test/regtest.py`, which only checks verifier results. Can we see any simple way of including a small number of timing-annotation regression tests?

Issue #219, opened 2017-04-06
Using SMACK for Rust code


cmr:
Sort of a meta-issue I guess.

Depends on #165, probably #54.

I'm experimenting with this now.

zvonimir:
We have an ongoing preliminary effort that is extending SMACK to be able to use it on Rust. If you are interested in this direction and would like to contribute, please let us know. There is certainly a lot of work still to be done :).

ZackPierce:
I'm curious about the current state of the preliminary efforts to support Rust.  Has anything been done beyond the https://github.com/smackers/smack/compare/rust-support prototype branch from a year ago?

I've tossed up a super minimal rust project within the [examples directory on a branch](https://github.com/ZackPierce/smack/tree/rusty), accompanied by a helper script that mimics some core bits of the SMACK toolchain. Currently the process dies in the middle of llvm2bpl.

HiDefender:
I'm interested in making this happen and have the time to devote to it. Where is the best place to start?

zvonimir:
@HiDefender, @ZackPierce, @cmr : FYI, we finally have a preliminary working (kind of) support for Rust in the rust-prelims branch. Feel free to try it out. I am sure that we'll identify a lot of outstanding issues and missing features in the next several weeks, and help with these would be greatly appreciated.

HiDefender:
@zvonimir Research has gone a different direction. Pretty sure I'll still need something like SMACK for Rust in the future, but for the time being I need to get my preliminaries working.

ruffsl:
Hello @zvonimir , I'd like jump in and try using smack with rust for information flow control. I see there are some ifc examples in the `rust-prelims` branch, with the latest commit aa871e5 on Jul 19, 2018.

I also see there is the branches:
* `modular-rust` with the latest commit f1630a0 on Jan 24, 2018
* `rust-llvm-7` with the latest commit ba405c6 on Jul 31, 2019.

For those just getting to know smack, rust, an IFC, where would be a good point to start, and what is the state/progress/purpose of each rust related branch thus far?

zvonimir:
Hey @ruffsl and thanks for your interest in SMACK. So currently the `rust-llvm-7` is the most up to date branch with support for Rust. Note that it actually supports LLVM 8 :).

We are actively working on Rust support, and I expect for this branch to be merged into develop soon. But that branch would be a good place to start. We should also create a simple tutorial for how to run SMACK on Rust, so keep your eyes open for that as well.

[...]

PR #231, opened 2017-06-05
Upgrade LLVM to 3.9

ZackPierce:
Upgrade LLVM dependency to 3.9, as part of an effort to tackle some related aspects of https://github.com/smackers/smack/issues/219 .

Currently split into multiple commits to make review easier, though I can collapse per the contribution guidelines after review.

The build changes involve switching to sourcing the apt.llvm.org repo rather than llvm-apt.ecranbleu.org repo.  I had to manually remove references to llvm-apt.ecranblue.org in `/etc/apt/sources.list`. Users starting from scratch (or using vagrant) should not have to worry about that step.

The C-land changes are largely renames and function signature changes, or the consequences thereof.

The only truly incidental alterations are build-script additions to support linux-neon-16 in order to make my particular dev cycle turnaround faster. I can remove that bit if desired.


ZackPierce:
Hmm, it looks like the Travis build [timed out on an unrelated request](https://travis-ci.org/smackers/smack/jobs/239608048) to Nuget. I'm uncertain what the policy on retries is. Perhaps or @zvonimir @michael-emmi know?

michael-emmi:
I’ve just triggered a restart on the failing build; hopefully it will go through this time.

bin/build.sh
```diff
@@ -221,21 +226,38 @@ then
     sudo zypper --non-interactive install ${DEPENDENCIES}
     ;;
 
+  linux-neon-1[6]*)
```
michael-emmi: Could this case possibly be merged with the following `linux-ubuntu` case? The only differences seem to be that `xential` replaces `trusty` in the LLVM apt repo URL. Merging might be achieved by adding a release-name table mapping `neon` to `xenial` and `ubuntu` to `trusty`.

bin/build.sh
```diff
@@ -162,7 +162,12 @@ linux-opensuse*)
 
 linux-ubuntu-1[46]*)
   Z3_DOWNLOAD_LINK="https://github.com/Z3Prover/z3/releases/download/z3-4.4.1/z3-4.4.1-x64-ubuntu-14.04.zip"
-  DEPENDENCIES+=" clang-3.8 llvm-3.8 mono-complete libz-dev libedit-dev"
+  DEPENDENCIES+=" clang-3.9 llvm-3.9 mono-complete libz-dev libedit-dev"
+  ;;
+
+linux-neon-1[6]*)
```
michael-emmi: Could this case possibly be merged with the previous `linux-ubuntu` case?

michael-emmi:
Thanks for the contribution @ZackPierce !

We will have to discuss when to merge this one with @zvonimir since previously we had been upgrading our LLVM compatibility according to some schedule which escapes me at the moment. In the worst case, we may have to maintain two separate branches for each LLVM version we want to support at any given time, e.g., 3.8 and 3.9; not sure if there’s any good reason to keep 3.8 around.

Your thoughts @zvonimir ?

One further technical comment/question: couldn’t we make the version-number symbolic in the `bin/build.sh` script? E.g., instead of changing every single instance of `xxxx-3.8` to `xxxx-3.9` we might have just written `xxxx-${LLVM_VERSION}` and updated the single `LLVM_VERSION` variable from `3.8` to `3.9`. Wouldn’t it be more reasonable to do this going forward?

zvonimir:
Indeed, thanks for the contribution @ZackPierce, especially since I have been in charge of updating LLVM versions in the past and it is not the most exciting task :).
I am traveling now and I won't have time to look at this in more detail until this coming Friday or so.
I think there is no need to keep LLVM 3.8 version around. We decided at some point not to provide strong backwards compatibility when it comes to LLVM. So our strategy is to stay around a year or so behind LLVM so that we also get their bug-patch release as well before we bump up our LLVM version.
I like @michael-emmi's suggestion to make the LLVM version in our build script symbolic. So maybe you could try to do that update?

ZackPierce:
Sure, I'll give those modifications a shot. Thanks for the speedy feedback.

ZackPierce:
Alright, now the latest commit collapses the the ubuntu/neon branches, detects ubuntu codenames (`trusty` vs `xenial`) and adds LLVM version parameterization.

Jinxxed:
Hi,

exactly 1 day before you submitted this pull request, I tried to make SMACK compile with LLVM 4.0 myself.
This seems to work pretty well, but as soon as I have a printf in the code somewhere, I get a Segmentation Fault, since IntType in LLVM is not a DerivedType anymore. I just checked and realized this is also the case for LLVM 3.9.
May I ask, how did you solve this issue?

Thank you!

Jinxxed:
To be more precise, I the Segfault happens in lib/smack/SmackRep.cpp:687, with a line like
```c
printf("%d %d", a, b);
```
in the source code

michael-emmi:
Shall we merge this one?

zvonimir:
Before merging such potentially significant updates, I like to run the whole SV-COMP set so that we can catch unforeseen problems that our regressions miss, including performance issues. We still did not get a chance to do this since we've been working on testing other updates, and hence we do not have a stable baseline for comparison.

zvonimir:
@ZackPierce : Sorry about the delay. We are now ready to thoroughly test this on the whole of SV-COMP benchmark set. Could you please rebase this to the latest develop and also fix these conflicts? Thanks!

ZackPierce:
Roger that, will give it a go.

ZackPierce:
Rebased, resolved conflicts, and added in an additional version-mismatch fix that was required due to the upstream work on develop (re: `getIntrinsicInstrCost` API changes).

zvonimir:
Thanks @ZackPierce!
@shaobo-he : Could you please run this on the whole of SV-COMP? Thx!

shaobo-he:
@zvonimir Yes, I am working on it.

PR #233, opened 2017-06-14
Added assumption to alloc that size is not negative

zvonimir:
For some reason when size was negative we used to allocate an
object of size 1. This seems like a hack to me and I am not sure
why it was introduced in the first place. I cleaned this up now
by requiring that size is not negative, and blocking executions
with negative size.

### DIFF
--- a/share/smack/lib/smack.c
+++ b/share/smack/lib/smack.c
@@ -2034,7 +2034,9 @@ void __SMACK_decls() {
   D("procedure $malloc(n: ref) returns (p: ref)\n"
     "modifies $allocatedCounter;\n"
     "{\n"
-    "  $allocatedCounter := $allocatedCounter + 1;\n"
+    "  if ($ne.ref.bool(n, $0.ref)) {\n"
+    "    $allocatedCounter := $allocatedCounter + 1;\n"
+    "  }\n"
     "  call p := $$alloc(n);\n"
     "}\n");
 
--- a/share/smack/lib/smack.c
+++ b/share/smack/lib/smack.c
@@ -2055,8 +2055,8 @@ void __SMACK_decls() {
   D("procedure {:inline 1} $$alloc(n: ref) returns (p: ref)\n"
     "modifies $Alloc, $CurrAddr;\n"
     "{\n"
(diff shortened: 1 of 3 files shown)


zvonimir:
I will also add that this update does not significantly alter our SV-COMP results.

share/smack/lib/smack.c
```diff
@@ -2053,18 +2053,17 @@ void __SMACK_decls() {
   D("procedure {:inline 1} $$alloc(n: ref) returns (p: ref)\n"
     "modifies $Alloc, $CurrAddr;\n"
     "{\n"
-    "  p := $CurrAddr;\n"
-    "  havoc $CurrAddr;\n"
+    "  assume $sge.ref.bool(n, $0.ref);\n"
```
michael-emmi: Won’t the assumption of `n >= 0` cause programs that do `malloc(-1)` to verify trivially? Is this situation somehow averted by type conversions?

michael-emmi: Doesn’t this request address some recently-opened issue by @Guoanshisb ? If so the commit should probably reference the issue.

zvonimir: Not really. This came up while we were discussing one of the issues he reported, but it is not directly related. I just wanted to address this one first.

zvonimir: In some situations when unbounded integers are used, yes, I believe you are right. But that is a more general issue related to our inability to correctly model unsigned integers (and casts between unsigned and signed) when unbounded integers are used. So I can create an example even without malloc where a bug would be missed due to this deficiency - if I remember a common patters we do not model correctly is casting a negative integer into an unsigned integer.

michael-emmi:
Could we add a regression here to ensure that the beneficial parts of this PR persist?

zvonimir:
I just added such regression.

zvonimir:
@michael-emmi : Are we ok to merge now? Thx!

michael-emmi:
I suppose so; I do not have a good feeling for the ramifications of these changes, so as long as regressions pass, and performance stays fine, then I am happy.

PR #234, opened 2017-06-15
Generalize testing

michael-emmi:
Adding @checkbpl and @checkout fields to regression tests to check the generated Boogie file and verifier output against arbitrary shell-command predicates.

### DIFF
--- a/test/basic/checking-fail-bpl.c
+++ b/test/basic/checking-fail-bpl.c
@@ -4,7 +4,7 @@
 
 // @expect error
 // @checkbpl grep -v "call bar"
-// @checkout grep "CALL foo"
+// @checkout grep "checking-fail-bpl.c(11,1)"
 
 void foo(void) {
 
--- a/test/basic/checking-fail-bpl.c
+++ b/test/basic/checking-fail-bpl.c
@@ -1,0 +1,17 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include "smack.h"
+
+// @expect error
+// @checkbpl grep -v "call bar"
+// @checkout grep "CALL foo"
+
+void foo(void) {
+
+}
+
+int main(void) {
+  foo();
+  assert(0);
+  return 0;
+}
--- a/test/config.yml
+++ b/test/config.yml
@@ -3,4 +3,6 @@ memory: [no-reuse-impls, no-reuse, reuse]
 time-limit: 120
 memory-limit: 450
 flags: [--unroll=2]
+checkbpl: []
+checkout: []
 skip: false
--- a/test/regtest.py
+++ b/test/regtest.py
@@ -15,9 +15,10 @@ import re
 import glob
 import time
 import sys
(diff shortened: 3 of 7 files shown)


test/basic/checking-fail-bpl.c
```diff
@@ -0,0 +1,17 @@
+#include <stdio.h>
```
zvonimir: The convention I have been using is that failing regressions should have file names ending in `_fail`. It would be good to stick with that.

test/config.yml
```diff
@@ -3,4 +3,6 @@ memory: [no-reuse-impls, no-reuse, reuse]
 time-limit: 120
 memory-limit: 450
 flags: [--unroll=2]
+checkbpl: []
```
zvonimir: Why do you have to set them to empty lists here?

test/regtest.py
```diff
@@ -105,10 +114,24 @@ def process_test(cmd, test, memory, verifier, expect, log_file):
   p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
   out, err  = p.communicate()
   elapsed = time.time() - t0
+  status = 0
+
+  bplfile = cmd[cmd.index('-bpl')+1]
+  with open(os.devnull, 'w') as devnull:
+    for f in checkbpl:
+      with open(bplfile) as bpl:
+        checker = subprocess.Popen(shlex.split(f), stdin=bpl, stdout=devnull, stderr=devnull)
```
zvonimir: It took me some time to understand this :). Why are you invoking grep Linux command as a separate process instead of doing a string search directly from Python?

michael-emmi: Because `grep` is just kind of a degenerate case of checking the generated Boogie file. I think we want to allow for more flexibility, hence the checking predicate can be an arbitrary command. E.g., you may want to type-check the generated Boogie code, which can be done by wrapping up `boogie /noVerify /doModSetAnalysis`. E.g., you may want to run some simple analysis on the generated Boogie code, which might be done by `bam --my-analysis`. This implementation is the simplest way of being relatively future-proof.

test/config.yml
michael-emmi: The `regtest.py` script requires that fields like `flags` and `checkbpl` that can be appended to by nested `config.yml` files be initialized in the root `test/config.yml`. Otherwise, e.g., this line would fail:

https://github.com/smackers/smack/blob/master/test/regtest.py#L80

test/basic/checking-fail-bpl.c
michael-emmi: True; perhaps `fail` was the wrong word to use, because these regressions do not “fail” in the sense that the others labeled by `_fail` do — i.e., by failing assertions.

I have changed the names of these regressions to avoid confusion.

zvonimir:
Looks good to me. Feel free to merge.

PR #236, opened 2017-06-18
Adding verifier-code metadata.

michael-emmi:
Adds metadata annotations in LLVM bitcode to distinguish between “erasable” instructions that pertain only to verification, e.g., values feeding into `assume` and `assert` statements, and regular instructions that will actually be executed in the generated code.

### DIFF
--- a/lib/smack/VerifierCodeMetadata.cpp
+++ b/lib/smack/VerifierCodeMetadata.cpp
@@ -51,10 +51,10 @@ namespace {
       if (N.find("__VERIFIER_") == 0)
         return true;
 
-      if (N.find("__SMACK") == 0)
+      if (N.find("__SMACK_") == 0)
         return true;
 
-      if (N.find("__CONTRACT") == 0)
+      if (N.find("__CONTRACT_") == 0)
         return true;
     }
     return false;
--- a/lib/smack/VerifierCodeMetadata.cpp
+++ b/lib/smack/VerifierCodeMetadata.cpp
@@ -64,6 +64,33 @@ namespace {
     }
     return false;
   }
+
+  bool onlyVerifierUsers(Instruction &I) {
+    std::queue<User*> users;
+    std::set<User*> known;
+
+    for (auto U : I.users()) {
+      users.push(U);
+      known.insert(U);
+    }
+
+    while (!users.empty()) {
+      if (auto K = dyn_cast<Instruction>(users.front())) {
+        if (!isMarked(*K))
+          return false;
+
+      } else {
+        for (auto UU : users.front()->users()) {
+          if (known.count(UU) == 0) {
+            users.push(UU);
+            known.insert(UU);
+          }
+        }
+      }
+      users.pop();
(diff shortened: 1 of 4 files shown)


lib/smack/VerifierCodeMetadata.cpp
```diff
@@ -0,0 +1,124 @@
+//
+// This file is distributed under the MIT License. See LICENSE for details.
+//
+
+#define DEBUG_TYPE "contracts"
```
zvonimir: I guess that the debug type should probably not be `contracts`.

lib/smack/VerifierCodeMetadata.cpp
```diff
@@ -0,0 +1,124 @@
...
+    workList.pop();
+  }
+  return true;
+}
+
+void VerifierCodeMetadata::visitCallInst(CallInst &I) {
+  auto marked = false;
+
+  if (auto F = I.getCalledFunction()) {
+    if (auto V = F->getName().find("__VERIFIER_") == 0) {
```
zvonimir: What about functions such as `__SMACK_code`?

lib/smack/VerifierCodeMetadata.cpp
```diff
@@ -0,0 +1,124 @@
...
+  }
+}
+
+void VerifierCodeMetadata::getAnalysisUsage(AnalysisUsage &AU) const {
+  AU.addRequired<DominatorTreeWrapperPass>();
+}
+
+bool VerifierCodeMetadata::runOnModule(Module &M) {
+  visit(M);
+  while (!workList.empty()) {
```
zvonimir: It would take me some time to check this logic, but I trust  that you did it correctly. I would guess that you are being conservative if you encounter something like a store instruction - since you are not tracking uses through memory.
Also, and this occurred to me while back, it might be useful to actually add comments here and there in the code we write :).

michael-emmi:
Hey @zvonimir thanks for your comments! The logic should be clearer now.

lib/smack/VerifierCodeMetadata.cpp
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  bool isVerifierFunctionCall(CallInst &I) {
+    if (auto F = I.getCalledFunction()) {
+      auto N = F->getName();
+
+      if (N.find("__VERIFIER_") == 0)
+        return true;
+
+      if (N.find("__SMACK") == 0)
```
zvonimir: To make it consistent with `VERIFIER`, this should probably be `__SMACK_`.

lib/smack/VerifierCodeMetadata.cpp
```diff
@@ -0,0 +1,140 @@
...
+        return true;
+
+      if (N.find("__CONTRACT") == 0)
```
zvonimir: To make it consistent with `VERIFIER`, this should probably be `__CONTRACT_`.

PR #241, opened 2017-06-25
Update versions of Corral, Boogie, and Z3

zvonimir:
Updated all these tools to their latest versions.


michael-emmi:
@zvonimir does the increase in regression timeout parameters mean that upgrading to newer versions of Boogie, Corral, and Z3 incurred a performance hit?

zvonimir:
Yes, for some benchmarks. We also performed a large scale comparison on the whole SV-COMP set of benchmarks. We observe slightly more time outs with the updated versions, but no alarming differences. This could be due to slightly different heuristics in these backend tools.

PR #243, opened 2017-06-26
Fixed an issue with the square root function

Checkmate50:


test/float/squareroot.c
```diff
@@ -0,0 +1,15 @@
+#include <stdio.h>
```
zvonimir: Create a failing version of this benchmark. In general, we always aim to create both a correct and a failing version.

test/float/squareroot.c
```diff
@@ -0,0 +1,15 @@
...
+#include <stdlib.h>
+#include "smack.h"
+
+// @expect verified
+
+int main() {
+    double a;
```
zvonimir: Wrong indentation. We always use 2 spaces.

Checkmate50:
There are some issues with what was included with this pull request

PR #251, opened 2017-07-21
Split load/store instructions of LLVM struct types

shaobo-he:
This pull request intends to model load/store instructions of LLVM struct types. Previously, SMACK translates a load instruction to a variable with an aggregate type into a load function returning `ref` type and  does not allow store instructions with aggregate store values. With this pull request, a load instruction of a struct type is translated into a series of l`oad` instructions and `insertvalue` instructions while a store instruction is translated into a series of `getelementptr` instructions (optional when the value operand is a constant) and `store` instructions. For example,

```llvm
 store { i64, i64 } { i64 ptrtoint (i8* (%"class.Memory::WriteBarrierPtr.8"*, i64)* @_ZN6Memory13HeapAllocator5AllocEm to i64), i64 0 }, { i64, i64 }* %10, align 8, !dbg !49654
```
==>
```llvm
  %20 = getelementptr { i64, i64 }, { i64, i64 }* %5, i32 0, i32 1, !dbg !42958
  store i64 0, i64* %20, !dbg !42958                                                                  
  %21 = getelementptr { i64, i64 }, { i64, i64 }* %5, i32 0, i32 0, !dbg !42958
  store i64 ptrtoint (i8* (%"class.Memory::WriteBarrierPtr.8"*, i64)* @_ZN6Memory13HeapAllocator5AllocEm to i64), i64* %21, !dbg !42958 
```
I added an option `--split-structs` such that the pass can be invoked on demand.


michael-emmi:
This PR needs a simple test which only succeeds with the `--split-structs` flag enabled. I came up with the following test which I thought would do the trick, but it doesn't:
````
#include <smack.h>

typedef struct {
  int x;
  int y;
  int z;
} S;

S foo() {
  S s;
  s.x = __VERIFIER_nondet_int();
  s.y = 2;
  s.z = __VERIFIER_nondet_int();
  return s;
}

int main() {
  S s = foo();
  assert(s.y == 2);
  return 0;
}
````
It seems like structs are indeed split in the translation with `--split-structs`, however Smack still reports a failed assertion.

**Comment by shaobo-he** [comment]
Hi @michael-emmi,

Thanks for taking a look at this pull request. Yes, I agree that we need a regression test. However, it seems to me that it's much easier to come up with a program in bitcode that exercises this PR. The example in your comment seems to require bit-precision since the first two fileds, `x,y` in `S` are coerced into an `i64`. How about using the following slightly modified version of the example?
```C
#include <smack.h>

typedef struct {
  int x;
  int y;
  int z;
} S;

S foo() {
  S s;
  s.x = 1;
  s.y = 2;
  s.z = 3;
  return s;
}

int main() {
  S s = foo();
  assert(s.z == 3);
  return 0;
}
```

On my machine, the assertions fails w.o. `--split-structs` while does not otherwise.

michael-emmi:
Sure, add that example then. Remember to add the line
````
// @flag --split-structs
````
to the regression.

File: include/smack/SplitStructLoadStore.h [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+#include "llvm/IR/Module.h"
+#include "llvm/Pass.h"
+#include "llvm/IR/IRBuilder.h"
+
+#include <vector>
+#include <utility>
+
+namespace smack {
+
+class SplitStructLoadStore : public llvm::ModulePass {
```
michael-emmi: According to the [LLVM docs](http://llvm.org/docs/WritingAnLLVMPass.html#pass-classes-and-requirements) it seems like this should be a `BasicBlockPass` and not a `ModulePass`:

> When choosing a superclass for your Pass, you should choose the most specific class possible, while still being able to meet the requirements listed. This gives the LLVM Pass Infrastructure information necessary to optimize how passes are run, so that the resultant compiler isn’t unnecessarily slow.

include/smack/SplitStructLoadStore.h
```diff
@@ -0,0 +1,31 @@
...
+public:
+  static char ID;
+
+  SplitStructLoadStore() : llvm::ModulePass(ID) {}
+  virtual bool runOnModule(llvm::Module& M);
+private:
+  const llvm::DataLayout* DL;
```
michael-emmi: This variable is initialized and never used.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+    std::vector<unsigned> ret;
+    for (auto p = lst.begin()+1; p != lst.end(); ++p)
+      ret.push_back(std::get<1>(*p));
+    return ret;
+}
+
+bool SplitStructLoadStore::runOnModule(Module& M)
+{
+  DL = &M.getDataLayout();
```
michael-emmi: This variable (`DL`) is initialized and never used.

include/smack/SplitStructLoadStore.h
```diff
@@ -0,0 +1,31 @@
...
+  std::vector<llvm::Instruction*> toRemove;
```
michael-emmi: Couldn't this be a local variable — e.g., in what is currently the `runInModule` function — rather than an instance variable?

share/smack/top.py
```diff
@@ -175,6 +175,9 @@ def arguments():
   translate_group.add_argument('--float', action="store_true", default=False,
     help='enable bit-precise floating-point functions')
 
+  translate_group.add_argument('--split-structs', action='store_true', default=False,
```
michael-emmi: Do we really want this to be optional? Are there any bad things that can happen by splitting structs into individual accesses? Clearly not doing it is less precise.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,151 @@
...
+  return true;
+}
+
+void SplitStructLoadStore::splitStructLoad(LoadInst* li)
+{
+  if (StructType* st = dyn_cast<StructType>(li->getType())) {
+    std::vector<std::pair<Value*, unsigned> > idx;
+    IRBuilder<> irb(li);
+    li->replaceAllUsesWith(buildStructs(&irb, li->getPointerOperand(), st, nullptr, idx));
+    toRemove.push_back(li);
```
michael-emmi: This could just as easily be done at the call site of `splitStructLoad`, making `toRemove` a local variable.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,151 @@
...
+  return cv;
+}
+
+void SplitStructLoadStore::splitStructStore(StoreInst* si, Value* ptr, Value* val)
+{
+  if (StructType* st = dyn_cast<StructType>(val->getType())) {
+    std::vector<std::pair<Value*, unsigned> > idx;
+    IRBuilder<> irb(si);
+    copyStructs(&irb, ptr, st, val, idx);
+    toRemove.push_back(si);
```
michael-emmi: This could just as easily be done at the call site of `splitStructStore`, making `toRemove` a local variable.

share/smack/top.py
shaobo-he: Let's make it optional for now since we haven't tested this flag extensively on SVCOMP benchmarks. If the results look fine, we can remove this flag.

shaobo-he:
@michael-emmi Thanks for your suggestions. I've refactored the code according to your comments and am waiting for the result of Travis CI.

zvonimir:
Hi Shaobo. Could run please run our usual SVCOMP extensive benchmarking steps to make sure this does not break anything? Please report back once that is done. Thanks!

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
+#include "smack/SplitStructLoadStore.h"
+#include "llvm/IR/InstIterator.h"
+
+namespace smack {
+
+using namespace llvm;
+
+std::vector<Value*> getFirsts(std::vector<std::pair<Value*, unsigned> > lst)
```
zvonimir: You have a coding style mistake here - we put open curly braces after function definition, meaning after `)`. Take a look at other files for examples and then please fix this.

include/smack/SplitStructLoadStore.h
```diff
@@ -0,0 +1,29 @@
...
+class SplitStructLoadStore : public llvm::BasicBlockPass {
+public:
+  static char ID;
+
+  SplitStructLoadStore() : llvm::BasicBlockPass(ID) {}
+  virtual bool runOnBasicBlock(llvm::BasicBlock& BB);
+
+private:
+  void splitStructLoad(llvm::LoadInst* li);
+  llvm::Value* buildStructs(llvm::IRBuilder<> *irb, llvm::Value* ptr, llvm::Type* ct, llvm::Value* val, std::vector<std::pair<llvm::Value*, unsigned> > idxs);
```
zvonimir: Please break these long lines somewhere around 80-100 characters.

lib/smack/SplitStructLoadStore.cpp
zvonimir: You probably don't need this space after `unsigned>`.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
...
+    std::vector<Value*> ret;
+    for (auto p = lst.begin(); p != lst.end(); ++p)
+      ret.push_back(std::get<0>(*p));
+    return ret;
+}
+
+std::vector<unsigned> getSeconds(std::vector<std::pair<Value*, unsigned> > lst)
+{
+    std::vector<unsigned> ret;
+    for (auto p = lst.begin()+1; p != lst.end(); ++p)
```
zvonimir: Wait, what happens here if `lst.begin == lst.end`?? Meaning, when you have only one element in this list. Is that situation possible?
If not, then it would be good to add an assertion here to make sure.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
...
+}
+
+bool SplitStructLoadStore::runOnBasicBlock(BasicBlock& BB)
+{
+  std::vector<Instruction*> toRemove;
+  for(Instruction& I : BB) {
+    if (LoadInst* li = dyn_cast<LoadInst>(&I)) {
+      bool isReturnValue = false;
+      for (auto u = li->user_begin(); u != li->user_end(); ++u)
+      {
```
zvonimir: Again, fix where you open curly braces.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  for (auto& i : toRemove)
+    i->eraseFromParent();
+  return true;
+}
+
+void SplitStructLoadStore::splitStructLoad(LoadInst* li)
+{
+  if (StructType* st = dyn_cast<StructType>(li->getType())) {
+    std::vector<std::pair<Value*, unsigned> > idx;
```
zvonimir: No space needed after `unsigned>`.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
...
+    IRBuilder<> irb(li);
+    li->replaceAllUsesWith(buildStructs(&irb, li->getPointerOperand(), st, nullptr, idx));
+  }
+}
+
+Value* SplitStructLoadStore::buildStructs(IRBuilder<> *irb, Value* ptr, Type* ct, Value* val, std::vector<std::pair<Value*, unsigned> > idxs)
```
zvonimir: Split this long line.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
...
+      bool isReturnValue = false;
+      for (auto u = li->user_begin(); u != li->user_end(); ++u)
+      {
+        if (auto ri = dyn_cast<ReturnInst>(*u)) {
+          isReturnValue = true;
+          break;
+        }
+      }
+      if (!isReturnValue && li->getType()->isAggregateType()) {
+        splitStructLoad(li);
```
zvonimir: Hmmm, in `splitStructLoad` you do not always perform this splitting since this function start with an if condition. And yet here you always add this instruction into `toRemove`. Is this the right thing to do even when line 58 evaluates to false?

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
...
+}
+
+Value* SplitStructLoadStore::buildStructs(IRBuilder<> *irb, Value* ptr, Type* ct, Value* val, std::vector<std::pair<Value*, unsigned> > idxs)
+{
+  LLVMContext& C = ptr->getContext();
+  Value* cv = val? val : UndefValue::get(ct);
+
+  if (ct->isIntegerTy() || ct->isPointerTy() || ct->isFloatingPointTy())
+    return irb->CreateInsertValue(val,
+             irb->CreateLoad(irb->CreateGEP(ptr, ArrayRef<Value*>(getFirsts(idxs)))),
```
zvonimir: This is not how we align (in general) when we split into multiple lines. See other files for examples.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
...
+               ArrayRef<unsigned>(getSeconds(idxs)));
+  else if (ArrayType* AT = dyn_cast<ArrayType>(ct)) {
+    for (unsigned i = AT->getNumElements(); i-- > 0; ) {
```
zvonimir: This for loop is a bit weird :). Why is `i--` not in the end as usual?

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+
+  return cv;
+}
+
+void SplitStructLoadStore::splitStructStore(StoreInst* si, Value* ptr, Value* val)
+{
+  if (StructType* st = dyn_cast<StructType>(val->getType())) {
+    std::vector<std::pair<Value*, unsigned> > idx;
```
zvonimir: Space after `unsigned` is not needed.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
...
+  else if (ArrayType* AT = dyn_cast<ArrayType>(ct)) {
+    for (unsigned i = AT->getNumElements(); i-- > 0; ) {
+      std::vector<std::pair<Value*, unsigned> > lidxs(idxs);
+      if (lidxs.empty())
+        lidxs.push_back(std::make_pair(ConstantInt::get(Type::getInt32Ty(C),0), 0));
+      lidxs.push_back(std::make_pair(ConstantInt::get(Type::getInt64Ty(C),i), i));
+      cv = buildStructs(irb, ptr, AT->getElementType(), cv, lidxs);
+    }
+  }
+  else if (StructType* ST = dyn_cast<StructType>(ct)) {
```
zvonimir: else if should be on the previous line after curly brace.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,148 @@
...
+    for (unsigned i = ST->getNumElements(); i-- > 0; ) {
+      std::vector<std::pair<Value*, unsigned> > lidxs(idxs);
+      if (lidxs.empty())
+        lidxs.push_back(std::make_pair(ConstantInt::get(Type::getInt32Ty(C),0), 0));
+      lidxs.push_back(std::make_pair(ConstantInt::get(Type::getInt32Ty(C),i), i));
+      cv = buildStructs(irb, ptr, ST->getElementType(i), cv, lidxs);
+    }
+  }
+
```
zvonimir: What happens if neither if branch nor else if branch above evaluate to true?
Good programming practice is to create an else branch and put llvm_unreachable in there as a sanity check. Grep for llvm_unreachable to see how to use it.

test/basic/split_structs.c
```diff
@@ -0,0 +1,27 @@
+#include "smack.h"
+
+//There exists a load/store instruction of type {i64, i32}
```
zvonimir: What does this comment mean? Please refine it. Also, put space after `//`.

test/basic/split_structs_fail.c
```diff
@@ -0,0 +1,27 @@
+#include "smack.h"
+
+//There exists a load/store instruction of type {i64, i32}
```
zvonimir: Look for my comment above.

lib/smack/SplitStructLoadStore.cpp
shaobo-he: I think the for loop is equivalent to ```for (unsigned i = AT->getNumElements() - 1; i >= 0; --i)```.

lib/smack/SplitStructLoadStore.cpp
shaobo-he: I think the way that this line is split makes sense to me because each argument of the top level function call is put into separate lines. It should help readability.

lib/smack/SplitStructLoadStore.cpp
shaobo-he: Thanks for the good catch. I just realized this file should be named as `SplitAggregateLoadStore` since it's supposed to handle both `struct` and `array` types. The reason I made this mistake is that it seems really hard to find test cases for `array` types. All the examples that I've seen is about `struct` types. I just removed the conditions in `splitStructLoad`, which means this pass handles both types while is not named accordingly.

zvonimir: Wait, why don't you change the file and class name then? Would it not make sense to do that and it is easy to do it?

lib/smack/SplitStructLoadStore.cpp
zvonimir: I understand they are semantically equivalent, but the way you implemented this loop is very unusual, and it is better to switch to a more standard way. In addition, LLVM coding guidelines prefer preincrement/decrement to postincrement/decrement:
http://llvm.org/docs/CodingStandards.html#prefer-preincrement
So please change this loop to what you suggested in your message above.

include/smack/SplitStructLoadStore.h
```diff
@@ -0,0 +1,30 @@
...
+
+  SplitStructLoadStore() : llvm::BasicBlockPass(ID) {}
+  virtual bool runOnBasicBlock(llvm::BasicBlock& BB);
+
+private:
+  void splitStructLoad(llvm::LoadInst* li);
+  llvm::Value* buildStructs(llvm::IRBuilder<> *irb, llvm::Value* ptr, llvm::Type* ct,
+                              llvm::Value* val, std::vector<std::pair<llvm::Value*, unsigned>> idxs);
+  void splitStructStore(llvm::StoreInst* si, llvm::Value* ptr, llvm::Value* val);
+  void copyStructs(llvm::IRBuilder<> *irb, llvm::Value* ptr, llvm::Type* ct, llvm::Value* val, std::vector<std::pair<llvm::Value*, unsigned>> idxs);
```
zvonimir: Split this long line. See LLVM coding guidelines.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,137 @@
+#include "smack/SplitStructLoadStore.h"
+#include "llvm/IR/InstIterator.h"
+
+namespace smack {
+
+using namespace llvm;
+
+std::vector<Value*> getFirsts(std::vector<std::pair<Value*, unsigned>> lst) {
+    std::vector<Value*> ret;
+    for (auto p = lst.begin(); p != lst.end(); ++p)
```
zvonimir: Do you need `begin` and `end` here? Can's you just write `for (auto p : lst)`?

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,137 @@
...
+    for (auto p = lst.begin()+1; p != lst.end(); ++p)
+      ret.push_back(std::get<1>(*p));
+    return ret;
+}
+
+bool SplitStructLoadStore::runOnBasicBlock(BasicBlock& BB) {
+  std::vector<Instruction*> toRemove;
+  for(Instruction& I : BB) {
+    if (LoadInst* li = dyn_cast<LoadInst>(&I)) {
+      bool isReturnValue = false;
```
zvonimir: Rewrite it like this:
http://llvm.org/docs/CodingStandards.html#turn-predicate-loops-into-predicate-functions

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,137 @@
+#include "smack/SplitStructLoadStore.h"
+#include "llvm/IR/InstIterator.h"
+
+namespace smack {
+
+using namespace llvm;
+
+std::vector<Value*> getFirsts(std::vector<std::pair<Value*, unsigned>> lst) {
```
zvonimir: Wait, shouldn't there be a more succinct way of doing this?
There must be a library function or something like that that helps you to extract first elements of this vector of pairs, no?
I googled it and this is one potentially nice hit: https://stackoverflow.com/questions/9094132/c-stdtransform-vector-of-pairs-first-to-new-vector

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,137 @@
...
+    std::vector<Value*> ret;
+    for (auto p = lst.begin(); p != lst.end(); ++p)
+      ret.push_back(std::get<0>(*p));
+    return ret;
+}
+
+std::vector<unsigned> getSeconds(std::vector<std::pair<Value*, unsigned>> lst) {
+    std::vector<unsigned> ret;
+    for (auto p = lst.begin()+1; p != lst.end(); ++p)
```
zvonimir: What happens when `lst.begin() == lst.end()`?? Is that possible?

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,137 @@
...
+  } else if (StructType* ST = dyn_cast<StructType>(ct)) {
+    for (unsigned i = ST->getNumElements(); i-- > 0; ) {
+      std::vector<std::pair<Value*, unsigned> > lidxs(idxs);
+      if (lidxs.empty())
+        lidxs.push_back(std::make_pair(ConstantInt::get(Type::getInt32Ty(C),0), 0));
+      lidxs.push_back(std::make_pair(ConstantInt::get(Type::getInt32Ty(C),i), i));
+      cv = buildStructs(irb, ptr, ST->getElementType(i), cv, lidxs);
+    }
+  } else
+    llvm_unreachable("Unsupported types");
```
zvonimir: This string should probably be `Unsupported type.`

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,137 @@
...
+    for (unsigned i = ST->getNumElements(); i-- > 0; ) {
+      auto A = cv? cv->getAggregateElement(i) : val;
+      std::vector<std::pair<Value*, unsigned> > lidxs(idxs);
+      if (lidxs.empty())
+        lidxs.push_back(std::make_pair(ConstantInt::get(Type::getInt32Ty(C),0), 0));
+      lidxs.push_back(std::make_pair(ConstantInt::get(Type::getInt32Ty(C),i), i));
+      copyStructs(irb, ptr, ST->getElementType(i), A, lidxs);
+    }
+  } else
+    llvm_unreachable("Unsupported types");
```
zvonimir: This string should probably be `Unsupported type`.

lib/smack/SplitStructLoadStore.cpp
shaobo-he: I don't think it's possible as long as we make sure that this pass always works on aggregate types, which has been done.

lib/smack/SplitStructLoadStore.cpp
shaobo-he: I think this function is fine because C++ does not offer simple built-in functions to do mapping from one list to another. The methods discussed in the link also require at least a function to be written.

shaobo-he:
@zvonimir Thanks for the suggestions. I've addressed most of issues that we agree on. Please take a look at it. If it's fine, then I will rename this file to reflect that fact that this PR should work on aggregate types.

lib/smack/SplitStructLoadStore.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+std::vector<unsigned> getSeconds(std::vector<std::pair<Value*, unsigned>> lst) {
+    std::vector<unsigned> ret;
+    for (auto p = lst.begin()+1; p != lst.end(); ++p)
+      ret.push_back(std::get<1>(*p));
+    return ret;
+}
+
+static bool isUsedByReturnInst(Value* v)
+{
```
zvonimir: Wrong place for this curly brace.

PR #253, opened 2017-08-15
Add support for constant structs being function returns

shaobo-he:
Consider the following example,

```LLVM
define { i32, i64 } @foo() {
  ret { i32, i64 } { i32 1, i64 2 }
}
```
It triggers an [assertion failure](https://github.com/smackers/smack/blob/master/lib/AssistDS/StructReturnToPointer.cpp#L114). This pull request intends to extend this pass to handle functions returning constant structs. It first allocates a place to hold the struct and then stores the constant into the space. For the example above, the PR converts the function body into the following form,

```LLVM
%1 = alloca { i32, i64 }
%2 = getelementptr { i32, i64 }, { i32, i64 }* %1, i32 0, i32 0
store i32 1, i32* %2
%3 = getelementptr { i32, i64 }, { i32, i64 }* %1, i32 0, i32 1
store i64 2, i64* %3        
%4 = load { i32, i64 }, { i32, i64 }* %1
ret { i32, i64 } %4
```
Note that this PR splits the struct constant once, which is unsound without PR #251.

### DIFF
--- a/lib/AssistDS/StructReturnToPointer.cpp
+++ b/lib/AssistDS/StructReturnToPointer.cpp
@@ -111,11 +111,12 @@ bool StructRet::runOnModule(Module& M) {
         if(!RI)
           continue;
         IRBuilder<> Builder(RI);
-        LoadInst *LI = dyn_cast<LoadInst>(RI->getOperand(0));
-        if (!LI) {
-          ConstantStruct* CS = dyn_cast<ConstantStruct>(RI->getReturnValue());
-          assert(CS && "Return should be preceded by a load instruction or is a constant");
-
+        if (auto LI = dyn_cast<LoadInst>(RI->getOperand(0))) {
+          Builder.CreateMemCpy(fargs.at(0),
+              LI->getPointerOperand(),
+              targetData.getTypeStoreSize(LI->getType()),
+              targetData.getPrefTypeAlignment(LI->getType()));
+        } else if (auto CS = dyn_cast<ConstantStruct>(RI->getReturnValue())) {
           StructType* ST = CS->getType();
           // We could store the struct into the allocated space pointed by the first
           // argument and then load it once SMACK can handle store inst of structs.
@@ -128,10 +129,7 @@ bool StructRet::runOnModule(Module& M) {
           assert(RI->getNumOperand == 1 && "Return should only have one operand");
           RI->setOperand(0, UndefValue::get(ST));
         } else
-          Builder.CreateMemCpy(fargs.at(0),
-              LI->getPointerOperand(),
-              targetData.getTypeStoreSize(LI->getType()),
-              targetData.getPrefTypeAlignment(LI->getType()));
+          llvm_unreachable("Unexpected struct-type return value.");
       }
     }
 
--- a/lib/AssistDS/StructReturnToPointer.cpp
+++ b/lib/AssistDS/StructReturnToPointer.cpp
@@ -110,9 +110,26 @@ bool StructRet::runOnModule(Module& M) {
         ReturnInst * RI = dyn_cast<ReturnInst>(I++);
         if(!RI)
           continue;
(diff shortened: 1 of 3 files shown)


michael-emmi:
Wouldn’t it be simpler to just store the constant into the target address `fargs.at(0)` using a single `store` instruction, rather than allocating memory, storing, the struct, then loading it again, and then doing a memcpy to the target address? That is 1 memory operation versus at least 4: alloc, store(s), load, memcpy.

Also, now that you’re handing two cases, `load` and `constant`, the logic should probably be “if load then … else if constant then … else fail with unexpected instruction preceding return.”

shaobo-he:
@michael-emmi Thanks again for your suggestions. Yes, using `store` instructions is a better solution than my original idea. I have updated the file accordingly. With respect to the logic, I think the current version is equivalent to what you proposed.

lib/AssistDS/StructReturnToPointer.cpp
```diff
@@ -110,13 +110,28 @@ bool StructRet::runOnModule(Module& M) {
...
           continue;
-        LoadInst *LI = dyn_cast<LoadInst>(RI->getOperand(0));
-        assert(LI && "Return should be preceded by a load instruction");
         IRBuilder<> Builder(RI);
-        Builder.CreateMemCpy(fargs.at(0),
-            LI->getPointerOperand(),
-            targetData.getTypeStoreSize(LI->getType()),
-            targetData.getPrefTypeAlignment(LI->getType()));
+        LoadInst *LI = dyn_cast<LoadInst>(RI->getOperand(0));
+        if (!LI) {
```
shaobo-he: Yes, the code in your comment is clearer. I've refactored the code to adopt it.

lib/AssistDS/StructReturnToPointer.cpp
```diff
@@ -110,13 +110,26 @@ bool StructRet::runOnModule(Module& M) {
...
+          // We could store the struct into the allocated space pointed by the first
+          // argument and then load it once SMACK can handle store inst of structs.
+          for (unsigned i = 0; i < ST->getNumElements(); ++i) {
+            std::vector<Value*> idxs = {ConstantInt::get(Type::getInt32Ty(CS->getContext()),0),
+                ConstantInt::get(Type::getInt32Ty(CS->getContext()),i)};
+            Builder.CreateStore(CS->getAggregateElement(i),
+                Builder.CreateGEP(fargs.at(0), ArrayRef<Value*>(idxs)));
+          }
+          assert(RI->getNumOperand == 1 && "Return should only have one operand");
+          RI->setOperand(0, UndefValue::get(ST));
```
michael-emmi: Why is the return value set to undefined when it was originally a constant?

Why isn’t the return value set to undefined when it was originally from a load instruction?

shaobo-he: If we don't replace the constant with an undefined value, then SMACK crashes because of this line: https://github.com/smackers/smack/blob/develop/lib/smack/SmackRep.cpp#L744, which is not the case if the return value is from a load instruction.

test/basic/struct_const_return_fail.c
```diff
@@ -0,0 +1,23 @@
+#include "smack.h"
+
+// @flag --clang-options=-O1
+// @expect error 
+
+typedef struct
+{
```
zvonimir: Wrong code formatting of open curly braces. Please fix this everywhere.

test/basic/struct_const_return.c
```diff
@@ -0,0 +1,23 @@
...
+typedef struct
+{
+  int a;
+  long b;
+} S;
+
+S foo()
+{
+  S x = {1, 2L};
+  assert(1);
```
zvonimir: Why do you have `assert(1)` here?

shaobo-he: It is used to prevent clang to optimize function `foo` away.

Issue #232, opened 2017-06-06
Current memory model is broken with the flag --bit-precise-pointers


shaobo-he:
If we use bit-vectors to model pointers, then the memory model we are using allows return values of memory allocation procedures to alias. Recall the definition of procedure `$$alloc`,

```boogie
procedure {:inline 1} $$alloc(n: ref) returns (p: ref)
modifies $CurrAddr;
{
  p := $CurrAddr;
  havoc $CurrAddr;
  if ($sgt.ref.bool(n, $0.ref)) {
    assume $sle.ref.bool($add.ref(p, n), $CurrAddr);
  } else {
    assume $sle.ref.bool($add.ref(p, $1.ref), $CurrAddr);
  }
}
```

Let's say there are three calls to it,

first time : `n = 4; p = 1024bv32; $CurrAddr = 2147483647bv32`
second time: `n = 4; p = 2147483647bv32; $CurrAddr = 2147483647bv32`
third time: `n = 4; p = 2147483647bv32; $CurrAddr=0`

The return pointer of the second call is equal to that of the third call because `$add.ref(2147483647bv32,4bv32)` is negative.

To reproduce this issue, consider the following C program,
```C
#include "smack.h"
#include <stdlib.h>
int zero() {return 0;}
int main()
{
  int* x = (int*)malloc(sizeof(int));
  int* y = (int*)malloc(sizeof(int));
  int* z = (int*)malloc(sizeof(int));
  if (zero())
    y = z;
  *y = 1;
  *z = 2;
  assert(*y == 1);
  return 0;
}
```

PR #255, opened 2017-08-19
Fix memory model with flag --bit-precise and --bit-precise-pointers enabled

shaobo-he:
This pull request intends to fix the issue described in issue #232. The root cause of the issue is that when signed addition overflows, the result is negative. The proposed solution is to replace the bit-vector addition with a semantically equivalent subtraction with unbounded integers such that overflow is avoided.


zvonimir:
Two things:
- Squash these 3 commits into one nicer commit
- Write a nice commit message for this one commit and in the end of the message say what you are fixing. In fact, message me on slack your commit message before you actually commit it.
Otherwise, this looks good to me.

PR #256, opened 2017-08-22
Updated Rust Support

keram88:
This should merge into the rust-support branch. As this fork works and is much newer than the old branch, this pull request should overwrite the old files when merging.


michael-emmi:
There are an awful lot of changes here relative to the current develop, and plenty of conflicts.

Can you figure out how to base your actual changes on the current develop branch?

zvonimir:
Sorry, this was my mistake. We are trying to create this pull request on top of a very deprecated branch, which was not a good idea. So we should start over from a newly created branch, which should resolve a bunch of these issues.

PR #259, opened 2017-09-08
Fix unsound region splitting.

### LINKED ISSUE #258: Unsound region assignment with nested structs
Given the following code, Smack somehow assigns distinct regions to the memory accesses of functions `f` and `g` even though they ought to alias due to identical call-site arguments `&t.z`; both assertions ought to hold, but the first fails given given Smack’s region assignment.
````
#include <smack.h>

struct S {
  int w;
};

struct T {
  int x;
  struct S y;
  struct S z;
};

void f(struct S *state)
{
  state->w = 0;
}

void g(struct S *state)
{
  state->w = 1;
}

int main(void) {
  struct T t;

  f(&t.y);
  f(&t.z);
  g(&t.z);
  	
  assert(t.y.w == 0);
  // ALSO: assert(t.z.w == 1);

  return 0;
}
````

michael-emmi:
* when some equivalent DSA nodes have distinct collapsed status
* fixes #258

### DIFF
--- a/include/smack/DSAWrapper.h
+++ b/include/smack/DSAWrapper.h
@@ -68,7 +68,7 @@ private:
   std::vector<const llvm::DSNode*> collectMemcpys(llvm::Module &M, MemcpyCollector* mcc);
   std::vector<const llvm::DSNode*> collectStaticInits(llvm::Module &M);
   llvm::DSGraph *getGraphForValue(const llvm::Value *V);
-  unsigned getOffset(const MemoryLocation* l);
+  int getOffset(const MemoryLocation* l);
 
 public:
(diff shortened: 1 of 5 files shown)


michael-emmi:
Can we merge this?

zvonimir:
@shaobo-he : please run this on SVCOMP as soon as you can. I am a bit worried that this might (significantly) influence the scalability of SMACK, and so it would be good to double-check that.

shaobo-he:
@zvonimir Ok. I will run it right now.

michael-emmi:
Great point; completely forgot to add regressions; added now; will merge after CI success.

PR #265, opened 2017-09-10
Portable library models.

### LINKED ISSUE #238: Memory safety regressions failing on OS X
All memory-safety regression tests fail (with `UNKNOWN` result) on OS X.

The problem is observed as follows (skipping generated warnings):
````
$ smack memory-safety/adjacent.c --memory-safety
SMACK program verifier version 1.8.0
Traceback (most recent call last):
  File "/usr/local/bin/smack", line 13, in <module>
    smack.top.main()
  File "/usr/local/bin/../share/smack/top.py", line 646, in main
    frontend(args)
  File "/usr/local/bin/../share/smack/top.py", line 270, in frontend
    return frontends()[lang](args)
  File "/usr/local/bin/../share/smack/top.py", line 332, in clang_frontend
    try_command(['llvm-link', '-o', args.linked_bc_file, args.bc_file] + build_libs(args))
  File "/usr/local/bin/../share/smack/top.py", line 301, in build_libs
    try_command(default_clang_compile_command(args, True) + ['-o', bc, c])
  File "/usr/local/bin/../share/smack/utils.py", line 69, in try_command
    raise Exception(output)
...
/usr/local/share/smack/lib/smack.c:2240:7: error: expected parameter declarator
char *strcpy(char *dest, const char *src) {
      ^
/usr/include/secure/_string.h:83:38: note: expanded from macro 'strcpy'
  __builtin___strcpy_chk (dest, src, __darwin_obsz (dest))
                                     ^
/usr/include/secure/_common.h:39:62: note: expanded from macro '__darwin_obsz'
#define __darwin_obsz(object) __builtin_object_size (object, _USE_FORTIFY_LEVEL > 1 ? 1 : 0)
                                                             ^
/usr/include/secure/_common.h:30:32: note: expanded from macro '_USE_FORTIFY_LEVEL'
#    define _USE_FORTIFY_LEVEL 2
                               ^
/usr/local/share/smack/lib/smack.c:2240:7: error: expected ')'
/usr/include/secure/_string.h:83:38: note: expanded from macro 'strcpy'
  __builtin___strcpy_chk (dest, src, __darwin_obsz (dest))
                                     ^
/usr/include/secure/_common.h:39:62: note: expanded from macro '__darwin_obsz'
#define __darwin_obsz(object) __builtin_object_size (object, _USE_FORTIFY_LEVEL > 1 ? 1 : 0)
                                                             ^
/usr/include/secure/_common.h:30:32: note: expanded from macro '_USE_FORTIFY_LEVEL'
#    define _USE_FORTIFY_LEVEL 2
                               ^
/usr/local/share/smack/lib/smack.c:2240:7: note: to match this '('
/usr/include/secure/_string.h:83:38: note: expanded from macro 'strcpy'
  __builtin___strcpy_chk (dest, src, __darwin_obsz (dest))
                                     ^
/usr/include/secure/_common.h:39:53: note: expanded from macro '__darwin_obsz'
#define __darwin_obsz(object) __builtin_object_size (object, _USE_FORTIFY_LEVEL > 1 ? 1 : 0)
                                                    ^
...
/usr/local/share/smack/lib/smack.c:2240:7: error: conflicting types for '__builtin___strcpy_chk'
/usr/include/secure/_string.h:83:3: note: expanded from macro 'strcpy'
  __builtin___strcpy_chk (dest, src, __darwin_obsz (dest))
  ^
/usr/local/share/smack/lib/smack.c:2240:7: note: '__builtin___strcpy_chk' is a builtin with type 'char *(char *, const char *, unsigned long)'
/usr/include/secure/_string.h:83:3: note: expanded from macro 'strcpy'
  __builtin___strcpy_chk (dest, src, __darwin_obsz (dest))
  ^
/usr/local/share/smack/lib/smack.c:2240:7: error: definition of builtin function '__builtin___strcpy_chk'
char *strcpy(char *dest, const char *src) {
      ^
/usr/include/secure/_string.h:83:3: note: expanded from macro 'strcpy'
  __builtin___strcpy_chk (dest, src, __darwin_obsz (dest))
  ^
...
14 warnings and 4 errors generated.
````

### PR DESCRIPTION
**Pull request by michael-emmi**
* E.g., for <string.h> and <math.h>
* Avoids conflicts with system headers by hijacking include path.
* Replacement header for `$MODULE` in `share/smack/include/$MODULE/$MODULE.h`.
* Replacement model for `$MODULE` in `share/smack/lib/$MODULE.c`.
* Only use custom headers when models are used too.
* Fixes #238.

### CODE REVIEW

**Comment by zvonimir** [comment]
@shaobo-he : could you run this on SVCOMP to make sure all is good?

File: share/smack/top.py [hunk 1]
```diff
@@ -305,6 +314,12 @@ def build_libs(args):
   if args.pthread:
     libs += ['pthread.c']
 
+  if args.memory_safety:
```
shaobo-he: Is there a particular reason that the definitions are linked only when memory safety is being checked? We need them to check signed integer overflows, too. Can we always link them?

michael-emmi: My worry is that linking these stubs may be a negative impact on verification scalability. I am supposing that many programs/assertions do not rely, e.g., on the details of string operations. Therefore I am in favor of conditional linkage. Admittedly, I have no data to support this opinion :-)

One solution to your problem is to link this for signed integer overflows too. Please feel free add that as an enabling condition.

In the long run though, I think a more generic mechanism for linking stubs will be useful/necessary. For instance, besides enabling or disabling a single predefined stub for `string.h` functions, we might want to choose between different stubs of the same `string.h` functions, e.g., which have weaker specifications yet are easier for the verifier to reason about. Or, for instance, we may want to link user-defined stubs.

zvonimir: For now, I agree with @michael-emmi : let's just add integer overflow flag in addition to memory safety to this check. @shaobo-he : can you please do this ASAP and then test this branch again?

In the long run, we could introduce a command like option called `link-models` or something like that, which we would use to select which models are linked in. Something like `link-models=pthreads,string` would link in the respective models. Then for SVCOMP we can enable those if particular flags are selected. I will create a separate issue for this.

shaobo-he: Done.

share/smack/lib/smack.c
```diff
@@ -3,8 +3,7 @@
 
 #include <smack.h>
 #include <limits.h>
-#include <string.h>
-#include <stdlib.h>
```
shaobo-he: @michael-emmi Is there a particular reason that `stdlib.h` is taken away? `malloc` is used in `smack.c`.

shaobo-he:
There is another minor issue that I couldn't figure out -- it seems that some functions that should have been removed by the dead declaration removal pass still exist. For example, in file [`termination-crafted-lit/AliasDarteFeautrierGonnord-SAS2010-loops_true-termination_false-no-overflow.c`](http://node0.smackbenchwebserver.smack.emulab.net/sv-benchmarks/c/termination-crafted-lit/AliasDarteFeautrierGonnord-SAS2010-loops_true-termination_false-no-overflow.c), function `__VERIFIER_nondet_char` appears in the bpl file while it is not actually used in the input c file. The develop branch does not have this issue. I spent some time in this example and found that there seems to exist a user of this function that never gets removed.

To reproduce, try `smack -x svcomp --verifier svcomp AliasDarteFeautrierGonnord-SAS2010-loops_true-termination_false-no-overflow.c -signed-integer-overflow -bpl test.bpl`

share/smack/lib/smack.c
michael-emmi: Fair point; this will probably lead to unnecessary warnings; adding back the `stdlib` include.

michael-emmi:
Hey @shaobo-he, good catch; pushed a fix; should work now.

The reason was rather subtle. We’d been giving incomplete declarations for the `nondet` functions, i.e., declaring them in `smack.h` as
````c
char __VERIFIER_nondet_char();
````
rather than
````c
char __VERIFIER_nondet_char(void);
````
These are actually two different things in C: the first declares a function with unknown argument types, while the second specifically says there are zero arguments.

When `smack.c` included both the string functions using these `nondet` functions and the definitions of `nondet` functions themselves, the incomplete declarations would be resolved completely in `smack.c`. When I separated `string.c` from `smack.c`, however,  these declarations were not resolved, and thus `string.c` did not know that `nondet` functions have `void` argument types. This caused bit casting in their invocations, introducing additional bit-cast constants in the bitcode. Subsequently, when dead function definitions were removed, the corresponding bit-cast constants were not removed, making it look like the bit-casted functions had users.

The patch I just pushed avoids introducing these bit-cast constants in the first place, by giving complete argument-type declarations in `smack.h`.

PR #266, opened 2017-09-18
Added modifies clause generation for Boogie files.

liammachado:
Modifies clause generation is now directly integrated into the SMACK tool, rather than being done by Corral/Boogie.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,6 +167,7 @@ add_library(smackTranslator STATIC
   include/smack/SmackRep.h
   include/smack/MemorySafetyChecker.h
   include/smack/SignedIntegerOverflowChecker.h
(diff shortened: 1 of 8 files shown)


michael-emmi:
Before getting into specifics, some very high-level comments:

1. This PR should be based on `develop` not `master`.

2. The `GenModifies` module is not using RTTI correctly. It seems to have been set up properly, according to https://llvm.org/docs/HowToSetUpLLVMStyleRTTI.html but for some reason many would-be `isa` and `dyn_cast` operations are replaced by terrible things like `stmt->getKind() == 5`.

3. The `GenModifies` module is not readable nor modular. We should follow basic principles of software engineering so that this code can be maintained going forward.

PR #270, opened 2017-09-21
Add support i1 and bv1 for languages that use these types

keram88:
Adds support for the 1-bit integer type for regular integers, and sign extension/truncation for 1-bit bit-vectors. The Rust compiler emits i1 and bv1 (in bv mode) for the boolean type. Other languages may require this feature as well.

### DIFF
--- a/share/smack/lib/smack.c
+++ b/share/smack/lib/smack.c
@@ -1407,6 +1407,7 @@ void __SMACK_decls(void) {
   D("function {:inline} $load.bv24(M: [ref] bv24, p: ref) returns (bv24) { M[p] }");
   D("function {:inline} $load.bv16(M: [ref] bv16, p: ref) returns (bv16) { M[p] }");
   D("function {:inline} $load.bv8(M: [ref] bv8, p: ref) returns (bv8) { M[p] }");
+  D("function {:inline} $load.bv1(M: [ref] bv1, p: ref) returns (bv1) { M[p] }");
 
   D("function {:inline} $load.bytes.bv128(M: [ref] bv8, p: ref) returns (bv128)"
     "{ $load.bytes.bv64(M, $add.ref(p, $8.ref)) ++ $load.bytes.bv64(M, p) }");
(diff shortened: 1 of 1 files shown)


zvonimir:
First of all, please write better commit messages in the future. Read our contribution guidelines on commit messages. This time around you can leave it as is.
Second, would it be possible to add a C regression that loads/stores i1? I suspect that is not easy and I don't see how to do it. Maybe try bit-fields?
Otherwise, this looks good.

keram88:
A Clang-specific regression is to use the **__builtin_add_overflow** intrinsic, or a related checked-arithmetic intrinsic. This produces code similar to
```
  %8 = call { i32, i1 } @llvm.uadd.with.overflow.i32(i32 %6, i32 %7)
  %9 = extractvalue { i32, i1 } %8, 1
  %10 = extractvalue { i32, i1 } %8, 0
```

"extractvalue { i32, i1 } %8, 1" is transformed into "load i1 ..." in the StructReturnToPointer pass. However, this requires simultaneous support for the llvm overflow intrinsics.

shaobo-he:
LGTM

share/smack/lib/smack.c
```diff
@@ -1455,6 +1456,7 @@ void __SMACK_decls(void) {
   D("function {:inline} $store.bv24(M: [ref] bv24, p: ref, v: bv24) returns ([ref] bv24) { M[p := v] }");
   D("function {:inline} $store.bv16(M: [ref] bv16, p: ref, v: bv16) returns ([ref] bv16) { M[p := v] }");
   D("function {:inline} $store.bv8(M: [ref] bv8, p: ref, v: bv8) returns ([ref] bv8) { M[p := v] }");
+  D("function {:inline} $store.bv8(M: [ref] bv1, p: ref, v: bv1) returns ([ref] bv1) { M[p := v] }");
```
shaobo-he: A typo. Should be `$store.bv1`.

PR #273, opened 2017-10-08
Adds support for the Symbooglix backend verifier.

liammachado:


zvonimir:
@liammachado : let's start by squashing all these small commits you have into just one commit. then write a nice commit message by following our guidelines for writing commit messages. Another thing - let's not run symbooglix on all regressions by default. Clearly we will not be able to make it run reliably that soon, and it would be good to merge this. So let's leave it out of our regression runs (for now).

Issue #174, opened 2016-05-10
Avoid havoc command for the encoding of select instruction


shaobo-he:
The alias analysis for Boogie programs used by angelic verification does not support **havoc** command which is part of the current encoding of select instruction. Recall the current encoding **havoc**s the return value and assumes the return value equal to two possible operands depending on the condition (https://github.com/smackers/smack/blob/develop/lib/smack/SmackInstGenerator.cpp#L422-L435).

I was wondering if the **havoc** command is needed here since LLVM uses SSA form, which means the return variable is not assigned before the select instruction. Another solution is to assign `if else` expression to the return variable.

michael-emmi:
SSA means single *static* assignment; that variable can be assigned multiple times in a loop.

Translating
````
z = select w, x, y
````
to
````
z := if w == 1 then x else y;
````
as you suggest sounds like a better idea for now.

But this begs a more serious question: why doesn’t that verification tool budge on this one? Last I knew `havoc` was a valid Boogie statement. Should we not be writing Boogie programs with the `havoc` statement? Since when was `havoc` blacklisted? Is Smack forever prohibited from emitting the `havoc` statement?

Any insight/elaboration on that would be useful. What does MSR say?

akashlal:
Testing

michael-emmi:
@akashlal Agreed, and I prefer Shaobo's translation anyways. Just making sure we still have the right to use `havoc` if needed 🙂

michael-emmi:
@Guoanshisb did you want me to implement this, or will you?

Guoanshisb:
@michael-emmi Let me try to implement it in `av-support` branch first. Then you can review my implementation after a pull request is submitted.

zvonimir:
@shaobo-he : has this been done?

shaobo-he:
The new encoding is implemented in `av-support` branch.

zvonimir:
Could we cherry-pick the needed commits or something like that, so that we can address this issue sooner rather than later? Or are you planning to merge av-support sometime soon?

shaobo-he:
I created a new pull request (#277) to address the issue. Please take a look at it.

PR #277, opened 2017-10-27
Rewrite select instruction encoding

shaobo-he:
The old encoding translates a select instruction to `havoc(x); assume(c==1==>x==v1 && c!=1==>x==v2);` where `c` is the condition and `v1`, `v2` is true value and false value, respectively.

The new encoding translates it to `x = if c ==1 then v1 else v2`.

This pull request addresses issue #174.

### DIFF
--- a/lib/smack/SmackInstGenerator.cpp
+++ b/lib/smack/SmackInstGenerator.cpp
@@ -490,15 +490,13 @@ void SmackInstGenerator::visitSelectInst(llvm::SelectInst& i) {
   processInstruction(i);
   std::string x = naming->get(i);
   const Expr
-  *c = rep->expr(i.getOperand(0)),
-   *v1 = rep->expr(i.getOperand(1)),
-    *v2 = rep->expr(i.getOperand(2));
+  *c = rep->expr(i.getCondition()),
+   *v1 = rep->expr(i.getTrueValue()),
+    *v2 = rep->expr(i.getFalseValue());
 
-  emit(Stmt::havoc(x));
-  emit(Stmt::assume(Expr::and_(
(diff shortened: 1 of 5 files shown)


zvonimir:
Do we have a regression for this already?

shaobo-he:
No, I don't. I plan to try it on all SV-COMP benchmarks. I'm running the commit that merges JJ's pull request in.

zvonimir:
Could you write a simple C program that generates select instructions?

shaobo-he:
Yes, I can. I thought Clang always compiles conditional expressions into multiple blocks with optimization level 0.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -490,15 +490,20 @@ void SmackInstGenerator::visitSelectInst(llvm::SelectInst& i) {
...
   std::string x = naming->get(i);
   const Expr
-  *c = rep->expr(i.getOperand(0)),
-   *v1 = rep->expr(i.getOperand(1)),
-    *v2 = rep->expr(i.getOperand(2));
+  *c = rep->expr(i.getCondition()),
+   *v1 = rep->expr(i.getTrueValue()),
+    *v2 = rep->expr(i.getFalseValue());
 
+  /*
```
zvonimir: Why don't you completely erase this instead of just commenting it out?

shaobo-he: It's not yet known if the new encoding would influence performance. So I keep the old encoding for reference.

zvonimir: Ok, got it, well, we should not merge until we figure it out and you erase it.

PR #278, opened 2017-11-06
Add support for recording function argument values

shaobo-he:
Currently SMACK doesn't emit record procedures for function arguments.  The `llvm.dbg.value` calls corresponding to function arguments are in the beginning of the entry block in a function. Source location annotations are not generated for these call instructions which must precede record procedure calls in order for them to work. This PR adds source location of the first non-debug instruction to the beginning of each block created.

Then values of non pointer function arguments will be printed as `function-name`:`arg`:`argument-name`.

### DIFF
--- a/include/smack/SmackInstGenerator.h
+++ b/include/smack/SmackInstGenerator.h
@@ -30,7 +30,6 @@ private:
 
   Block* currBlock;
   llvm::BasicBlock::const_iterator nextInst;
-  llvm::Instruction* firstNonDebugInst;
   std::map<const llvm::BasicBlock*, Block*> blockMap;
   std::map<const llvm::Value*, std::string> sourceNames;
 
@@ -51,7 +50,7 @@ public:
 
 public:
   SmackInstGenerator(llvm::LoopInfo& LI, SmackRep* R, ProcDecl* P, Naming* N)
-    : loops(LI), rep(R), proc(P), naming(N), firstNonDebugInst(nullptr) {}
+    : loops(LI), rep(R), proc(P), naming(N) {}
 
 
   void visitBasicBlock(llvm::BasicBlock& bb);
--- a/include/smack/SmackInstGenerator.h
+++ b/include/smack/SmackInstGenerator.h
@@ -30,7 +30,7 @@ private:
 
   Block* currBlock;
   llvm::BasicBlock::const_iterator nextInst;
-  llvm::Instruction* fstNonDebugInst;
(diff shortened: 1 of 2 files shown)


include/smack/SmackInstGenerator.h
```diff
@@ -30,6 +30,7 @@ class SmackInstGenerator : public llvm::InstVisitor<SmackInstGenerator> {
 
   Block* currBlock;
   llvm::BasicBlock::const_iterator nextInst;
+  llvm::Instruction* fstNonDebugInst;
```
zvonimir: Please name this variable `firstNonDebugInst`.
Also, shouldn't you initialize it to null or something like that?

shaobo-he: Done. Yes, I should have initialized it to null.

michael-emmi:
Can you explain the purpose of the `firstNonDebugInst ` variable?

shaobo-he:
@michael-emmi I think the purpose is to not annotate the first non debug instruction twice by recording it.

michael-emmi:
@shaobo-he can you please elaborate? e.g., are you saying `annotate` will be called twice on some instruction otherwise? e.g., why would that happen?

shaobo-he:
Yes, `annotate` will be called on the first non debug instruction twice without using this variable. In `visitBasicBlock` the first non debug instruction is searched and annotated. Then the same instruction is annotated again when it is visited.

michael-emmi:
@shaobo-he I still do not understand this issue very well. I think you could help me by explaining this carefully, using a very simple example, to show precisely what problem your patch is solving, and then, separately, precisely how your solution addresses this problem.

In case you don’t have time for that, then let me ask what you meant by:
> Source location annotations are not generated for these call instructions which must precede record procedure calls in order for them to work

What is it that does not work?

And why is it that we do not generate annotations for those instructions?

Also: why are we not combining the existing `smack:arg` scheme with your more general `smack:fn:arg` scheme?

michael-emmi:
@shaobo-he is there any harm annotating an instruction twice? the `firstNonDebugInst` variable seems confusing and unnecessary. I would be happy to merge this PR without that variable.

shaobo-he:
@michael-emmi sorry for the late reply. Avoiding annotating an instruction twice is just to make the output Boogie program smaller or cleaner. Performance-wise, I don't think it matters. So I removed it.

michael-emmi:
Ok this looks better now.

PR #280, opened 2017-11-16
Update Boogie and Corral links

Hirse:
This PR updates the links to point to their GitHub repositories instead of Codeplex.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -51,6 +51,6 @@ tools and packages, which come with their own licenses:
 - LLVM, clang, LLVM runtime (http://llvm.org/)
 - mono (http://www.mono-project.com/)
 - Boogie (https://github.com/boogie-org/boogie/)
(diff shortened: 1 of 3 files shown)


zvonimir:
Thanks for this patch!
Two things:
- The develop branch is where all the action happens. So please create a pull request to the develop branch.
- I just noticed that we are using an old link to Corral at a couple of other places as well in our documentation. Hopefully you could grep for those and fix it everywhere to make it consistent.
Thanks!

Hirse:
Sure. 
Sorry, I didn't check for the branches.

zvonimir:
I think you will have to close this pull request and create a new one on top of the develop branch. I don't think there is a way to change it at this point.

zvonimir:
Closing without merging since it is pulling from master.

PR #283, opened 2017-11-17
Add Limited Objective-C support to SMACK

jjgarzella:
Adds processing in top.py to handle very basic examples of Objective-C. Support is not fully-featured, but I think it's worth it to pull this in soon, if only to properly define how top.py will look with many programming languages.


share/smack/top.py
```diff
@@ -296,6 +297,12 @@ def smack_headers():
 def smack_lib():
   return os.path.join(smack_root(), 'share', 'smack', 'lib')
 
+def objc_clang_compile_command(args, lib = False):
+  cmd = default_clang_compile_command(args,lib)
+  objc_flags = try_command(['gnustep-config', '--objc-flags'])
```
michael-emmi: Is this part specific to Linux? it would be good to have something that works on OS X too, especially given that a Objective C is an Apple-centric technology.

share/smack/top.py
```diff
@@ -344,8 +351,17 @@ def llvm_frontend(args):
...
 
-  bitcodes = []
   compile_command = default_clang_compile_command(args)
+  clang_frontend_helper(compile_command, args)
+
+def objc_clang_frontend(args):
+  """Generate Boogie code from Objective-C language source(s)."""
+  compile_command = objc_clang_compile_command(args)
+  clang_frontend_helper(compile_command, args)
+
```
zvonimir: Please remove one new line here.

share/smack/top.py
jjgarzella: Yes, this part is specific to Linux. For macOS support, a different addition will be necessary. I agree that it will be useful, though, it's just not part of this PR

jjgarzella:
I think I should clarify the "limited support" that this PR gives: SMACK can now verify single-file objective-c programs (in a .m), on Linux once the GNUstep objective-c runtime has been installed.

share/smack/top.py
zvonimir: I kind of understand why we need gnustep-config - to be able to determine which flags are needed to compile Objective C programs in Linux.
How different are Linux flags from OS X flags?

jjgarzella: From what I can tell, for this simple case, the default command that works for C will work, but if we are linking multiple files "-framework Foundation" may be required. Let me emphasize that I have NOT tested this.

zvonimir: Check this out: https://stackoverflow.com/questions/110362/how-can-i-find-the-current-os-in-python
If you invoke `platform.system` in python, it is easy to find out if you are running in Linux or OS X.
So maybe that would be a relatively easy way to make this code cross-platform enabled. Thoughts?

jjgarzella: I've added this below... left the OS X section blank for now.

share/smack/top.py
```diff
@@ -296,6 +297,15 @@ def smack_headers():
 def smack_lib():
   return os.path.join(smack_root(), 'share', 'smack', 'lib')
 
+def objc_clang_compile_command(args, lib = False):
+  cmd = default_clang_compile_command(args,lib)
+  if sys.platform in ['linux', 'linux2']:
+    objc_flags = try_command(['gnustep-config', '--objc-flags'])
+    cmd += objc_flags.split()
+  elif sys.platform == 'darwin'
```
zvonimir: I think you have a syntax error here - missing `:`.

share/smack/top.py
```diff
@@ -296,6 +297,15 @@ def smack_headers():
...
+    pass # add future macOS modifications here.
```
zvonimir: Instead of `pass`, you should do `sys.exit` here with an appropriate message. Search for `exit` in this file for an example.

share/smack/top.py
zvonimir: All regressions are failing, and I would guess it is because of this.

bin/build.sh
```diff
@@ -451,4 +454,17 @@ then
   puts "Regression tests complete"
 fi
 
+if [ ${INSTALL_OBJECTIVEC} -eq 1 ]
+then
+  puts "Installing GNU Objective-C for linux"
+
+  sudo apt-get -y install gobjc gnustep gnustep-make gnustep-common gnustep-devel
```
shaobo-he: I think we'd probably want to rewrite this line like this https://github.com/smackers/smack/blob/develop/bin/build.sh#L187.

bin/build.sh
```diff
@@ -451,4 +454,17 @@ then
...
+  # The version numbers here will have to change by OS.
+  sudo ln -s /usr/lib/gcc/x86_64-linux-gnu/4.8/include/objc /usr/local/include/objc
```
shaobo-he: May I ask why a symbolic link is required here?

bin/build.sh
```diff
@@ -451,4 +454,17 @@ then
...
+  echo ". /usr/share/GNUstep/Makefiles/GNUstep.sh" >> ~/.bashrc
```
shaobo-he: If this step is required, I think it's better we put `. /usr/share/GNUstep/Makefiles/GNUstep.sh` into `smack.environment`.

share/smack/top.py
```diff
@@ -297,6 +298,17 @@ def smack_headers():
 def smack_lib():
   return os.path.join(smack_root(), 'share', 'smack', 'lib')
 
+def objc_clang_compile_command(args, lib = False):
+  cmd = default_clang_compile_command(args,lib)
+  if sys.platform in ['linux', 'linux2']:
+    objc_flags = try_command(['gnustep-config', '--objc-flags'])
+    cmd += objc_flags.split()
+  elif sys.platform == 'darwin':
+    sys.exit("Objective-C not yet supported on macOS")
```
shaobo-he: We should merge the two else branches as they mean the same.

share/smack/top.py
```diff
@@ -345,8 +357,19 @@ def llvm_frontend(args):
...
   compile_command = default_clang_compile_command(args)
+  clang_frontend_helper(compile_command, args)
+
+def objc_clang_frontend(args):
+  """Generate Boogie code from Objective-C language source(s)."""
+
+  compile_command = objc_clang_compile_command(args)
+  clang_frontend_helper(compile_command, args)
+
+def clang_frontend_helper(compile_command, args):
```
shaobo-he: I think maybe it's better to name this function as something like `link-bc-files`.

share/smack/top.py
jjgarzella: If I remember right, this was suggested by Zvonimir or Micheal Emmi, but the comments disappeared after a force push... I can merge them if we want.

share/smack/top.py
jjgarzella: I agree. It looks like this function doesn't have anything to do with clang, and that it will be used by other languages. It looks like the llvm frontend and the rust frontend don't use it though, so I propose `default_link_bc_files`.

bin/build.sh
jjgarzella: The way that apt installs the objc libraries, it puts the objc.h file into a place where clang can't find it. Rather than make clang include this directory, I symlink it into the include folder so it sees it by default (this was the internet's solution to this error: https://ubuntuforums.org/showthread.php?t=1412542)

shaobo-he: I think we need to fix this otherwise SMACK's Objective-C support cannot be locally installed. Maybe this link is helpful, https://stackoverflow.com/questions/11475124/unable-to-find-objc-h-during-clang-compile.

bin/build.sh
jjgarzella: Is smack.environment a generated file? I can't find it in the repo

shaobo-he: Yes. It's generated by `build.sh`. See this: https://github.com/smackers/smack/blob/develop/bin/build.sh#L48.

bin/build.sh
```diff
@@ -249,7 +258,7 @@ then
     sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
     echo "deb http://download.mono-project.com/repo/ubuntu trusty main" | sudo tee /etc/apt/sources.list.d/mono-official.list
     sudo apt-get update
-    sudo apt-get install -y ${DEPENDENCIES}
+    sudo apt-get install -y --force-yes ${DEPENDENCIES}
```
zvonimir: Wait, aren't `-y` and `--force-yes` synonyms?

bin/build.sh
```diff
@@ -216,6 +219,12 @@ do
   esac
 done
 
+if [ ${INSTALL_OBJECTIVEC} -eq 1 ]
+then
+
```
zvonimir: Remove this new line.

bin/build.sh
```diff
@@ -216,6 +219,12 @@ do
...
+  DEPENDENCIES+=" gobjc gnustep gnustep-make gnustep-common gnustep-devel"
+
```
zvonimir: Remove this new line.

bin/build.sh
```diff
@@ -451,4 +460,16 @@ then
   puts "Regression tests complete"
 fi
 
+if [ ${INSTALL_OBJECTIVEC} -eq 1 ]
```
zvonimir: This should not be in the end. Move it just before `BUILD_Z3`.

bin/build.sh
```diff
@@ -451,4 +460,16 @@ then
...
+then
+  puts "Installing GNU Objective-C for linux"
+
+  # The version numbers here will have to change by OS.
+  sudo ln -s /usr/lib/gcc/x86_64-linux-gnu/4.8/include/objc /usr/local/include/objc
+  echo ". /usr/share/GNUstep/Makefiles/GNUstep.sh" >> ${SMACKENV} 
+  source ${SMACKENV}
```
zvonimir: Once you move this up, there is no need to source `SMACKENV` here any more.

bin/build.sh
jjgarzella: [Actually, no](https://linux.die.net/man/8/apt-get). One of the Objective-C packages does something deemed undesirable by apt, so simply using -y does not work here. I'm not sure exactly which package does it - this is something I'd like to look into when I fix the symlink in the future.

zvonimir: Oh, man, here is what man pages say about this:
```
Force yes. This is a dangerous option that will cause apt-get to continue without prompting if it is doing something potentially harmful. It should not be used except in very special situations. Using --force-yes can potentially destroy your system!
```
I would like to avoid having `force-yes` here. A workaround would be that you do not extend `DEPENDENCIES` but rather install them separately in the `INSTALL_OBJECTIVEC` if condition. Does that make sense?

jjgarzella: Yes, this is the original strategy I was using before the review. I will revert it when I get time.

PR #288, opened 2017-12-07
More robust contract extraction.

### LINKED ISSUE #275: Assertion failure in contract extraction
This code raises an assertion failure:
````c
#include <smack-contracts.h>

int foo(void);

void bar(void) {
  int x = foo();

  while (x)
    invariant(x);
}
````
Here’s the command line:
````
$ smack -t a.c
SMACK program verifier version 1.9.0
Assertion failed: (clones.count(O.get()) && "Unvisited block."), function projection, file /Users/mje/Code/smack/lib/smack/ExtractContracts.cpp, line 105.
````

michael-emmi:
This PR improves the extraction of code contracts by leveraging LLVM mechanisms for code extraction and cloning. This effectively fixes #275. As a side effect, contract quantifiers are currently disabled, to be implemented in future PRs on demand.


zvonimir:
This seems like an improvement over our previous implementation. I have a minor question for you (more for my own benefit) - why are you using `size()` instead of `!empty()` to figure out if a container is not empty?
I would typically use `empty` in such cases. I read a bit about this, and it seems that invoking `empty` is slightly preferable since the standard guarantees it is always constant time. Anyhow, just curious if you have a good reason for doing this.
Finally, why is supporting quantifiers hard(er) in this implementation?

michael-emmi:
Hey @zvonimir I did not bother to consider the difference between `size()` and `!empty()`. Performance did not seem to be an issue, but I am happy to make that change if you wish :-)

Supporting quantifiers may not be super hard, but it is an extra effort due to introducing new bindings for quantified variables. I do not have time to deal with this at the moment, and my key client of invariants does not need quantifiers :-)

Issue #172, opened 2016-04-28
Vector operations


michael-emmi:
Here’s a fun piece of code that we cannot handle at all.
````
#include <immintrin.h>
#include <smack.h>

int main() {
  __m128i A = _mm_set_epi32(13,12,11,10);
  __m128i B = _mm_set_epi32(23,22,21,20);

  A = _mm_shuffle_epi32(A,2*1 + 3*4 + 2*16 + 3*64);
  B = _mm_shuffle_epi32(B,2*1 + 3*4 + 2*16 + 3*64);

  __m128i C = _mm_blend_epi16(A,B,0xf0);

  assert(C[0] != 0);
  return 0;
}
````
I am able to compile this as follows:
````
clang -S -emit-llvm a.c -I/usr/local/share/smack/include -mpclmul -mmmx -msse -msse2 -msse3 -msse4.1 -msse4.2 -maes
llvm2bpl a.ll -ll b.ll -entry-points=main
````
Note that besides the fact that we don’t handle vector instructions well — especially `shufflevector` — it seems that we also do not handle regular (e.g., arithmetic) operations on vector types correctly.

cmr:
I'm very interested in working on this, for using ct-verif with some vectorized Rust cipher implementations.

michael-emmi:
I can imagine this being implemented in a few ways, with varying degrees of possibly-premature optimization. The first would be to encode LLVM vectors with Boogie maps. For instance,
````
// arbitrarily-sized vectors of i16 elements
type v.i16 = [int] i16;

// some vector operations
function $add.v4.i16(x: v.i16, y: v.i16) returns (v.i16);
function $insertelement.v4.i16(x: v.i16, val: i16, idx: int) returns (v.i16);
function $extractelement.v4.i16(x: v.i16, idx: int) returns (i16);
function $shufflevector.v4.i16.m6(x: v.i16, y: v.i16, …);

// some vector-type variables and operations
var x, y, z: v.16;
y := $insertelement.v4.i16(x, 13, 1);
z := $add.v4.i16(x, y);
assert $extractelement.v4.i16(z, 1) == 20;
````
where, e.g., the `.m6` suffix fixes that shuffle operation to masks of size 6.

The second would be to avoid using maps; for instance,
````
// some vector operations
procedure $add.v4.i16(x.1: i16, x.2: i16, …, y.4: i16) returns (z.1: i16, …, z.4: i16);
[...]

PR #290, opened 2017-12-08
Initial support for SIMD vector operations.

michael-emmi:
This PR adds initial support for handling SIMD vector operations, as brought up in #172. Since many vector operations tend to introduce all kinds of bit casts, many programs cannot be proved correct without bit precision. This PR does not attempt to deal with bit precision. Instead, this PR focuses on providing the framework for handling all SIMD instructions, possibly yielding uninterpreted Boogie functions for those (bit cast) instructions that cannot be modeled precisely with Boogie integers.


zvonimir:
This looks good to me. Really nice addition. You will have to fix some conflicts though. Then I think we can merge it.

zvonimir:
I just remembered one more thing. Dan Liew told me recently that there is an LLVM pass that gets rid of (most) vector instructions. Have you looked into that at all? That would maybe be a good pass to run first.
I cannot find it at this point, but I could ask Dan about it.

michael-emmi:
I assume he was referring to [scalarization](http://llvm.org/doxygen/Scalarizer_8cpp_source.html)?

While this does enable scalar optimizations like constant propagation in some cases, and can convert many vector operations to scalar operations, it doesn’t really attempt to get rid of vector types, e.g., as storage. For instance, consider the following function:
````c
__m128i foo(__m128i a, __m128i b) {
  return _mm_add_epi64(a,b);
}
````
Without scalarization this compiles to:
````ll
define <2 x i64> @foo(<2 x i64>, <2 x i64>) #0 {
  %3 = add <2 x i64> %0, %1
  ret <2 x i64> %3
}
````

With scalarization this compiles to:
````ll
define <2 x i64> @foo(<2 x i64>, <2 x i64>) #0 {
  %.i0 = extractelement <2 x i64> %0, i32 0
  %.i01 = extractelement <2 x i64> %1, i32 0
  %.i02 = add i64 %.i0, %.i01
  %.i1 = extractelement <2 x i64> %0, i32 1
  %.i13 = extractelement <2 x i64> %1, i32 1
  %.i14 = add i64 %.i1, %.i13
  %.upto0 = insertelement <2 x i64> undef, i64 %.i02, i32 0
  %3 = insertelement <2 x i64> %.upto0, i64 %.i14, i32 1
  ret <2 x i64> %3
}
````

Note that there are still vector types involved. And even if there were a pass to expand vector storage into scalar storage, this would still not handle the case of LLVM compiler intrinsics which have no built in mapping to scalar operations. For instance, consider the following function:

````c
__m128i foo(__m128i a, __m128i b) {
  return _mm_clmulepi64_si128(a,b,0);
}
````

Both with and without scalarization, this compiles to:
````ll
define <2 x i64> @foo(<2 x i64>, <2 x i64>) #0 {
  %3 = call <2 x i64> @llvm.x86.pclmulqdq(<2 x i64> %0, <2 x i64> %1, i8 0)
  ret <2 x i64> %3
}
````

Furthermore, getting rid of vectors may be counterproductive to the real purpose of my extension, which intends to model low-level code, as close to whats executed as possible.

zvonimir:
Got it. Good, good, please merge it.

PR #293, opened 2017-12-15
Caching of build dependencies in Travis

zvonimir:
I implemented caching of folders during CI that contain our dependencies (Z3, Boogie, Corral, etc.). This should make our CI builds faster and more predictable.


zvonimir:
I will squash all the commits before merging.
It seems that caching of apt packages is unfortunately still not possible.
Finally, this will require that we manually erase old cache whenever we updated a dependency. It is likely possible to make this automatic as well (check versions, erase and reinstall dependencies if needed), but that would complicate the build script, and given that we rarely update these dependencies, I did not bother to do it.

michael-emmi:
This is cool.

Your comment makes me wonder: why not condition the install of each dependency on a function like `! upToDate(path)` instead of the `! -d path` condition you have now? Then extending the condition to take version upgrades into account could be encapsulated by `upToDate`. I have a feeling this extension may prevent some future headaches :-)

This also makes me wonder: which builds share the cache, and when does the cache become invalidated naturally? For instance, if we build two commits on distinct branches with distinct versions of some dependency which may not be backward/forward compatible, will one of them necessarily break? e.g., will builds from distinct branches share the same cache entry?

michael-emmi:
Just read the docs: says each branch gets its own cache, so the problem I worried about should not happen. Would still be nice to implement an `upToDate` function though :-)

zvonimir:
The problem with adding `upToDate` is that this check is not uniform across all dependencies. Meaning, it is currently since I am only checking whether the folder is there or not, but if I was to check e.g. versions as well, then the implementation of that would not be the same for e.g. Z3 and Boogie. So I am not sure if it is worth doing that.

michael-emmi:
But we have the freedom to make the `upToDate` check very uniform, since we control the installation of our dependencies. For instance, why not just write a `VERSION` file into the directory of each dependency containing the version number according to `bin/versions`.

zvonimir:
Ok, I think now we are good. Please take another look, and then we can hopefully merge it. Thx!

bin/build.sh
```diff
@@ -335,90 +348,97 @@ then
...
 
 
-if [ ${BUILD_BOOGIE} -eq 1 ]
-then
-  puts "Building Boogie"
-
-  if [ ! -d "$BOOGIE_DIR" ] ; then
-    git clone https://github.com/boogie-org/boogie.git ${BOOGIE_DIR}
+if [ ${BUILD_BOOGIE} -eq 1 ] ; then
+  if [ "$(upToDate $BOOGIE_DIR $BOOGIE_COMMIT)" == "false" ] ; then
```
michael-emmi: Just nitpicking here, but why not use the standard exit-code condition here? e.g., where `upToDate` returns 0 on success, and the condition is
````
upToDate $BOOGIE_DIR $BOOGIE_COMMIT -eq 0
````

michael-emmi:
Looks very cool!

bin/build.sh
zvonimir: I think my last commit fixed this. Good point.

PR #298, opened 2018-01-29
Attempt to remove the struct-ret pass

shaobo-he:
This PR is an effort to consolidate the struct-ret pass and the split-aggregate-value pass. I just realized that if we always enable splitting aggregate value load/store, then we don't need the struct-ret pass any more. Furthermore, the struct-ret pass has been creating a number of trouble when SMACK is applied to C++/Rust.

Specifically this PR does,
1. refactor aggregate value splitting
2. move constant struct return handling to the aggregate value
splitting pass
3. always enable the aggregate value splitting pass


lib/smack/SplitAggregateValue.cpp
```diff
@@ -0,0 +1,139 @@
...
+      ArrayRef<unsigned>(getSeconds(idxs)));
+  }
+  ri->setOperand(0, V);
+}
+
+void SplitAggregateValue::visitAggregateValue(Constant* baseVal, Type* T, IndexT idxs,
+                                                    std::vector<InfoT>& info, LLVMContext& C){
+  Constant* newBaseVal = baseVal;
+  if(T->isIntegerTy() || T->isFloatingPointTy() || T->isPointerTy())
+      info.push_back(std::make_pair(idxs, newBaseVal));
```
michael-emmi: While you’re here, why not improve this code using uniform initialization, e.g., replacing every instance of `std::make_pair(A,B)` with `{A, B}`.

shaobo-he:
@michael-emmi Thank you for the suggestions. If it looks good to you, I will rebase this branch and we can merge it.

PR #300, opened 2018-02-12
Disable Gold flags Ubuntu Artful (17.10) adds by default

elefthei:
A simple fix to allow building on Ubuntu Artful (17.10).

Strip away--no-keep-files-mapped, --no-map-whole-files which Artful adds by default on LLVM-3.9. It now compiles Smack on Artful.


zvonimir:
Thanks for this fix! Please create a pull request into the develop branch (not master).
See: http://nvie.com/posts/a-successful-git-branching-model/

elefthei:
Done.

zvonimir:
Ah, I just realized that just switching to merge into develop is not good enough - since your incoming branch is master. So now three commits are being merged, two of which should not be there.
You will have to redo your commit on a develop branch on your end, and then create a pull request on the develop branch on our end. I hope this makes sense.

PR #305, opened 2018-02-25
Add some usage notes

shaobo-he:
We have seen a few issues created by users who are not familiar with SMACK's flags. This document intends to provide examples of typical use cases, helping users to enable proper flags for the programs that they are interested in.

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -114,3 +114,5 @@ verifier leverages to generate more informative error traces. Then, the generate
 file is translated into Boogie code, which is in turn passed to the chosen back-end
 verifier.
 
+For mode advanced usage scenarios, please refer to our [usage notes](usage-notes.md).
+
--- a/docs/usage-notes.md
+++ b/docs/usage-notes.md
@@ -1,0 +1,105 @@
+This document shows several usage scenarios of SMACK that require special flags.
+
+## Loops and Recusive Functions
+First of all, please keep in mind that SMACK is a *bounded* verifier, which
+means that in the presense of loops and recursive functions, the verification
+process unrolls them up to the bound `N` specified by the flag `--unroll <N>`.
+Conceptually, unrolling a loop means transforming a loop into a sequence (length
+`N`) of if-else statements, the inner most of which halts the program. Recursive
+functions are handled by inlining the function `N` times and the inner most
+recursive call halts the program. Therefore, not unrolling a loop or a recursive
+function to a sufficient bound can lead to missed bugs. In other words, when
+SMACK reports that there are no bugs in a program, it actually means that the
+program is safe within the bound specified by the user.
+
+### Example
+```C
+#include "smack.h"
+
+int main (void) {
+  long x = __VERIFIER_nondet_long();
+  long y, z = 0;
+  assume(x > 2);
+  for (y = 0; y < x; ++ y) z++;
+  assert(z != x);
+  return 0;
+}
+```
+The assertion in this program should fail. However, if SMACK is invoked with the
+default unrolling bound `1`, it reports no errors. This is because the loop has
+to be unrolled at least 4 times (i.e., after `y` gets value 3, the minimum value
+of `x`, in the 3rd iteration) for the assertion to be reachable.
+
+## Bitwise Operations and Integer Casts
+If the program to verify contains bitwise operations or integer casts, then the
+flag `--bit-precise` may be required. The reason is that SMACK uses the SMT
+theory of integers to encode machine integers by default, where bitwise
+operations are encoded using uninterpreted functions returning arbitrary values.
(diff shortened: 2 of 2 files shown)


docs/usage-notes.md
```diff
@@ -0,0 +1,52 @@
...
+#include "smack.h"
+
+int main () {
+  unsigned y = __VERIFIER_nondet_unsigned_int();
+  assume (y < 4U);
+  y >>= 2U;
+  assert (!y);
+}
+```
+This program should verify. However, if the we enable SMACK in its default mode, an error would be reported.
```
zvonimir: This program should verify. However, if we run SMACK in its default mode, it reports an assertion violation.

docs/usage-notes.md
```diff
@@ -0,0 +1,52 @@
...
+    a.c(4,16): Trace: Thread=1  (y = -669)
+    ...
+    a.c(6,5): Trace: Thread=1  (y = 17)
+    ...
+SMACK found an error.
+```
+
+Some steps in the error trace are omitted. As you can see, the concrete values of `y` in the error trace before and after the bitwise right shift operation do not follow its semantics because it is modeled as an uninterpreted function.
+
+For this case, we need the `--bit-precise` flag that, as its name indicates, enables bit-precise modeling of machine integers.
```
zvonimir: In this case

docs/usage-notes.md
```diff
@@ -0,0 +1,52 @@
...
+
+```
+$ smack a.c --bit-precise
+SMACK program verifier version 1.9.0
+SMACK found no errors with unroll bound 1.
+```
+
+If the program that you would like to verify contains such patterns, please try this flag although it would cause SMACK to take much longer to perform the verification.
```
zvonimir: If the program that you would like to verify contains such patterns, try enabling this flag. However, note that enabling bit-precise reasoning often degrades the performance of SMACK, and causes for it to take much longer to perform the verification.

docs/usage-notes.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Floating-Point Arithmetic
+Similar to machine integers, floating-point numbers and arithmetic are modeled using the theory of integers and uninterpreted functions, respectively. Therefore, if the assertions to verify depend on precise modeling of floating-point representations, the flag `--float` is needed.
+
+The floating-point support of SMACK is still under active development. It is recommended that the flag `--float` is used in combination with the flag `--bit-precise`. Moreover, reasoning about floating-point numbers is often very slow. Please let us know if you encounter any performance issues. We can share some experiences that may ease the situation.
```
zvonimir: This is improved now, and so `--bit-precise` is typically not needed. I think these two paragraphs should be merged.
I think we should say:
```
Note that occasionally `--bit-precise` has to be used `--float`. Moreover, reasoning about floating-point numbers is often very slow. Please let us know if you encounter any performance issues. We can share some experiences that may ease the situation.
```

docs/usage-notes.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Concurrency
+Reasoning about pthreads is supported by SMACK with the flag `--pthread`. Please use this flag when you would like to verify a multi-threaded program with pthreads.
```
zvonimir: Add a note about having to increase the number of context switches.

docs/usage-notes.md
```diff
@@ -0,0 +1,52 @@
+This document shows several usage scenarios of SMACK that require special flags.
```
zvonimir: I wonder if you should also add a paragraph or two explaining unrolling (with an example).

zvonimir: Oh, and you should like this file from the top-level README file.

zvonimir:
Could you limit all lines of text to 80 characters?
Then I'll proofread it and maybe we could merge it soon. Thx!

shaobo-he:
> Could you limit all lines of text to 80 characters?
> Then I'll proofread it and maybe we could merge it soon. Thx!

Done.

docs/running.md
```diff
@@ -114,3 +114,5 @@ verifier leverages to generate more informative error traces. Then, the generate
 file is translated into Boogie code, which is in turn passed to the chosen back-end
 verifier.
 
+For advanced usages, please refer to the [usage notes](usage-notes.md).
```
zvonimir: `For mode advanced usage scenarios, please refer to our usage notes.`

zvonimir:
I think it'll be the easiest if I carefully proofread this once we merge it.

zvonimir:
Please do this one small fix, and then squash the commits.

docs/running.md
shaobo-he: Done.

PR #306, opened 2018-02-28
Fixes for floating-point support

shaobo-he:
This pull request intends to fix issues in SMACK's floating-point support. It is not yet completed but I would like to create the pull request first because there are multiple design decisions worth of discussion.

The issues addressed are as follows,
- [x] Change the type of uninterpreted floats
- [x] Support half-precision
- [x] Decouple FP support from the `--bit-precise` flag
- [x] Bug fixes
- [ ] Fix SMACK's `math.h`
- [ ] Add more regressions


lib/smack/SmackRep.cpp
```diff
@@ -834,8 +836,14 @@ const Expr* SmackRep::cmp(const llvm::ConstantExpr* CE) {
...
-  std::string fn = Naming::CMPINST_TABLE.at(predicate);
-  return Expr::fn(opName(fn, {lhs->getType()}), expr(lhs, isUnsigned), expr(rhs, isUnsigned));
+  std::string fn = opName(Naming::CMPINST_TABLE.at(predicate), {lhs->getType()});
+  const Expr* e1 = expr(lhs, isUnsigned);
+  const Expr* e2 = expr(rhs, isUnsigned);
+  if (lhs->getType()->isFloatingPointTy())
+    return Expr::if_then_else(Expr::fn(fn+".bool", e1, e2), integerLit(1UL,1), integerLit(0UL,1));
+  else
+    return Expr::fn(fn, e1, e2);
+;
```
zvonimir: Seems like there is an extra `;` here.

lib/smack/SmackRep.cpp
```diff
@@ -834,8 +836,14 @@ const Expr* SmackRep::cmp(const llvm::ConstantExpr* CE) {
 }
 
 const Expr* SmackRep::cmp(unsigned predicate, const llvm::Value* lhs, const llvm::Value* rhs, bool isUnsigned) {
-  std::string fn = Naming::CMPINST_TABLE.at(predicate);
-  return Expr::fn(opName(fn, {lhs->getType()}), expr(lhs, isUnsigned), expr(rhs, isUnsigned));
+  std::string fn = opName(Naming::CMPINST_TABLE.at(predicate), {lhs->getType()});
+  const Expr* e1 = expr(lhs, isUnsigned);
+  const Expr* e2 = expr(rhs, isUnsigned);
+  if (lhs->getType()->isFloatingPointTy())
```
zvonimir: I don't understand why different expressions are generated for integers vs floating-points. Please clarify. Thx!

lib/smack/SmackRep.cpp
```diff
@@ -1068,6 +1079,14 @@ std::string SmackRep::getPrelude() {
       << "\n";
   }
 
+  //s << "// Context-aware float-integer conversions" << "\n";
```
zvonimir: Could we remove this completely instead of commenting it out?
I would prefer if we would do that.

lib/smack/SmackRep.cpp
```diff
@@ -514,14 +516,28 @@ const Stmt* SmackRep::returnValueAnnotation(const CallInst& CI) {
 //
 // }
 
+bool SmackRep::isUnsafeFloatAccess(const Type* elemTy, const Type* resultTy) {
+  bool unsafeFloat = false;
+  if (elemTy->isFloatingPointTy()) {
+    if (!resultTy || (resultTy->isIntegerTy() && resultTy->getIntegerBitWidth() == 8UL)) {
+      if (!SmackOptions::BitPrecise)
+        unsafeFloat = true;
```
zvonimir: Just return `true` here right away.

lib/smack/SmackRep.cpp
```diff
@@ -514,14 +516,28 @@ const Stmt* SmackRep::returnValueAnnotation(const CallInst& CI) {
...
+    } else
+      assert(resultTy->isFloatingPointTy() && "Unsupported map result type.");
+  }
+  return unsafeFloat;
```
zvonimir: Return `false` here and get rid of this local variable.

lib/smack/SmackRep.cpp
```diff
@@ -514,14 +516,28 @@ const Stmt* SmackRep::returnValueAnnotation(const CallInst& CI) {
 //
 // }
 
+bool SmackRep::isUnsafeFloatAccess(const Type* elemTy, const Type* resultTy) {
+  bool unsafeFloat = false;
+  if (elemTy->isFloatingPointTy()) {
+    if (!resultTy || (resultTy->isIntegerTy() && resultTy->getIntegerBitWidth() == 8UL)) {
+      if (!SmackOptions::BitPrecise)
```
zvonimir: Can't you merge this if condition with the one on line 521 above?

lib/smack/SmackRep.cpp
zvonimir: I still don't understand why different expressions are generated for ints and different for floating-point. Please comment.

lib/smack/SmackRep.cpp
zvonimir: Can't you remove completely this commented out section?

lib/smack/SmackRep.cpp
shaobo-he: Sorry, I do not get it. Can you clarify it?

lib/smack/SmackRep.cpp
zvonimir: I see, thanks for the clarification. So why are then floating-point and integer comparisons not implemented uniformly? Maybe this is just a temporary fix, which is fine, but I would guess that eventually we should have two versions of floating-point comparisons as well, which would avoid the introduction of this if-then-else instruction. Is that the plan?

lib/smack/SmackRep.cpp
zvonimir: This whole thing can be just one if condition instead of 3 nested ones:
```
if (!SmackOptions::BitPrecise && elemTy->isFloatingPointTy() &&
  (!resultTy || (resultTy->isIntegerTy() && resultTy->getIntegerBitWidth() == 8UL))) {
```

lib/smack/SmackRep.cpp
```diff
@@ -514,14 +516,27 @@ const Stmt* SmackRep::returnValueAnnotation(const CallInst& CI) {
 //
 // }
 
+bool SmackRep::isUnsafeFloatAccess(const Type* elemTy, const Type* resultTy) {
+  if (elemTy->isFloatingPointTy()) {
```
zvonimir: I think I already mentioned this, but why don't you merge these into just one if condition?

lib/smack/SmackRep.cpp
```diff
@@ -514,14 +516,27 @@ const Stmt* SmackRep::returnValueAnnotation(const CallInst& CI) {
...
+    if (!resultTy || (resultTy->isIntegerTy() && resultTy->getIntegerBitWidth() == 8UL)) {
+      if (!SmackOptions::BitPrecise)
+        return true;
+    } else
+      assert(resultTy->isFloatingPointTy() && "Unsupported map result type.");
```
zvonimir: Hmmm, it is weird to have just an assert statement in this else branch. Could this somehow be rewritten?

lib/smack/SmackRep.cpp
shaobo-he: I don't think merging the sub-conditions would work because there is an assertion involved. I've rewritten the conditions so that they are more readable.

shaobo-he:
Note that the fixes to load/store to collapsed maps are not correctly implemented. For example,

```C
#include <assert.h>

int main()
{
  int i = 2147483647;
  float f = *((float*) &i);
  int i1 = *((int*) &f);
  assert (i == i1);
  return 0;
}
```

This assertion should hold. I found it difficult to model this program correctly.

zvonimir:
Why is it hard to model that program?

share/smack/lib/math.c
```diff
@@ -125,49 +126,50 @@ float copysignf(float x, float y) {
...
-  __SMACK_code("@ := if $issubnormal.bvfloat(dtf(@)) then 1bv32 else 0bv32;", ret, x);
+  __SMACK_code("@ := if $issubnormal.bvfloat.bool(@) then $1 else $0;", ret, x);
   return ret;
 }
+*/
 
 int __iszerof(float x) {
   int ret = __VERIFIER_nondet_int();
-  __SMACK_code("@ := if $iszero.bvfloat(dtf(@)) then 1bv32 else 0bv32;", ret, x);
+  __SMACK_code("@ := if $iszero.bvfloat.bool(dtf(@)) then $1 else $0;", ret, x);
```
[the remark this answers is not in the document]
michael-emmi: I don’t understand what you mean by:

> types of variadic arguments to variadic functions would be prompted. For this case, x would be lifted to a double.

Also, what about the lack of `dtf` at previous uses, on Lines 40, 129, and 135. Are these lines incorrect as they are? Why has the call to `dtf` been removed there?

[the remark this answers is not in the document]
shaobo-he: Sorry for the confusion. There shouldn't be any inconsistency in using `dtf`. Line 40, 129, 135 are not correct. Calls to `dtf` in these locations are mistakenly removed as a previous misunderstanding of why `dtf` is used here.

We should add `dtf` calls to make Boogie type check because arguments of variable argument function may be changed to more precise types. https://stackoverflow.com/questions/11270588/variadic-function-va-arg-doesnt-work-with-float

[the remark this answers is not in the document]
michael-emmi: So we are going to fix this before approval right?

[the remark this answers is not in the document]
zvonimir: Aren't these lines commented out?
Basically, @shaobo-he noticed other problems with a bunch of functions in `math.c`, and he commented those out in this pull request. @liammachado has been working on a proper fix for all the commented out functions.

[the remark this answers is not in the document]
michael-emmi: Oh, that is true; the PR’s code diff did not make that clear. I retract this point then.

PR #307, opened 2018-03-10
Simplify smack.c using macro tricks

shaobo-he:
This pull request attempts to simplify the declarations in smack.c. Specifically, the current implementation of `smack.c` declares integer conversions per bitwidth pair (e.g., `DECLARE(INLINE_CONVERSION,i128,i96,$trunc,{i});`). This PR tries to compress these declarations into one macro that produces bitwidth pairs.

I think we might not need such simplifications and the current implementation is still a prototype.


share/smack/include/smack-macros.h
```diff
@@ -0,0 +1,142 @@
+// shamelessly stole these macro definitions from https://github.com/pfultz2/Cloak/wiki/C-Preprocessor-tricks,-tips,-and-idioms
```
zvonimir: Please add LICENSE info to this new file. Also, break this first long comment into two lines.

shaobo-he:
@michael-emmi thank you for your feedback. Actually @keram88 has also mentioned that smack.c is better generated via a script, which I totally agree. We should implement it.

I'd like to run this branch on SV-COMP benchmarks before we merge it.

keram88:
I am interested in writing the a python generator for smack.c's 'boiler-plate' code.

michael-emmi:
@keram88 this could be a script in `share/smack` invoked by `top.py` that generates a unique `smack-XYZ.c` in the working directory, which `top.py` then sources to generate `smack-XYZ.bc` in place of the usual lib, e.g., `/usr/local/share/smack/lib/smack.c`.

zvonimir:
@shaobo-he : Please create an issue related to the mentioned python scripts and move this discussion there. Also, can we merge this? How do SVCOMP results look like?

shaobo-he:
@zvonimir SV-COMP results look good. I think it's good to be merged. Sorry for the delay. The d820 resources were pretty limited last week.

Issue #310, opened 2018-03-30
Unsupported Constant Type


jjgarzella:
SMACK currently supports a limited subset of the types of LLVM IR constants. The following type of constant is used by the D programming language, but SMACK crashes whenever such a constant is used by code. 

Examples:
```
@.str = private unnamed_addr constant [13 x i8] c"Hello World!\00" ; [#uses = 1]
@.str.1 = private unnamed_addr constant [10 x i8] c"null this\00" ; [#uses = 1]
```

[examples.zip](https://github.com/smackers/smack/files/1864465/examples.zip)

To run the examples, use `smack [filename].ll --entry-points _Dmain`

shaobo-he:
@jjgarzella This issue actually has nothing to do with the constant arrays representing strings. The reason is that constant structures are used as arguments.

```
define i32 @_Dmain({ i64, { i64, i8* }* } %unnamed) #0 comdat {
  call void @_D3std5stdio16__T7writelnTAyaZ7writelnFNfAyaZv({ i64, i8* } { i64 12, i8* getelementptr inbounds ([13 x i8], [13 x i8]* @.str, i32 0, i32 0) })
  ret i32 0
}
```
A structure of `{ i64, i8* }` is passed as the argument.

PR #311, opened 2018-03-30
Rewrite constant aggregate arguments to constructed values

shaobo-he:
Fixes issue #310

### DIFF
--- a/include/smack/SplitAggregateValue.h
+++ b/include/smack/SplitAggregateValue.h
@@ -25,6 +25,8 @@ private:
   void splitAggregateLoad(llvm::LoadInst* li, std::vector<InfoT>& info);
   void splitAggregateStore(llvm::StoreInst* si, std::vector<InfoT>& info);
   void splitConstantReturn(llvm::ReturnInst* ri, std::vector<InfoT>& info);
+  void splitConstantArg(llvm::CallInst* ci, unsigned i, std::vector<InfoT>& info);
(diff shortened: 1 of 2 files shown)


michael-emmi:
Why does the `createInsertedValue` function create both `insertvalue` and `store` instructions?

shaobo-he:
@michael-emmi I think the reason is that the succeeding optimization passes used by SMACK try to re-build the aggregate constant. So I had to store each element into an allocated space and use the value loaded back to build the returned aggregate value.

michael-emmi:
@shaobo-he can you be more specific? which pass(es) require this?

Putting aside for the moment the difference between a constant value versus an allocated value, it seems that there is a bit of redundancy between what is done in `createInsertedValue` and `splitAggregate(Load|Store)`.

Thus another way to ask my question is: would it be cleaner to refactor this so that `createInsertedValue` just deals with turning the constant aggregate value into an allocated aggregate value, and then the `splitAggregate(Load|Store)` functions do the actual splitting?

From a maintainability perspective this might be better, since changing how aggregates are split is independent from the allocation status of constants.

shaobo-he:
Originally I thought that I could just turn an aggregate constant into an aggregate value using `insertvalue` instructions. Certain unknown pass(es) would this series of `insertvalue` into the input aggregate constant. I'm not sure which pass does this.

Do you mean we store the aggregate constant into an allocated space and load the return value from the space, where both the store and load instructions are split via calls to `splitAggregate(Load|Store)` functions?

michael-emmi:
This looks much nicer now!

PR #316, opened 2018-05-09
D support

jjgarzella:
This adds top.py support for D. 

Unfortunately, we can't test this with D files until we upgrade to Ubuntu 16.04. Right now, ldc doesn't have a version for Ubuntu 14.04 that can be installed with a package manager. When we have 16.04, we can test this. However, I still think it's good to put in right now while it's in the foreground of our attention.

For the same reason, build.sh hasn't been modified to have an INSTALL_D option. In fact, it's obscure enough and simple enough that I almost don't think we should pollute build.sh with D.

### DIFF
--- a/share/smack/include/smack.di
+++ b/share/smack/include/smack.di
@@ -1,0 +1,4 @@
+
+extern (C): void __VERIFIER_assert(int);
+
+//extern (C): void __smack_assert(int);
--- a/share/smack/top.py
+++ b/share/smack/top.py
@@ -364,7 +364,6 @@ def objc_clang_frontend(args):
     sys.exit("Objective-C not yet supported on macOS")
   else:
     sys.exit("Objective-C not supported for this operating system.")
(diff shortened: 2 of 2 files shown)


share/smack/include/smack.di
```diff
@@ -0,0 +1,2 @@
+
```
zvonimir: Please add license info - you can simply copy-paste it from almost any other file.

share/smack/include/smack.di
```diff
@@ -0,0 +1,2 @@
...
+extern (C): void __VERIFIER_assert(int);
```
zvonimir: What about `__VERIFIER_nondet_int` and such?

jjgarzella: The only method required for the benchmarks I've been using so far is `__VERIFIER_assert`. I can add those method signatures here, but they will be untested and may or may not work.

PR #317, opened 2018-05-28
Fortran support

jjgarzella:
Add top.py support for using smack on Fortran files 

SMACK exposes two functions to fortran: a raw __verifier_assert() which takes an integer, and assert() which takes a LOGICAL type.

### DIFF
--- a/share/smack/lib/smack.f90
+++ b/share/smack/lib/smack.f90
@@ -1,0 +1,27 @@
+
+module smack
+  implicit none
+      interface
+        subroutine __verifier_assert(cond) bind(c, name="__VERIFIER_assert")
+          use, intrinsic :: iso_c_binding, only: c_int
+          implicit none
+          integer(c_int), value :: cond
+        end subroutine __verifier_assert
+      end interface
+  contains
+    subroutine assert(cond) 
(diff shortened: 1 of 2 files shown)


share/smack/lib/smack.f90
```diff
@@ -0,0 +1,27 @@
+
```
zvonimir: Add license info (copy-paste from some other file).

share/smack/lib/smack.f90
```diff
@@ -0,0 +1,27 @@
...
+module smack
```
zvonimir: What about `__SMACK_nondet_int` and such?

share/smack/top.py
```diff
@@ -367,6 +369,39 @@ def objc_clang_frontend(args):
   compile_command = objc_clang_compile_command(args)
   default_link_bc_files(compile_command, args)
 
+def fortran_frontend(args):
+  """Generate Boogie code from Fortran language source(s)."""
```
zvonimir: Please add more comments so that we know a year down the road what basically almost each line is doing. Especially comments hacks.

PR #320, opened 2018-06-02
Fix phi nodes that get their values from a loop body

### LINKED ISSUE #314: Problematic translation of phi instruction
Hey guys,

Ian found an issue with SMACK's translation of phi instruction. Basically the current implementation adds assignment at the end of each predecessor block. This can be problematic if the incoming value is a value from the block where phi instruction resides. For example, consider the following CFG,

```
ENTRY: ...

BODY:  i = phi [0, ENTRY], [incr, BODY]
       incr = add i 1
       br *,  BODY, CONT

CONT: br EXIT

EXIT: f = phi [i, CONT], [*,*]
```
, the Boogie translation would be,

```
ENTRY: ...

BODY: incr := i + 1;
      i := incr;

CONT: f := i;

EXIT: ...
```
If we have a path ENTRY --> BODY --> CONT --> EXIT, we should expect `f` to be equal to `0` while in the Boogie program it's `1`.

keram88:
This adds an LLVM pass to fix #314.

Specifically, if a phi node in a loop header gets its value from a block in the loop body, and the source block branches when terminating, the generated Boogie code incorrectly updates the variable corresponding to the phi node.  To fix this, a forwarding block is added between the source block and the header. The generated Boogie code then updates the value for the phi node in the forwarding block, and other destination blocks keep the non-updated value for the phi node. Loops of this form are usually generated when compiling with optimization.

### DIFF
--- a/lib/smack/NormalizeLoops.cpp
+++ b/lib/smack/NormalizeLoops.cpp
@@ -1,0 +1,154 @@
+//
+// This file is distributed under the MIT License. See LICENSE for details.
+//
+
+//
+// This pass normalizes loop structures to allow easier generation of Boogie
+// code when a loop edge comes from a branch. Specifically, a forwarding block
+// is added between a branching block and the loop header.
+//
+
+#include "smack/NormalizeLoops.h"
+#include "smack/Naming.h"
+#include "llvm/IR/Module.h"
+#include "llvm/IR/InstIterator.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/ValueSymbolTable.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/Analysis/LoopInfo.h"
+#include "llvm/IR/Dominators.h"
+#include "llvm/Support/raw_ostream.h"
+#include <map>
+#include <vector>
+#include <set>
+
+namespace smack {
+
+using namespace llvm;
+
+// Register LoopInfo
+void NormalizeLoops::getAnalysisUsage(AnalysisUsage& AU) const {
+  AU.addRequired<LoopInfoWrapperPass>();
+}
+
+/**
+ * Creates a basic block which unconditionally branches to \param target.
+ */
+BasicBlock* makeForwardingBlock(BasicBlock* target) {
+  auto& context = target->getContext();
+  BasicBlock* result = BasicBlock::Create(context, "forwarder", target->getParent());
+  BranchInst* branch = BranchInst::Create(target);
+  result->getInstList().push_back(branch);
+  return result;
+}
+
+/**
+ * Returns a vector of each successor index in \param ti which equals \param
+ * headerBlock.
+ */
+std::vector<unsigned> getSuccessorIndices(TerminatorInst* ti, BasicBlock* headerBlock) {
+  std::vector<unsigned> result;
+  unsigned numSuccs = ti->getNumSuccessors();
+  
+  if (numSuccs <= 1) { return result; }
+  
+  for (unsigned i = 0; i < numSuccs; ++i) {
+    if (ti->getSuccessor(i) == headerBlock) {
+      result.push_back(i);
+    }
+  }
+  return result;
+}
+
+/**
+ * Splits any edge between \param BB and \param headerBlock in a loop body,
+ * by adding a forwarding block between \param BB and \param headerBlock. The
+ * \param blockMap is updated to map between \param BB and the newly created
+ * forwarding block.
+ */
+void splitBlock(BasicBlock* BB, BasicBlock* headerBlock,
+                std::map<BasicBlock*, BasicBlock*>& blockMap) {
+  TerminatorInst* ti = BB->getTerminator();
+
+  // Get a list of all successor indices which point to headerBlock.
+  std::vector<unsigned> succIndices = getSuccessorIndices(ti, headerBlock);
+
+  // TerminatorInsts may have the header block listed multiple times as a
+  // successor block, e.g., in a SwitchInst. We need to make sure each such
+  // index uses the same forwarding block.
+  if (succIndices.size()) {
+    BasicBlock* forwarder = makeForwardingBlock(headerBlock);
+    blockMap[BB] = forwarder;
+    for (auto succIndex : succIndices) {
+      ti->setSuccessor(succIndex, forwarder);
+    }
+  }
+}
+
+/**
+ * \param blockMap a map from old BasicBlocks and new BasicBlocks.
+ * Fixes any PHI nodes in \param phis to get their incoming block from the new
+ * blocks in \param blockMap.
+ */
+void processPhis(std::vector<PHINode*>& phis, std::map<BasicBlock*, BasicBlock*>& blockMap) {
+  for (PHINode* phi : phis) {
+    unsigned numIncoming = phi->getNumIncomingValues();
+    for (unsigned i = 0; i < numIncoming; ++i) {
+      BasicBlock* incomingBlock = phi->getIncomingBlock(i);
+      if (blockMap.count(incomingBlock)) {
+        phi->setIncomingBlock(i, blockMap[incomingBlock]);
+      }
+    }
+  }
+}
+
+void processLoop(Loop *loop) {
+  std::vector<PHINode*> phis;
+  std::set<BasicBlock*> phiIncBlocks;
+  std::set<BasicBlock*> loopBlocks(loop->block_begin(), loop->block_end());
+  BasicBlock* headerBlock = loop->getHeader();
+  errs() << headerBlock->getParent()->getName() << ": " << *loop << "\n";
+  // Gather Phis in the header
+  for (Instruction& I : *headerBlock) {
+    if (auto* phi = dyn_cast<PHINode>(&I)) {
+      phis.push_back(phi);
+      phiIncBlocks.insert(phi->block_begin(), phi->block_end());
+    }
+  }
+
+  std::map<BasicBlock*, BasicBlock*> blockMap;
+
+  // Add forwarding blocks
+  for (BasicBlock* BB : phiIncBlocks) {
+    if (loopBlocks.count(BB) == 0) { continue; }
+    splitBlock(BB, headerBlock, blockMap);
+  }
+
+  // Fix up Phi nodes
+  processPhis(phis, blockMap);
+
+  // Handle subloops
+  for (Loop *subloop : loop->getSubLoops()) {
+    processLoop(subloop);
+  }
+}
+
+bool NormalizeLoops::runOnModule(Module& m) {
+  for (auto F = m.begin(), FEnd = m.end(); F != FEnd; ++F) {
+    if (F->isIntrinsic() || F->empty()) { continue; }
+    LoopInfo& loopInfo = getAnalysis<LoopInfoWrapperPass>(*F).getLoopInfo();
+    for (LoopInfo::iterator LI = loopInfo.begin(), LIEnd = loopInfo.end(); LI != LIEnd; ++LI) {
+      processLoop(*LI);
+    }
+  }
+
+  return true;
+}
+
+// Pass ID variable
+char NormalizeLoops::ID = 0;
+
+const char* NormalizeLoops::getPassName() const {
+  return "NormalizeLoops";
+}
+}
--- a/lib/smack/NormalizeLoops.cpp
+++ b/lib/smack/NormalizeLoops.cpp
@@ -17,7 +17,6 @@
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/Analysis/LoopInfo.h"
 #include "llvm/IR/Dominators.h"
-#include "llvm/Support/raw_ostream.h"
 #include <map>
 #include <vector>
 #include <set>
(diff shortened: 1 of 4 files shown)


lib/smack/NormalizeLoops.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+BasicBlock* mkForwardingBlock(BasicBlock* target, Function* func) {
+  auto& context = target->getContext();
+  BasicBlock* result = BasicBlock::Create(context, "forwarder", func);
+  BranchInst* branch = BranchInst::Create(target);
+  result->getInstList().push_back(branch);
+  return result;
+}
+
+void processPhis(std::vector<PHINode*>& phis, std::map<BasicBlock*, BasicBlock*>& blockMap) {
```
zvonimir: Write Doxygen-style comment about what this function does.

lib/smack/NormalizeLoops.cpp
```diff
@@ -0,0 +1,116 @@
...
+namespace smack {
+
+using namespace llvm;
+
+// Register LoopInfo
+void NormalizeLoops::getAnalysisUsage(AnalysisUsage& AU) const {
+  AU.addRequired<LoopInfoWrapperPass>();
+}
+
+BasicBlock* mkForwardingBlock(BasicBlock* target, Function* func) {
```
zvonimir: Just call it `makeForwardingBlock`

lib/smack/NormalizeLoops.cpp
```diff
@@ -0,0 +1,116 @@
...
+  result->getInstList().push_back(branch);
+  return result;
+}
+
+void processPhis(std::vector<PHINode*>& phis, std::map<BasicBlock*, BasicBlock*>& blockMap) {
+  for (PHINode* phi : phis) {
+    unsigned numIncoming = phi->getNumIncomingValues();
+    for (unsigned i = 0; i < numIncoming; ++i) {
+      BasicBlock* incomingBlock = phi->getIncomingBlock(i);
+      if (blockMap.count(incomingBlock) == 1) {
```
zvonimir: Can this ever be > 1?
If not, maybe adding an assertion `assert blockMap.count <= 1` before this line would be a good idea. Just in case.

lib/smack/NormalizeLoops.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+void processLoop(Loop *loop, Function* func) {
+  std::vector<PHINode*> phis;
+  std::set<BasicBlock*> phiIncBlocks;
+  BasicBlock* headerBlock = loop->getHeader();
+
+  // Gather Phis in the header
+  for (Instruction& I : *headerBlock) {
+    PHINode* phi = dyn_cast<PHINode>(&I);
```
zvonimir: Move this line into the if condition, like this: http://llvm.org/docs/ProgrammersManual.html#the-isa-cast-and-dyn-cast-templates

lib/smack/NormalizeLoops.cpp
```diff
@@ -0,0 +1,116 @@
...
+    if (phi) {
+      phis.push_back(phi);
+      phiIncBlocks.insert(phi->block_begin(), phi->block_end());
+    }
+  }
+
+  std::map<BasicBlock*, BasicBlock*> blockMap;
+
+  // Add forwarding blocks
+  for (BasicBlock* BB : phiIncBlocks) {
```
zvonimir: Pull this into a separate function since this one is pretty big.

lib/smack/NormalizeLoops.cpp
keram88: Done.

lib/smack/NormalizeLoops.cpp
keram88: Done.

lib/smack/NormalizeLoops.cpp
keram88: This is a simple map, so it can only return 0 or 1. The specific value is no longer checked since multi-maps behave differently.

lib/smack/NormalizeLoops.cpp
keram88: Done.

lib/smack/NormalizeLoops.cpp
keram88: Done.

lib/smack/NormalizeLoops.cpp
```diff
@@ -0,0 +1,134 @@
...
+  return result;
+}
+
+/**
+ * Splits any edge between \param BB and \param headerBlock in a loop body,
+ * by adding a forwarding block between \param BB and \param headerBlock. The
+ * \param blockMap is updated to map between \param BB and the newly created
+ * forwarding block.
+ */
+void splitBlock(BasicBlock* BB, BasicBlock* headerBlock, Function* func, std::map<BasicBlock*,BasicBlock*>& blockMap) {
```
zvonimir: Please split this argument list into two lines.

zvonimir:
I left one more simple comment. In addition to that, you should squish everything into one commit. You already wrote a nice message for this pull request that you can use as your commit message. Except that you should move the "fixes" part to the end of the message (and not have it in the beginning).
Then I think we should be almost ready to merge. I am running SVCOMP now to create baseline, and we should be able to run this branch tomorrow I hope to do a comparison.

lib/smack/NormalizeLoops.cpp
keram88: Done.

lib/smack/NormalizeLoops.cpp
```diff
@@ -0,0 +1,134 @@
...
+  unsigned numSuccs = ti->getNumSuccessors();
+  if (numSuccs <= 1) { return; }
+  BasicBlock* forwarder = nullptr;
+  for (unsigned i = 0; i < numSuccs; ++i) {
+    BasicBlock* succ = ti->getSuccessor(i);
+    if (succ != headerBlock) { continue; }
+
+    // Successor is the header block and this block has multiple successors,
+    // so it needs a forwarding block.
+    if (forwarder == nullptr) {
```
michael-emmi: Declaring `forwarder` outside of the loop, and checking this condition here seems fairly convoluted, given that this control location is only reachable for one single block: `headerBlock`. This condition should thus be guaranteed to be true, and there should be no reason to initialize `forwarder` to `nullptr`. In fact, the only reason for this loop is to determine the successor index of `headerBlock`, if any. This code would be much clearer with a call to some `getSuccessorIndex(ti, headerBlock)` function, which returns the successor index of `headerBlock`, or `-1` if its not a successor of `ti`.

lib/smack/NormalizeLoops.cpp
```diff
@@ -0,0 +1,134 @@
...
+    for (unsigned i = 0; i < numIncoming; ++i) {
+      BasicBlock* incomingBlock = phi->getIncomingBlock(i);
+      if (blockMap.count(incomingBlock)) {
+        phi->setIncomingBlock(i, blockMap[incomingBlock]);
+      }
+    }
+  }
+}
+
+void processLoop(Loop *loop, Function* func) {
```
michael-emmi: There’s no need for `func` to be a parameter here or in `splitBlock` or `makeForwardingBlock`, since it can be determined when it is needed in `makeForwardingBlock` as `target->getParent()`.

lib/smack/NormalizeLoops.cpp
```diff
@@ -0,0 +1,134 @@
...
+  for (BasicBlock* BB : phiIncBlocks) {
+    splitBlock(BB, headerBlock, func, blockMap);
+  }
+
+  // Fix up Phi nodes
+  processPhis(phis, blockMap);
+
+  // Handle subloops
+  for (Loop *subloop : loop->getSubLoops()) {
+    processLoop(subloop, func);
```
michael-emmi: Just curious: have we tested nested loops, or even thought about the possibility of what happens when the processing the nested loop splits blocks? For instance, whether this code ends up processing the original (unspilt) blocks of the enclosing loop, or the updated (split) blocks?

lib/smack/NormalizeLoops.cpp
keram88: This is to deal with switch instructions, which may point to the header more than once.

lib/smack/NormalizeLoops.cpp
keram88: I've tested it with nested loops, both handwritten and compiled, and loop headers appear to be unique to the loop. (Su- loops have different headers.) However, the documentation doesn't say that this behavior is guaranteed.

lib/smack/NormalizeLoops.cpp
keram88: Done.

lib/smack/NormalizeLoops.cpp
michael-emmi: Oh ok, fair point. But this subtlety that you point out can not be inferred by reading this code, unless you already have switch instructions in mind. It would thus help future maintenance to make this clearer. I would suggest a variation on my previous suggestion: a call to some `getSuccessorIndices(ti, headerBlock)` function, which returns the successor indices of `headerBlock`. You can then return early if this list is empty, replacing the check of `numSuccs <= 1`, and otherwise create the forwarding block.

keram88: This is done. However, it still may be desirable to check the number of successors to prevent a forwarding block from being created between a body block that branches unconditionally to the loop header.

keram88:
@michael-emmi what you mention is possible. If loops do have unique headers, then any processing of subloops would not affect the work done in the parent function, since only edges to the loop header are changed.

This has a subloop (bb2) which branches to the outer loop header (bb0):
``` llvm
  entry:
    %0 = add nsw i32 0, 0
    br label %bb0
  bb0:
    %1 = phi i32 [ %0, %entry ], [ %b, %bb2 ], [%2, %bb0]
    %2 = add i32 %1, 1
    %cmp0 = icmp eq i32 %2, 5
    br i1 %cmp0, label %bb0, label %bb1
  bb1:
    %cmpexit = icmp eq i32 %2, 5
    br i1 %cmpexit, label %exit, label %bb2
  bb2:
    %a = phi i32 [ %b, %bb2 ], [ %2, %bb1 ]
    %b = add nsw i32 %a, 1
    %cmp1 = icmp eq i32 %a, 4
    br i1 %cmp1, label %bb0, label %bb2
  exit:
    call void @__VERIFIER_assert(i32 %0), !dbg !20, !verifier.code !19
    ret i32 0, !dbg !21, !verifier.code !17
```

It is transformed into:
``` llvm
entry:
  %0 = add nsw i32 0, 0, !verifier.code !16
  br label %bb0, !verifier.code !16

bb0:                                              ; preds = %forwarder1, %forwarder, %entry
  %1 = phi i32 [ %0, %entry ], [ %b, %forwarder1 ], [ %2, %forwarder ], !verifier.code !16
  %2 = add i32 %1, 1, !verifier.code !16
  %cmp0 = icmp eq i32 %2, 5, !verifier.code !16
  br i1 %cmp0, label %forwarder, label %bb1, !verifier.code !16

bb1:                                              ; preds = %bb0
  %cmpexit = icmp eq i32 %2, 5, !verifier.code !16
  br i1 %cmpexit, label %exit, label %bb2, !verifier.code !16

bb2:                                              ; preds = %forwarder2, %bb1
  %a = phi i32 [ %b, %forwarder2 ], [ %2, %bb1 ], !verifier.code !16
  %b = add nsw i32 %a, 1, !verifier.code !16
  %cmp1 = icmp eq i32 %a, 4, !verifier.code !16
  br i1 %cmp1, label %forwarder1, label %forwarder2, !verifier.code !16

exit:                                             ; preds = %bb1
  call void @__VERIFIER_assert.3(i32 %0), !dbg !17, !verifier.code !18
  ret i32 0, !dbg !19, !verifier.code !16

forwarder:                                        ; preds = %bb0
  br label %bb0, !verifier.code !16

forwarder1:                                       ; preds = %bb2
  br label %bb0, !verifier.code !16

forwarder2:                                       ; preds = %bb2
  br label %bb2, !verifier.code !16
```

michael-emmi:
Excellent answer!

michael-emmi:
Won’t we be adding a regression? Issue #314 should probably suggest something?

keram88:
I'm not sure what to add as a regression. I could have some of our existing tests compile at O3, but compiler optimizations are unpredictable between versions. I could also add my LLVM-IR tests, they won't be future compatible, but at least we would know they are failing when LLVM gets upgraded.

zvonimir:
@michael-emmi, regarding adding regressions, I created a new issue for that feature: https://github.com/smackers/smack/issues/321
Let's discuss that separately under that issue since it requires updates to our regression infrastructure.

zvonimir:
I say we merge this (after I run SVCOMP benchmarks for comparison), and deal with regressions as soon as we merge it in a separate pull request.

michael-emmi:
But why not have a C-code regression which uses `-O3`? Yes, this may not be perfectly predictable, but isn’t it anyhow more stable than an LLVM-code regression?

zvonimir:
We encountered examples where we could not figure out how to generate a C regression at all. The difference between using O3 is LLVM IR is that when we use O3 we don't really know what in the end gets generated. Hence, we won't really know if we are still testing for the right thing. I guess after every LLVM version update, someone would have to double-check whether the O3-generated LLVM IR still looks like what we need. I hope this makes sense.

Issue #197, opened 2016-12-19
Reimplementing memory safety checking


zvonimir:
Memory safety checking should be implemented as a transformation over Boogie ASTs after all. The problem with the current implementation, which introduces calls to memory safety checking function in LLVM IR, is that it typically causes for DSA to merge everything into one region (since this same function is called everywhere with different pointers). I think that the solution would be to reimplement memory safety checking as a pass over the generate Boogie ASTs. This pass would inject calls to our memory safety checking function directly in Boogie, thereby not causing problems with DSA.

zvonimir:
@michael-emmi : I would appreciate your feedback/thoughts.

michael-emmi:
Well I think that it is always better to implement things at the LLVM level, so long as it is possible, and does not force a collapse of memory regions.

Before going 

````
````

michael-emmi:
Well I think that it is always better to implement on LLVM IR, as long as it is possible to do so, without forcing complete region collapse.

In this case, I think there may still be ways to convince DSA not to collapse regions. Consider the following example.
````
#include <smack.h>

int foo(int*) __attribute__((const));

void use(int);

int main(void) {
  
  int x = 0;
  int y = 1;
  
  use(foo(&x));
  use(foo(&y));
    
  return 0;
}
````
In this case the `const` attribute means that the `foo` function does not have any effects besides its return value; see http://nshipster.com/__attribute__/

If you run this through Smack you will see that the regions of `x` and `y` are not merged on account of calls to `foo`, unless you remove the `const` attribute. For this to work out, `foo` must return a value, and that value must be consumed, e.g., by the `use` function — LLVM would otherwise recognize calls to `foo` as superfluous.

Perhaps the same trick would work with the memory-safety checking procedure?

zvonimir:
I agree that sticking with LLVM IR is better when possible. So I guess let's see if it is doable in this case.
Take a look at this regression:
https://github.com/smackers/smack/blob/master/test/memory-safety/structure1.c
When you execute SMACK on this with `--memory-safety` enabled, it will report collapsing of arguments when the memory safety function check gets invoked. Adding the `const` attribute does not help in this case. In fact, I am not sure exactly why the collapse happens since we are only casting them into `i8*`. I think we should explore this further.

michael-emmi:
But there is only one region in this example even with memory safety turned off.

zvonimir:
[...]

PR #322, opened 2018-06-08
Avoid mem safety region collapse

michael-emmi:
Attempting to salvage a possibly outdated fix for Issue #197 to avoid region collapse due to memory safety checking.

### DIFF
--- a/lib/smack/MemorySafetyChecker.cpp
+++ b/lib/smack/MemorySafetyChecker.cpp
@@ -40,8 +40,13 @@ void inserMemoryAccessCheck(Value* memoryPointer, Instruction* I, DataLayout* da
 
 bool MemorySafetyChecker::runOnModule(Module& m) {
   DataLayout* dataLayout = new DataLayout(&m);
-  Function* memorySafetyFunction = m.getFunction(Naming::MEMORY_SAFETY_FUNCTION);
+  LLVMContext& C = m.getContext();
+
+  Function* memorySafetyFunction = dyn_cast<Function>(m.getOrInsertFunction(Naming::MEMORY_SAFETY_FUNCTION, FunctionType::get(Type::getVoidTy(C), {PointerType::getUnqual(Type::getInt8Ty(C)), Type::getInt64Ty(C)}, false)));
   assert(memorySafetyFunction != NULL && "Memory safety function must be present.");
+  memorySafetyFunction->addFnAttr(Attribute::AttrKind::ReadNone);
+  memorySafetyFunction->addFnAttr(Attribute::AttrKind::NoUnwind);
+
   for (auto& F : m) {
     if (!Naming::isSmackName(F.getName())) {
       if (SmackOptions::isEntryPoint(F.getName())) {
@@ -78,4 +83,3 @@ bool MemorySafetyChecker::runOnModule(Module& m) {
 // Pass ID variable
 char MemorySafetyChecker::ID = 0;
 }
-
--- a/lib/smack/MemorySafetyChecker.cpp
+++ b/lib/smack/MemorySafetyChecker.cpp
@@ -16,68 +16,95 @@ namespace smack {
 
 using namespace llvm;
 
-void insertMemoryLeakCheck(Function& F, Module& m) {
-  Function* memoryLeakCheckFunction = m.getFunction(Naming::MEMORY_LEAK_FUNCTION);
-  assert (memoryLeakCheckFunction != NULL && "Memory leak check function must be present.");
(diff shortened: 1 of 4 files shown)


zvonimir:
Well, some regressions are failing. I guess that would be the first thing to investigate.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -16,66 +16,97 @@ namespace smack {
...
-  assert (memoryLeakCheckFunction != NULL && "Memory leak check function must be present.");
-  for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
-    if (isa<ReturnInst>(&*I)) {
-      CallInst::Create(memoryLeakCheckFunction, "", &*I);
+Function* MemorySafetyChecker::getLeakCheckFunction(Module& M) {
+  if (!leakCheckFunction.count(&M)) {
+    auto F = M.getFunction(Naming::MEMORY_LEAK_FUNCTION);
+    assert (F && "Memory leak check function must be present.");
+    leakCheckFunction[&M] = F;
+	}
```
zvonimir: Weird/wrong indentation.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -16,66 +16,97 @@ namespace smack {
...
+  return leakCheckFunction[&M];
+      }
```
zvonimir: Weird/wrong indentation.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -16,66 +16,97 @@ namespace smack {
...
+    F->addFnAttr(Attribute::AttrKind::NoUnwind);
+    safetyCheckFunction[&M] = F;
+  }
+  return safetyCheckFunction[&M];
+}
+
+void MemorySafetyChecker::insertMemoryLeakCheck(Instruction* I) {
+  auto& M = *I->getParent()->getParent()->getParent();
+  CallInst::Create(getLeakCheckFunction(M), "", I);
     }
```
zvonimir: Weird/wrong indentation.

lib/smack/MemorySafetyChecker.cpp
```diff
@@ -16,66 +16,97 @@ namespace smack {
...
+
+void MemorySafetyChecker::insertMemoryAccessCheck(Value* addr, Value* size, Instruction* I) {
+  auto& M = *I->getParent()->getParent()->getParent();
+  auto& C = M.getContext();
+  auto T = PointerType::getUnqual(Type::getInt8Ty(C));
+  CallInst::Create(getSafetyCheckFunction(M), {
+    CastInst::Create(Instruction::BitCast, addr, T, "", I),
+    CastInst::CreateBitOrPointerCast(size, T, "", I)
+  }, "", I);
   }
```
zvonimir: Weird/wrong indentation.

zvonimir:
I've just spent some time studying SVCOMP results, where this branch times out more than the develop branch, despite typically it seems splitting memory more.
One thing I discovered is that this branch adds memory checking to `__SMACK_static_init`, which for some benchmarks generated a huge number of additional memory checks as compared to the develop branch. So it would be good to avoid putting memory checks into `__SMACK_static_init`. I suspect this should be an easy fix.

zvonimir:
Another thing I observed. The new implementation always generates the following pattern:
```
$p12 := $bitcast.ref.ref($p11);
$p13 := $bitcast.ref.ref($i2p.i64.ref(8));
call __SMACK_check_memory_safety($p12, $p13);
```
The old one generates the following:
```
$p11 := $bitcast.ref.ref($p10);
call __SMACK_check_memory_safety($p11, 8);
```
As you can see, the new implementation has an addition bitcast, instead of passing the constant directly. I am not sure how much that affects the performance.

michael-emmi:
Good catch: this was missing a tiny piece of logic that avoids inserting the memory-safety checks in Smack-generated functions. Commit 1174fa4 fixes that, which should bring the SVCOMP results.

The additional bit cast is needed to make sure that integer-literal size arguments have the same type as pointers, so that memory-safety-checking arithmetic can be done in the arithmetic theory underlying pointers, which would typically be integers rather than bit vectors.

zvonimir:
This looks good to me. Just waiting for SVCOMP to finish.

zvonimir:
It is slightly better on SVCOMP results, in particular in the memory safety categories. The difference is quite small, and it times out maybe on 10 benchmarks less. Merging now.

PR #327, opened 2018-06-17
Rounding mode support

liammachado:
This adds support for Boogie's 'rmode' type and the fenv.h functions
fegetround() and fesetround().


lib/smack/SmackRep.cpp
```diff
@@ -1166,6 +1166,7 @@ Decl* SmackRep::getInitFuncs() {
   Block* b = Block::block();
   for (auto name : initFuncs)
     b->addStmt(Stmt::call(name));
+  b->addStmt(Stmt::assign(Expr::id("$rmode"), Expr::rmode("RNE")));
```
zvonimir: You should use `RMODE_VAR` here instead of the hard-coded variable name.

share/smack/lib/fenv.c
```diff
@@ -0,0 +1,26 @@
+#include <fenv.h>
+#include <smack.h>
+
+int fegetround()
+{
```
zvonimir: Change this into `int fegetround(void) {`

share/smack/lib/fenv.c
```diff
@@ -0,0 +1,26 @@
...
+  int ret = __VERIFIER_nondet_int();
+  __SMACK_code("if ($rmode == RNE) {@ := 1;}", ret);
+  __SMACK_code("if ($rmode == RNA) {@ := 2;}", ret);
+  __SMACK_code("if ($rmode == RTP) {@ := 3;}", ret);
+  __SMACK_code("if ($rmode == RTN) {@ := 4;}", ret);
+  __SMACK_code("if ($rmode == RTZ) {@ := 5;}", ret);
+  return ret;
+}
+
+int fesetround(int rm)
```
zvonimir: Do `int fesetround (int rm) {`

share/smack/lib/fenv.c
```diff
@@ -0,0 +1,26 @@
...
+{
+  switch (rm) {
```
zvonimir: Here is what the spec I found says:
```
Changes the currently selected rounding mode to round. If round does not correspond to one of the supported rounding modes nothing is changed. fesetround returns zero if it changed the rounding mode, or a nonzero value if the mode is not supported.
```
So this code should be changed to return a non-zero value if the mode is not supported (in which case it also does not get changed). I hope this makes sense.

share/smack/lib/fenv.c
```diff
@@ -0,0 +1,26 @@
+#include <fenv.h>
+#include <smack.h>
+
+int fegetround()
+{
+  int ret = __VERIFIER_nondet_int();
```
zvonimir: This spec says:
```
Returns the floating point rounding macro describing the current rounding direction or a negative value if the direction cannot be determined.
```
So `ret` should be negative if rounding mode cannot be determined. I am not sure if that could ever happen, but I guess it would be still good to account for it.

zvonimir: I think an easy fix is to add:
```
assume(ret < 0);
```
right after the declaration of ret.

include/smack/BoogieAst.h
```diff
@@ -83,6 +84,13 @@ class BoolLit : public Expr {
   void print(std::ostream& os) const;
 };
 
+class RModeLit : public Expr {
```
zvonimir: Where is this used?

zvonimir: Don't you have to update all floating-point operations that are generated by smack to use `$rmode`?

zvonimir:
Could you please also add several regressions for this?

include/smack/BoogieAst.h
liammachado: This is used by the change to SmackRep.cpp above. I update the floating-point operations in [this pull request]<https://github.com/smackers/smack/pull/328>.

zvonimir: I see. Ok, this should then probably have a check to make sure that only valid rounding modes are used. I guess we should define them as string constants in `Naming.h` as well. And then I think that `val` should not be a string type, but rather an `enum` type.

lib/smack/SmackRep.cpp
```diff
@@ -820,7 +820,11 @@ const Expr* SmackRep::cast(const llvm::ConstantExpr* CE) {
 }
 
 const Expr* SmackRep::cast(unsigned opcode, const llvm::Value* v, const llvm::Type* t) {
-  return Expr::fn(opName(Naming::INSTRUCTION_TABLE.at(opcode), {v->getType(), t}), expr(v));
+  std::string fn = Naming::INSTRUCTION_TABLE.at(opcode);
+  if (fn[1] == 'f' || fn[3] == '2') { //$fptrunc, $fpext, $fp2ui, $fp2si, $ui2fp, $si2fp
```
zvonimir: It would be better to list all of the opcodes here, instead of relying on string names.

lib/smack/SmackRep.cpp
```diff
@@ -833,6 +837,9 @@ const Expr* SmackRep::bop(const llvm::BinaryOperator* BO) {
 
 const Expr* SmackRep::bop(unsigned opcode, const llvm::Value* lhs, const llvm::Value* rhs, const llvm::Type* t) {
   std::string fn = Naming::INSTRUCTION_TABLE.at(opcode);
+  if (fn[1] == 'f' && fn[2] != 'r') { //$fadd, $fsub, $fmul, $fdiv
```
zvonimir: Same as above.

test/rmode/change_rm.c
```diff
@@ -0,0 +1,18 @@
+#include "smack.h"
+#include <fenv.h>
+
+// @expect verified
+
+int main()
```
zvonimir: Should be `int main(void) {`

test/rmode/change_rm.c
```diff
@@ -0,0 +1,18 @@
+#include "smack.h"
```
zvonimir: Create a failing version (called change_rm_fail.c) as well.

test/rmode/change_rm.c
zvonimir: This is true for other regressions as well.

test/rmode/change_rm.c
zvonimir: Do this for all of them.

test/rmode/config.yml
```diff
@@ -0,0 +1 @@
+flags: [--float]
```
zvonimir: Let's not create a new regressions folder just yet. Please put all of these into the floats folder.

zvonimir:
Several regressions are failing for some reason.

liammachado:
The regressions that are failing are float/floor.c and float/floor_fail.c. They are failing because I have changed the declarations of many floating-point functions in smack.c, but have not updated math.c to reflect these changes. The changes to math.c are in [this pull request](https://github.com/smackers/smack/pull/328).

test/float/set_rm_invalid_fail.c
```diff
@@ -0,0 +1,14 @@
+#include "smack.h"
+#include <fenv.h>
+
+// @expect error
+
+int main(void)
+{
```
zvonimir: Move this into the previous line, like this: `int main(void) {`
Do that with all your regressions to make them consistent with others.

zvonimir:
One more thing I almost forgot. Please add a regression involving a floating-point operation (or several of them) that passes only when rounding mode is successfully changed.

zvonimir:
Also, could you bring in this fix from your math.c pull request so that all regressions pass here?
It is a bit weird to merge something with failing regressions :). Thanks!

include/smack/Naming.h
```diff
@@ -90,6 +91,9 @@ class Naming {
   static const std::string MEMORY_SAFETY_FUNCTION;
   static const std::string MEMORY_LEAK_FUNCTION;
 
+  enum RMode { RNE, RNA, RTP, RTN, RTZ };
```
keram88: This is used like an enum class in Naming.cpp. Please update this declaration accordingly.

keram88:
It would be nice to have a regression like
```C
// ...
fesetround(FE_DOWNWARD);
if (__VERIFIER_nondet_int())
  fesetround(FE_UPWARD);
// ...
x = Some FP calculation;
assert(x == downward_result || x == upward_result);
```
And the corresponding failure cases:
assert(x == downward_result); and
assert(x == upward_result);

This is mainly to because of my suspicion of global state. Other than that, this looks like it will model rounding correctly.

lib/smack/SmackRep.cpp
```diff
@@ -820,7 +820,11 @@ const Expr* SmackRep::cast(const llvm::ConstantExpr* CE) {
 }
 
 const Expr* SmackRep::cast(unsigned opcode, const llvm::Value* v, const llvm::Type* t) {
-  return Expr::fn(opName(Naming::INSTRUCTION_TABLE.at(opcode), {v->getType(), t}), expr(v));
+  std::string fn = Naming::INSTRUCTION_TABLE.at(opcode);
+  if (fn == "$fptrunc" || fn == "$fpext" || fn == "$si2fp") {
```
zvonimir: Can't you use `opcode` here instead of `fn`?

lib/smack/SmackRep.cpp
```diff
@@ -833,6 +837,9 @@ const Expr* SmackRep::bop(const llvm::BinaryOperator* BO) {
 
 const Expr* SmackRep::bop(unsigned opcode, const llvm::Value* lhs, const llvm::Value* rhs, const llvm::Type* t) {
   std::string fn = Naming::INSTRUCTION_TABLE.at(opcode);
+  if (fn == "$fadd" || fn == "$fsub" || fn == "$fmul" || fn == "$fdiv") {
```
zvonimir: Can't you use `opcode` here instead of `fn`?

share/smack/lib/fenv.c
```diff
@@ -0,0 +1,26 @@
+#include <fenv.h>
```
zvonimir: Missing LICENSE stuff.

test/float/get_rm_invalid.c
```diff
@@ -0,0 +1,13 @@
+#include "smack.h"
+#include <fenv.h>
+
+// @expect verified
+
+int main(void) {
+  int rm = fegetround();
+  assume(rm < 1 || rm > 5);
```
zvonimir: Hmmm, I don't get this regression. What is it testing for? Please clarify.

include/smack/Naming.h
```diff
@@ -90,6 +91,9 @@ class Naming {
   static const std::string MEMORY_SAFETY_FUNCTION;
   static const std::string MEMORY_LEAK_FUNCTION;
 
+  enum class RMode { RNE, RNA, RTP, RTN, RTZ };
```
zvonimir: Ok, I gave this some more thought. The right place for this enum is in the RModeLit class if you think about it. It should not be in Naming.h. I think that we also should not have RMODE_TABLE. I would instead put that directly into the print function of RModeLit.
If you look at where the other naming tables are used, you'll see that they are not used for printing, but instead while particular Boogie expressions are being generated in SmackRep. So I think what I suggested above makes more sense.
Finally, I think if you do what you suggested, you will probably also be able to revert your changes to cmake files since I think you won't run into linking issues.

zvonimir: One more thing. See here:
https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly
I think this enum type should be named RModeKind to respect the LLVM coding style.

lib/smack/SmackRep.cpp
zvonimir: Meaning, use `Instruction::FPTrunc` instead of `$fptrunc`.

zvonimir: How come `Instruction::FPToUI` is not in this if condition?

include/smack/Naming.h
liammachado: I agree with your reasoning. However, if we go with this, then the Expr class would have the following line:
`static const Expr* rmode(RModeLit::RModeKind v);`
But RModeLit is a subclass of Expr, so this would not be legal (a superclass accessing a subclass's variable). Can you think of a way around this, or am I missing something?

test/float/get_rm_invalid.c
liammachado: This tests for the fact that `fegetround()` returns a negative value if the rounding mode cannot be determined. In other words, if rm was not assigned a value between 1 and 5 (which represent the range of possible rounding modes, then it must be negative.

lib/smack/SmackRep.cpp
liammachado: Yes, this is a much better idea.

lib/smack/SmackRep.cpp
liammachado: I initally did have fp2si, fp2ui, and ui2fp in this if condition. However, I realized that these functions do not have any builtin functions equivalent to them, according to the [SMT-LIB standard](http://smtlib.cs.uiowa.edu/theories-FloatingPoint.shtml). Instead, these are modeled using uninterpreted functions, as seen [here](https://github.com/smackers/smack/pull/327/files#diff-85650691a95426ad3aec3acc90105837R878), [here](https://github.com/smackers/smack/pull/327/files#diff-85650691a95426ad3aec3acc90105837R977), and [here](https://github.com/smackers/smack/pull/327/files#diff-85650691a95426ad3aec3acc90105837L981) in smack.c.

include/smack/Naming.h
zvonimir: Hmmm, right. But enum is just a type and not a class member, i.e., it is not like a class field. So if you make it public in the subclass, can you not simply refer to it in the superclass?
I am not totally sure about this, but it is worth quickly trying it out.

liammachado: It's an issue of circular dependency. For RModeLit::RModeKind to be visible to Expr, RModeLit must be defined before Expr. However, for RModeLit to extend Expr, Expr must be defined before RModeLit. Basically there is no way for Expr to access anything within RModeLit, regardless of whether they are types or class members.

PR #330, opened 2018-06-19
Floating point conversion declarations for Presage

keram88:
This adds undefined declarations for certain floating point conversions used by Presage. As the Presage tests do not depend on floating point for their verifier conditions, it is preferable to have some of the strangely named functions such as `$bitcast.i64.float` when running without floating point support on.


zvonimir:
Please squash this into one commit.

keram88:
Squashed.

PR #336, opened 2018-07-03
Fix to fp conversion functions

liammachado:
This does the following:
-Changes fp conversion functions such that they all accept an rmode
-Changes si2fp to be uninterpreted when --bit-precise is not set


lib/smack/SmackRep.cpp
```diff
@@ -821,7 +821,7 @@ const Expr* SmackRep::cast(const llvm::ConstantExpr* CE) {
 
 const Expr* SmackRep::cast(unsigned opcode, const llvm::Value* v, const llvm::Type* t) {
   std::string fn = Naming::INSTRUCTION_TABLE.at(opcode);
-  if (opcode == Instruction::FPTrunc || opcode == Instruction::FPExt || opcode == Instruction::SIToFP) {
+  if (opcode == Instruction::FPTrunc || opcode == Instruction::FPExt || opcode == Instruction::SIToFP || opcode == Instruction::UIToFP || opcode == Instruction::FPToSI || opcode == Instruction::FPToUI) {
```
zvonimir: Please break this up into two lines.

PR #345, opened 2018-07-16
Validate file arguments

shaobo-he:
Previously SMACK exits with an argument parsing error when the input
file is not valid. This commit rewrites the logic and puts it as an
action class. Moreover, the action class added also handles the validation
of output files.

I think this pull request helps us with the case where we want to debug a program by dumping its `.ll` file while accidentally write some nonsense file path as the argument. Then we wait a while and SMACK just can't write to that file.


zvonimir:
Something is broken it seems. Please check the failing regressions.

PR #346, opened 2018-07-20
Fixed implementations of copysign and __signbit

liammachado:
This fixes these two functions so that they are able to examine/set the
sign bit of NaN.

### DIFF
--- a/share/smack/include/math/math.h
+++ b/share/smack/include/math/math.h
@@ -17,24 +17,20 @@
 #define fpclassify(x) (sizeof(x) == sizeof(double) ? __fpclassify(x) : __fpclassifyf(x))
 #define isfinite(x) (sizeof(x) == sizeof(double) ? __finite(x) : __finitef(x))
 
-typedef union
-{
+typedef union {
   float f;
-  struct
-  {
-    int sigAndExp: 31;
-    int sign: 1;
+  struct {
+    unsigned sigAndExp: 31;
+    unsigned sign: 1;
   } i;
 } fi;
 
-typedef union
-{
+typedef union {
   double d;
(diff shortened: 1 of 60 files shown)


share/smack/include/math/math.h
```diff
@@ -17,6 +17,27 @@
 #define fpclassify(x) (sizeof(x) == sizeof(double) ? __fpclassify(x) : __fpclassifyf(x))
 #define isfinite(x) (sizeof(x) == sizeof(double) ? __finite(x) : __finitef(x))
 
+typedef union
+{
+  float f;
+  struct
+  {
+    int sigAndExp: 31;
```
zvonimir: Hmmm, you are using bit fields here. I am not sure if we support them properly in SMACK.
It would be good to first create a simple non-floating-point regression to check if bit fields actually work as expected.

share/smack/include/math/math.h
```diff
@@ -17,6 +17,27 @@
 #define fpclassify(x) (sizeof(x) == sizeof(double) ? __fpclassify(x) : __fpclassifyf(x))
 #define isfinite(x) (sizeof(x) == sizeof(double) ? __finite(x) : __finitef(x))
 
+typedef union
```
zvonimir: Open curly braces in the same line.

zvonimir:
Some regressions are failing. This should be fixed before we merge this.

share/smack/include/math/math.h
zvonimir: Maybe this should be `unsigned int`?

zvonimir:
Also, do we already have a regression that is testing for this?

PR #354, opened 2018-08-06
Removed all uses of INC(x) and DEC(x) macros

liammachado:
This pull request:
-Removes the INC(x) and DEC(x) macros from smack-macros.h
-Removes all macros in smack-macros.h and smack.c which use INC(x) or
DEC(x)
-Moves code generation that used INC(x) and DEC(x) to top.py


share/smack/top.py
```diff
@@ -1,6 +1,7 @@
 import argparse
 import errno
 import io
+import itertools
```
zvonimir: What do you need this module for?

share/smack/top.py
```diff
@@ -497,10 +498,136 @@ def llvm_to_bpl(args):
...
+      t1 = "%s%s" % (prefix, sizes[i])
+      t2 = "%s%s" % (prefix, sizes[j])
+      if "trunc" not in func_decl:
+        t1, t2 = t2, t1
+      tmp = args + (t1, t2, sizes[j])
+      bpl += func_decl.format(*tmp)
+  return bpl
+
+def generate_prelude(args):
+  with open(args.bpl_file, 'a+') as f:
```
zvonimir: I think you should open the file only after you created the `bpl` string. No need to keep it open while string creation is happening.

share/smack/top.py
```diff
@@ -497,10 +498,136 @@ def llvm_to_bpl(args):
   if args.float: cmd += ['-float']
   if args.modular: cmd += ['-modular']
   try_command(cmd, console=True)
+  generate_prelude(args)
   annotate_bpl(args)
   property_selection(args)
   transform_bpl(args)
 
+UNINTERPRETED_UNARY_OP = "function {0}.{1}(i: {1}) returns ({1});\n"
```
zvonimir: Let's move this to a new python file after all. Maybe name it `prelude.py`.

shaobo-he:
My general suggestion is to create a function class to represent functions in Boogie so that we don't need to write error-prone strings at the beginning of `prelude.py`. For example, we can have such a function class defined as follows,

```Python
class Function
  def __init__(self, attrs, name, arg_types, ret_type, body):
    self.attrs = attrs
    self.name = name
    self.types = types
    self.vars = map(lambda t : 'i'+t, types)
    self.body = body

  def toStr(self):
    pass
```

Then `UNINTERPRETED_UNARY_OP` would simply become `lambda bitwidth, name: Function(None, '.'.join([name, mk_int(bitwidth)], [mk_int(bitwidth)], mk_int(bitwidth), None)` where `mk_int = lambda bitwidth: 'int'+bitwidth`.

share/smack/prelude.py
```diff
@@ -0,0 +1,124 @@
+UNINTERPRETED_UNARY_OP = "function {0}.{1}(i: {1}) returns ({1});\n"
```
shaobo-he: Please see my comments earlier.

share/smack/prelude.py
```diff
@@ -0,0 +1,124 @@
...
+      t1 = "%s%s" % (prefix, sizes[i])
+      t2 = "%s%s" % (prefix, sizes[j])
+      if "trunc" not in func_decl:
+        t1, t2 = t2, t1
+      tmp = args + (t1, t2, sizes[j])
+      bpl += func_decl.format(*tmp)
+  return bpl
+
+def generate_prelude(args):
+  bpl = declare_each_type("bv", BVBUILTIN_BINARY_OP, "bvadd", "$add")
```
zvonimir: Kind of related to @shaobo-he's comment, I wonder if you could refactor this into loops (over lists of strings representing function names such as `$add`) instead of listing them explicitly line by line.

share/smack/prelude.py
```diff
@@ -0,0 +1,129 @@
...
+INLINE_BVBUILTIN_BINARY_PRED = lambda bitwidth, name: Function('inline', None, '.'.join([name, mk_bv(bitwidth)]), [mk_bv(bitwidth), mk_bv(bitwidth)], 'bv1', 'if ' + '.'.join([name, mk_bv(bitwidth), 'bool']) + '(i1,i2) then 1bv1 else 0bv1')
+INLINE_CONVERSION = lambda bitwidth2, bitwidth1, name, body: Function('inline', None, '.'.join([name, mk_int(bitwidth1), mk_int(bitwidth2)]), [mk_int(bitwidth1)], mk_int(bitwidth2), body)
+TRUNC_OP = lambda bitwidth1, bitwidth2: Function('inline', None, '.'.join(['$trunc', mk_int(bitwidth1), mk_int(bitwidth2)]), [mk_int(bitwidth1)], mk_int(bitwidth2), 'i1')
+TRUNC_BV_OP = lambda bitwidth1, bitwidth2: Function('inline', None, '.'.join(['$trunc', mk_bv(bitwidth1), mk_bv(bitwidth2)]), [mk_bv(bitwidth1)], mk_bv(bitwidth2), 'i1[' + str(bitwidth2) + ':0]')
+
+sizes = [1, 8, 16, 24, 32, 40, 48, 56, 64, 88, 96, 128]
+binary_ops = ['$add', '$sub', '$mul', '$sdiv', '$smod', '$srem', '$udiv', '$urem', '$shl', '$lshr', '$ashr', '$and', '$or', '$xor', '$nand']
```
zvonimir: Could we not have just one list containing names without `$` or `bv` prefixes. Then from that list you could add `$` or `bv` as needed. That would be better I think.

share/smack/prelude.py
```diff
@@ -0,0 +1,129 @@
...
+      func_str += ' {'+ self.body + '}'
+    else:
+      func_str += ';'
+    func_str += '\n'
+    return func_str
+
+mk_int = lambda bitwidth: 'i' + str(bitwidth)
+mk_bv = lambda bitwidth: 'bv' + str(bitwidth)
+
+UNINTERPRETED_UNARY_OP = lambda bitwidth, name: Function(None, None, '.'.join([name, mk_int(bitwidth)]), [mk_int(bitwidth)], mk_int(bitwidth), None)
```
zvonimir: Could you format this to look nicer and more readable somehow? Maybe split arguments to Function into multiple lines?

share/smack/prelude.py
```diff
@@ -0,0 +1,129 @@
...
+INLINE_CONVERSION = lambda bitwidth2, bitwidth1, name, body: Function('inline', None, '.'.join([name, mk_int(bitwidth1), mk_int(bitwidth2)]), [mk_int(bitwidth1)], mk_int(bitwidth2), body)
+TRUNC_OP = lambda bitwidth1, bitwidth2: Function('inline', None, '.'.join(['$trunc', mk_int(bitwidth1), mk_int(bitwidth2)]), [mk_int(bitwidth1)], mk_int(bitwidth2), 'i1')
+TRUNC_BV_OP = lambda bitwidth1, bitwidth2: Function('inline', None, '.'.join(['$trunc', mk_bv(bitwidth1), mk_bv(bitwidth2)]), [mk_bv(bitwidth1)], mk_bv(bitwidth2), 'i1[' + str(bitwidth2) + ':0]')
+
+sizes = [1, 8, 16, 24, 32, 40, 48, 56, 64, 88, 96, 128]
+binary_ops = ['$add', '$sub', '$mul', '$sdiv', '$smod', '$srem', '$udiv', '$urem', '$shl', '$lshr', '$ashr', '$and', '$or', '$xor', '$nand']
+bvbuiltin_binary_funcs = ['bvadd', 'bvsub', 'bvmul', 'bvsdiv', 'bvsmod', 'bvsrem', 'bvudiv', 'bvurem', 'bvshl', 'bvlshr', 'bvashr', 'bvand', 'bvor', 'bvxor', 'bvnand']
+builtin_binary_funcs = ['div', 'mod', 'rem', 'div', 'rem']
+binary_ops_body = ['i1 + i2', 'i1 - i2', 'i1 * i2']
+comp_ops = ['$ule', '$ult', '$uge', '$ugt', '$sle', '$slt', '$sge', '$sgt']
```
zvonimir: Same as above. Maybe you could have just one array, and then you add `$` or `bv` as needed.

share/smack/prelude.py
```diff
@@ -0,0 +1,250 @@
...
+    for j in range(0,i):
+      tmp = (sizes[i], sizes[j]) + args
+      bpl += func_decl(*tmp).toStr()
+  return bpl
+
+def generate_prelude(args):
+  bpl = ''
+  for i in range(len(binary_ops)):
+    op_name = '$' + binary_ops[i]
+    bpl += declare_each_type(BVBUILTIN_BINARY_OP, 'bv' + binary_ops[i], op_name)
```
michael-emmi: It would be clearer to just iterate over the sizes rather than hide it in these functions, e.g.:
```Python
bpl += map(lambda size: bv_builtin_binary_op(bit_width = size, builtin_name = 'bv' + op, name = '$' + op, sizes)
```
or alternatively:
```Python
for size in sizes:
    bpl += bv_builtin_binary_op(…)
```

share/smack/prelude.py
```diff
@@ -0,0 +1,250 @@
...
+
+binary_ops = ['add', 'sub', 'mul', 'sdiv', 'smod', 'srem', 'udiv', 'urem', 'shl', 'lshr', 'ashr', 'and', 'or', 'xor', 'nand']
+comp_ops = ['ule', 'ult', 'uge', 'ugt', 'sle', 'slt', 'sge', 'sgt']
+eq_ops = ['eq', 'ne']
+
+binary_ops_body = ['i1 + i2', 'i1 - i2', 'i1 * i2']
+comp_ops_body = ['i1 <= i2', 'i1 < i2', 'i1 >= i2', 'i1 > i2', 'i1 <= i2', 'i1 < i2', 'i1 >= i2', 'i1 > i2']
+eq_ops_body = ['i1 == i2', 'i1 != i2']
+
+def declare_each_type(func_decl, *args):
```
michael-emmi: These functions seem to obscure things rather than simplify and clarify. See comment below on iterating over sizes.

share/smack/prelude.py
```diff
@@ -0,0 +1,250 @@
...
+def declare_each_type_nested(func_decl, *args):
+  bpl = ''
+  for i in range(len(sizes)):
+    for j in range(0,i):
+      tmp = (sizes[i], sizes[j]) + args
+      bpl += func_decl(*tmp).toStr()
+  return bpl
+
+def generate_prelude(args):
+  bpl = ''
```
michael-emmi: If you defined `bpl` as a list, you could extend it more naturally with `bpl += map(lambda size: …, sizes)`.

share/smack/prelude.py
```diff
@@ -0,0 +1,250 @@
...
+
+  bpl += declare_each_type_nested(INLINE_CONVERSION, '$zext', 'i1')
+  bpl += declare_each_type_nested(INLINE_CONVERSION, '$sext', 'i1')
+
+  bpl += declare_each_type(SAFE_LOAD_OP, 'i1[i2]')
+  bpl += declare_each_type(SAFE_LOAD_BV_OP, 'i1[i2]')
+  bpl += declare_each_type(SAFE_STORE_OP, 'i1[i2 := i3]')
+  bpl += declare_each_type(SAFE_STORE_BV_OP, 'i1[i2 := i3]')
+
+  f = open(args.bpl_file, 'a+')
```
michael-emmi: Perhaps you can separate the writing of a given prelude to a file, e.g., `append_prelude`, from the construction of a prelude, e.g., `get_prelude`.

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
+import itertools
+
+class Attribute:
+  def __init__(self, attribute, attribute_arg = None):
```
michael-emmi: This seems vague and redundant. How about `name` and `arguments` rather than `attribute` and `attribute_arg`. (Note that Boogie attribute arguments are generally a comma separated list, so this should probably be a list too.)

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
...
+class Function:
+  def __init__(self, name, attributes = [], arguments = [], return_type = None, body = None):
+    self.name = name
+    self.attributes = attributes
+    self.arguments = arguments
+    self.return_type = return_type
+    self.body = body
+
+  def toStr(self):
+    function_str = 'function'
```
michael-emmi: This method of string building is a bit messy; also inefficient:
http://blog.mclemon.io/python-efficient-string-concatenation-in-python-2016-edition

Better to append each chunk to a list `chunks = []` and return `" ".join(chunks)` at the end.

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
...
+      function_str += ' returns (' + self.return_type + ')'
+    if self.body is not None:
+      function_str += ' {'+ self.body + '}'
+    else:
+      function_str += ';'
+    function_str += '\n'
+    return function_str
+
+
+def int(bit_width):
```
michael-emmi: This argument can be shortened to `width`.

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
...
+  return 'i' + str(bit_width)
+
+def bv(bit_width):
```
michael-emmi: This argument can be shortened to `width`.

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
...
+  return 'bv' + str(bit_width)
+
+def uninterpreted_unary_op(bit_width, base_name):
```
michael-emmi: These arguments can be shortened to `width` and `name`.

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
...
+  return Function(
+    name = base_name + '.' + int(bit_width),
+    arguments = [Argument(name = 'i', type = int(bit_width))],
```
michael-emmi: You can pass the parameters `Argument('i', int(width))` directly without keyword arguments, since there are only two.

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
...
+
+def int(bit_width):
+  return 'i' + str(bit_width)
+
+def bv(bit_width):
+  return 'bv' + str(bit_width)
+
+def uninterpreted_unary_op(bit_width, base_name):
+  return Function(
+    name = base_name + '.' + int(bit_width),
```
michael-emmi: Why not declare a local variable `type = int(bit_width)` rather than repeating the latter over and over.

share/smack/prelude.py
michael-emmi: This comment can be applied throughout.

share/smack/prelude.py
michael-emmi: This comment can be applied throughout.

share/smack/prelude.py
michael-emmi: This comment can be applied throughout.

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
...
+    arguments = [Argument(name = 'i1', type = int(bit_width)), Argument(name = 'i2', type = int(bit_width))],
+    return_type = int(bit_width)
+  )
+
+def inline_binary_op(bit_width, base_name, body):
+  return Function(
+    name = base_name + '.' + int(bit_width),
+    attributes = [Attribute(attribute = 'inline')],
```
michael-emmi: No need to use keyword arguments with a single argument which is obviously the attribute name. Just write `Attribute('inline')`.

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
...
+    bpl += map(lambda size: bvbuiltin_binary_op(bit_width = size, builtin_name = 'bv' + op, base_name = '$' + op).toStr(), sizes)
+    if 0 <= i <= 2:
+      bpl += map(lambda size: inline_binary_op(bit_width = size, base_name = '$' + op, body = binary_ops_body[i]).toStr(), sizes)
+    elif 3 <= i <= 7:
+      bpl += map(lambda size: builtin_binary_op(bit_width = size, builtin_name = op[1:], base_name = '$' + op).toStr(), sizes)
+    else:
+      bpl += map(lambda size: uninterpreted_binary_op(bit_width = size, base_name = '$' + op).toStr(), sizes)
+
+  for i, op in enumerate(comp_ops):
+    bpl += map(lambda size: bvbuiltin_binary_comp(bit_width = size, builtin_name = 'bv' + op, base_name = '$' + op).toStr(), sizes)
```
michael-emmi: Does it matter the order in which these declarations are listed? Because this could be shortened significantly:
```Python
for i, op in enumerate(comp_ops):
    for size in sizes:
        bpl += bvbuiltin_binary_cop(…)
        bpl += inline_bvbuiltin_binary_pred(…)
        …
```

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
...
+
+  bpl += map(lambda size_pair: inline_conversion(bit_width = size_pair[0], bit_width_2 = size_pair[1], base_name = '$zext', body = 'i').toStr(), itertools.combinations(sizes, 2))
+  bpl += map(lambda size_pair: inline_conversion(bit_width = size_pair[0], bit_width_2 = size_pair[1], base_name = '$sext', body = 'i').toStr(), itertools.combinations(sizes, 2))
+
+  bpl += map(lambda size: safe_load_op(bit_width = size, body = 'M[p]').toStr(), sizes)
+  bpl += map(lambda size: safe_load_bv_op(bit_width = size, body = 'M[p]').toStr(), sizes)
+  bpl += map(lambda size: safe_store_op(bit_width = size, body = 'M[p := v]').toStr(), sizes)
+  bpl += map(lambda size: safe_store_bv_op(bit_width = size, body = 'M[p := v]').toStr(), sizes)
+
+  return ''.join(bpl)
```
michael-emmi: You could get rid of all of the previous calls to `toStr` by replacing this line with:
```Python
return ' '.join([str(x) for x in bpl])
```

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
+import itertools
+
+class Attribute:
+  def __init__(self, attribute, attribute_arg = None):
+    self.attribute = attribute
+    self.attribute_arg = attribute_arg
+
+  def toStr(self):
```
michael-emmi: This is not idiomatic Python; the convention is to define the `__str__(self)` method, which gets called when you call `str(x)` on your object.

share/smack/prelude.py
```diff
@@ -0,0 +1,290 @@
...
+def inline_binary_op(bit_width, base_name, body):
+  return Function(
+    name = base_name + '.' + int(bit_width),
+    attributes = [Attribute(attribute = 'inline')],
+    arguments = [Argument(name = 'i1', type = int(bit_width)), Argument(name = 'i2', type = int(bit_width))],
+    return_type = int(bit_width),
+    body = body
+  )
+
+def builtin_binary_op(bit_width, builtin_name, base_name):
```
shaobo-he: I would recommend that we group these functions by arity. For example, we can define a function as follows,

```def binary_op(bit_width, base_name, type, return_type, attrib=None, body=None):
  pass
```

share/smack/prelude.py
shaobo-he: Btw, we can write a function to avoid explicitly create a list of arguments. For example,

```python
def mk_args(type, num):
   return [Argument(str(i), type) for i in range(num)]
```

share/smack/prelude.py
liammachado: To be clear, are you proposing replacing builtin_binary_op, bvbuiltin_binary_op, etc. with binary_op, or are you proposing that builtin_binary_op, bvbuiltin_binary_op, etc. calls binary_op?

liammachado:
I've made changes that address Mike/Shaobo's additional suggestions. Do you guys have any more feedback?

michael-emmi:
This looks wildly better to me now 👍

PR #356, opened 2018-08-07
Compatibility with Python-3-default systems.

michael-emmi:
Smack only works with Python version 2. On systems where version 3 is the default, Smack can still work simply by specifying `python2` as the interpreter.


zvonimir:
Could you do this for all SMACK python scripts while you are at it?

PR #362, opened 2018-08-09
Added test for uninterpreted functions.

michael-emmi:
Inspired by a question asked by Nathan Chong, I thought it would be nice to share and preserve this recipe for functions which are _uninterpreted_ in the logical sense, in that a given call `f(x)` may return any value, but for `x == y` we are guaranteed that `f(x) == f(y)`.

### DIFF
--- a/test/basic/uninterpreted.c
+++ b/test/basic/uninterpreted.c
@@ -1,0 +1,19 @@
+#include <smack.h>
+
+// @expect verified
+
+int foo(int x) {
+  int y = __VERIFIER_nondet_int();
+  __SMACK_top_decl("function FOO(x: int): int;");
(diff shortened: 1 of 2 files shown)


zvonimir:
Ah, we have too many regressions in the `basic` folder now, and CI is timing out (there is a max time limit on every CI job). It would be good to split `basic` into two folders, and make sure to run CI on both of them. See travis config file for that. Maybe one meaningful folder would be `pointers` or something like that.

michael-emmi:
This was already a problem before I added these regressions — see all other concurrent pull requests. But I guess I’ll go ahead and fit it anyways, in a separate PR.

zvonimir:
Ah, sorry, I noticed it only now. Thanks for fixing!

PR #368, opened 2018-08-30
Add Rust installer to build.sh

keram88:
This allows build.sh to optionally install the Rust compiler for upcoming support for that language.


bin/build.sh
```diff
@@ -265,6 +266,14 @@ if [ ${INSTALL_DEPENDENCIES} -eq 1 ] && [ "$TRAVIS" != "true" ] ; then
       sudo apt-get install -y gobjc gnustep gnustep-make gnustep-common gnustep-devel
     fi
 
+    if [ ${INSTALL_RUST} -eq 1 ] ; then
```
zvonimir: This should be moved below where other languages are. Here we are only installing dependencies/packages.

bin/build.sh
```diff
@@ -365,6 +366,15 @@ if [ ${INSTALL_OBJECTIVEC} -eq 1 ] ; then
   puts "Installed Objective-C"
 fi 
 
+if [ ${INSTALL_RUST} -eq 1 ] ; then
+  ${WGET} https://static.rust-lang.org/dist/${RUST_VERSION}/rust-nightly-x86_64-unknown-linux-gnu.tar.gz -O rust.tar.gz
```
zvonimir: Add `puts Installing Rust` to match other installation blocks. Then this should be done. Squash your commits and I'll merge them.

keram88: Done

PR #371, opened 2018-09-09
Add support for the llvm.expect intrinsic

keram88:
This models the llvm.expect intrinsic when generating boogie code. Since this intrinsic behaves as the identity function for its first argument, the function call is simply replaced with its first argument. I.e,
```
%b = call i32 @llvm.expect.i32(i32 %a, i32 0)
->
$b := $a
```


shaobo-he:
LGTM. Maybe it's better if we put all the code that handle LLVM intrinsics into one pass (e.g., the one for overflow checking) such that we don't need these special cases of call instruction?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -594,6 +594,12 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
     WARN("ignoring llvm.debug call.");
     emit(Stmt::skip());
 
+  } else if (name.find("llvm.expect.") == 0) {
```
zvonimir: The style of this if condition should match the others in this if-else-if block. Please update this.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -594,6 +594,12 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
...
+    // The llvm.expect.* takes two values, a value v and an expected value.
```
zvonimir: Update to:
```
// The llvm.expect.* has two arguments: a value v and an expected value
// that is supposed to be used by optimizers.
// Semantically, this function simply evaluates to v.
```

lib/smack/SmackInstGenerator.cpp
keram88: Do you mean compare it to string::npos? My reasoning is that we want to be sure "llvm." is at the beginning of the function name. This guarantees that we are actually replacing an LLVM function, and not some other function. Perhaps I should change llvm.dbg to compare to 0, but it seems unlikely to matter either way.

zvonimir: I see, good point. I think what you wrote could probably be said for all of these conditions, including the ones below. So to be consistent, let's still use `!= std::string::npos` here.

shaobo-he:
@zvonimir The way I see the overflow checking pass is that it handles overflow intrinsics. As far as I remember it, this pass transforms these intrinsics even without overflow checking enabled. Therefore, we can rename this pass and put all intrinsic handling code in it.

lib/smack/SmackInstGenerator.cpp

lib/smack/SmackInstGenerator.cpp

PR #372, opened 2018-09-29
Rust frontend

keram88:
This adds the front end and libraries for Rust support. The biggest thing that I think needs to be looked at is the way the `smack.rs` library is included (lines 221-231 in `frontend.py`). A directory named `smack` is created where the test file is, and a symbolic link to smack's `smack.rs` is placed there. This is needed to enable the ability to compile the program normally and still have it be verifiable by SMACK.

### DIFF
--- a/share/smack/lib/smack.rs
+++ b/share/smack/lib/smack.rs
@@ -454,51 +454,6 @@ macro_rules! vec {
   };  
 }
 
-// #[cfg(verifier = "smack")]
-// pub struct String {
-//     backing_store: Vec<u8>,
-// }
-
-// #[cfg(verifier = "smack")]
-// impl String {
-//   pub fn from(s: &str) -> String {
-//     let mut bs = vec![];
-//     // let bytes = s.as_bytes();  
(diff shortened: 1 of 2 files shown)


share/smack/lib/smack.rs
```diff
@@ -0,0 +1,532 @@
...
+      result
+    }
+  };  
+}
+
+// #[cfg(verifier = "smack")]
+// pub struct String {
+//     backing_store: Vec<u8>,
+// }
+
```
jjgarzella: Just for clarification, why is all of this commented out? I assume it was causing slowdowns or something?

zvonimir: Good point @jjgarzella. @keram88 please add a brief comment explaining why this is commented out. Then I'll approve it.

keram88: Shaobo and I were using `String` in a benchmark. The way I wrote it doesn't quite work the way Rust's `String` works. I've removed it instead.

PR #381, opened 2018-10-31
D compile

jjgarzella:
Fix a few issues with compiling D files. 

TODO: fix the commit history of this PR so it only has one commit


zvonimir:
Looks good to me. Not sure how you will fix the history, but yeah, would be good to do it.

jjgarzella:
It required a bit of nasty force updates, but we should be good now.

Issue #382, opened 2018-10-31
Overflow due to left shift not being detected


zvonimir:
Left shift operation can cause an overflow, as per C11 standard. That happens when the result of shifting exceeds the max number that can be stored in the used signed integer type. For example, a signed integer overflow error should be reported for this example:
```
int main(void) {
  int x = __VERIFIER_nondet_int();
  int y;
  assume(x >= 1024);
  y = x << 30; // this overflows
  return y;
}
```
However, when we run SMACK on this example with `--integer-overflow` enabled it misses this overflow.

keram88:
```
The result of E1 << E2 is E1 left-shifted E2 bit positions; vacated
bits are ﬁlled with zeros. If E1 has an unsigned type, the value of
the result is E1 * 2^E2, reduced modulo one more than the
maximum value representable in the result type. If E1 has a signed
type and nonnegative value, and E1 * 2^E2 is representable in the
result type, then that is the resulting value; otherwise, the behavior
is undeﬁned.
```
Am I right that this means any non-negative left-shift is defined for unsigned values? If this is the case, then it seems for now that we're forced to choose between false negatives and false positives:
1) LLVM doesn't have an instruction like `llvm.add.with.overflow` for left shifts.
2) The only thing I can get Clang to do is generate a call to `__ubsan_handle_shift_out_of_bounds` with the `-fsanitize=shift` option, but it does this for every left shift including those with unsigned arguments.

keram88:
I was wrong about what Clang does with UBSan:  it generates the code to correctly detect undefined behaviors regarding left shifts, then conditionally calls `__ubsan_handle_shift_out_of_bounds` which prints out the error then terminates the program.

PR #383, opened 2018-11-01
Add checking for left-shift overflows in C programs

keram88:
Changes:
* `--integer-overflow` enables the clang option `-fsanitize=shift`.
* `__ubsan_handle_shift_out_of_bounds` is transformed to assert false with the `overflow` annotation.
* Tests for shifting with a negative negative LHS, left shifting non-negative signed-values, and unsigned values are included.

UBSan does all the work for us to detect undefined left bit shifts. It will check that the shift amount is appropriate for the bit width. For signed left shifts, it performs a right shift of `(bit-width - 1 - left_shift_amt)` bits and checks if this is zero. If it's not zero, then the left shift would shift a `1` past the MSB, which is undefined.

This is meant to fix issue #382.

### DIFF
--- a/lib/smack/IntegerOverflowChecker.cpp
+++ b/lib/smack/IntegerOverflowChecker.cpp
@@ -94,7 +94,7 @@ Value* IntegerOverflowChecker::createResult(Value* v, int bits, Instruction* i) 
  * This adds a call instruction to __SMACK_check_overflow to determine if an
  * overflow occured as indicated by flag.
  */
-void IntegerOverflowChecker::addCheck(Function* co, BinaryOperator* flag, Instruction* i) {
+void IntegerOverflowChecker::addCheck(Function* co, Value* flag, Instruction* i) {
   ArrayRef<Value*> args(CastInst::CreateIntegerCast(flag, co->arg_begin()->getType(), false, "", i));
   CallInst::Create(co, args, "", i);
 }
@@ -103,7 +103,7 @@ void IntegerOverflowChecker::addCheck(Function* co, BinaryOperator* flag, Instru
  * This inserts a call to assume with flag negated to prevent the verifier
  * from exploring paths past a __SMACK_check_overflow
  */
-void IntegerOverflowChecker::addBlockingAssume(Function* va, BinaryOperator* flag, Instruction* i) {
+void IntegerOverflowChecker::addBlockingAssume(Function* va, Value* flag, Instruction* i) {
   ArrayRef<Value*> args(CastInst::CreateIntegerCast(BinaryOperator::CreateNot(flag, "", i),
         va->arg_begin()->getType(), false, "", i));
   CallInst::Create(va, args, "", i);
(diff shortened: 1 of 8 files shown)


lib/smack/SmackInstGenerator.cpp
```diff
@@ -600,6 +600,8 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
     Value* val = ci.getArgOperand(0);
     emit(Stmt::assign(rep->expr(&ci), rep->expr(val)));
     
+  } else if (name.find("__ubsan_handle_shift_out_of_bounds") != std::string::npos) {
```
zvonimir: Could we move this into integer overflow pass?

test/bits/left_shift_negative.c
```diff
@@ -0,0 +1,12 @@
+#include "smack.h"
```
zvonimir: Add `_fail` to the end of the file name since that is our convention for files containing errors.

keram88:
That would make more sense. I don't know why I didn't think of that.

test/bits/left_shift_negative.c
```diff
@@ -0,0 +1,12 @@
...
+
+// @expect error
+// @flag --integer-overflow
+// @flag --bit-precise
```
zvonimir: No need for `--bit-precise` flag since config.yml already sets it.

test/bits/left_shift_overflow.c
```diff
@@ -0,0 +1,13 @@
+#include "smack.h"
+
+// @expect verified
+// @flag --integer-overflow
+// @flag --bit-precise
```
zvonimir: No need for `--bit-precise` flag since config.yml already sets it.

test/bits/left_shift_overflow_fail.c
```diff
@@ -0,0 +1,13 @@
+#include "smack.h"
+
+// @expect error
+// @flag --integer-overflow
+// @flag --bit-precise
```
zvonimir: No need for `--bit-precise` flag since config.yml already sets it.

test/bits/left_shift_unsigned.c
```diff
@@ -0,0 +1,13 @@
+#include "smack.h"
+
+// @expect verified
+// @flag --integer-overflow
+// @flag --bit-precise
```
zvonimir: No need for `--bit-precise` flag since config.yml already sets it.

lib/smack/IntegerOverflowChecker.cpp
```diff
@@ -119,6 +119,20 @@ bool IntegerOverflowChecker::runOnModule(Module& m) {
     if (Naming::isSmackName(F.getName()))
       continue;
     for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
+      // Add check for UBSan left shift
+      if (auto chkshft = dyn_cast<CallInst>(&*I)) {
+        if (chkshft->getCalledFunction() != nullptr &&
+            chkshft->getCalledFunction()->getName().find("__ubsan_handle_shift_out_of_bounds") != std::string::npos &&
```
zvonimir: I think you have to check here first if function actually has name using `hasName`. Maybe make it similar to this line:
https://github.com/smackers/smack/blob/901550325ca9112fb6d8008b19a9e66976ed169b/lib/smack/IntegerOverflowChecker.cpp#L140

keram88: Done.

lib/smack/IntegerOverflowChecker.cpp
```diff
@@ -119,6 +119,21 @@ bool IntegerOverflowChecker::runOnModule(Module& m) {
     if (Naming::isSmackName(F.getName()))
       continue;
     for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
+      // Add check for UBSan left shift
+      if (auto chkshft = dyn_cast<CallInst>(&*I)) {
+        if (chkshft->getCalledFunction() != nullptr &&
```
zvonimir: Maybe introduce a local function variable here so that you avoid calling `getCalledFunction` so many times.
Everything else looks good. So please squash these commits and in your final commit messages say that it `Fixes #383`.
Thanks!

keram88: Done. Poor Travis.

PR #385, opened 2018-11-06
Rust entry point fix

keram88:
This fixes setting the entry point for Rust program.

### DIFF
--- a/lib/smack/SmackInstGenerator.cpp
+++ b/lib/smack/SmackInstGenerator.cpp
@@ -599,7 +599,20 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
     // Semantically, this function simply returns the value v.
     Value* val = ci.getArgOperand(0);
     emit(Stmt::assign(rep->expr(&ci), rep->expr(val)));
-    
+
+    // Set the entry point for Rust programs
+  } else if (name.find(Naming::RUST_ENTRY) != std::string::npos) {
+    auto castExpr = ci.getArgOperand(0);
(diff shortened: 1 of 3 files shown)


lib/smack/SmackInstGenerator.cpp
```diff
@@ -600,9 +600,28 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
     Value* val = ci.getArgOperand(0);
     emit(Stmt::assign(rep->expr(&ci), rep->expr(val)));
     
+  } else if (name.find("std") != std::string::npos &&
+             name.find("rt") != std::string::npos &&
```
zvonimir: Could we not replace these with some kind of regular expressions?

zvonimir: Also, please add comments for what each `else if` block handles/does.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -600,9 +600,28 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
...
+
+  } else if (name.find("std") != std::string::npos &&
+             name.find("panicking") != std::string::npos &&
+             name.find("begin_panic") != std::string::npos) {
+    emit(Stmt::assert_(Expr::lit(false), {Attr::attr("rust_panic")}));
+    
   } else if (name.find(Naming::VALUE_PROC) != std::string::npos) {
     emit(rep->valueAnnotation(ci));
-
+
```
zvonimir: You added it seems spaces in this empty line.

PR #389, opened 2018-11-17
[WIP] Upgrade LLVM to 7.0

ecstatic-morse:
Sorry for the many PRs, reloading the page made github try to merge against master.

All tests with memory model `no-reuse-impls` either pass or timeout except for `basic/uninterpreted.c` which fails and `simd/shuffle{,_fail}.c` which won't compile due to the size of the control mask passed to `_mm_blend_epi32`. Some tests are failing with `reuse`.

I ask a specific question regarding generating Boogie code in [this commit's message](https://github.com/smackers/smack/commit/81682ca297607fe86db224a039965cc607da4e19).


zvonimir:
I will have more time to devote to this pull request after the SVCOMP deadline, which is in a week.
In the meantime, I looked at the CI results a bit. It is, for example, quite weird that this regression is now timing out: https://github.com/smackers/smack/blob/master/test/basic/limits.c
It is a really basic regression that does not touch memory, and so I would expect that even the generated bpl files should be the same when comparing old and new LLVM versions. Could you maybe compare the bpl files SMACK generates before and after your updates?

zvonimir:
Hi @ecstatic-morse. After we gave this some thought, we decided to move step by step to hopefully allow us easier root causing of any issues we might encounter. So I started working on an update to LLVM 5.0.1. As it turns out, there are many changes that have to be made for that update that you already made in this pull request.
Would you maybe be interested to help us out by first creating a pull request that update to LLVM 5.0.1? We would greatly appreciate your help.

zvonimir:
Sorry, I meant to say 4.0.1, not 5.0.1.

zvonimir:
I cherry-picked most of these commits into the develop branch while I was doing LLVM version updates recently. Thanks again @ecstatic-morse for your help!

PR #396, opened 2018-12-13
Fixed layout of external memory

zvonimir:
I am running SVCOMP to see if this has any effect on performance. We could maybe review it in the meantime.

### DIFF
--- a/include/smack/SmackRep.h
+++ b/include/smack/SmackRep.h
@@ -38,8 +38,8 @@ protected:
   std::vector<std::string> bplGlobals;
   std::map<const llvm::Value*, unsigned> globalAllocations;
 
-  long globalsBottom;
-  long externsBottom;
+  long globalsOffset;
+  long externsOffset;
   unsigned uniqueFpNum;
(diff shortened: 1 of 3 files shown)


michael-emmi:
What about the declaration of `$EXTERNS_BOTTOM` from `SmackRep::getPrelude`? Shouldn’t this also be computed as on offset from `$GLOBALS_BOTTOM`?

Perhaps the variable name `SmackRep::externsBottom` is now misleading, since this is not the “bottom” but rather an ”offset.”

zvonimir:
Good point. Check it out now.

bitcalc:
When will this fix be available in the master branch?

zvonimir:
There is a bit of delay due to xmas holidays. I'll start running SVCOMP benchmarks today to do a detailed performance comparison. I expect I'll merge this into the develop branch in 3-4 days from now. It'll only get merged into master with the next release.

zvonimir:
On SVCOMP benchmarks, this branch actually does slightly better than develop. So I am merging this.

Issue #384, opened 2018-11-02
isfinite() macro giving an error


jjgarzella:
When importing math.h (as a c++ header `#include <math.h>`), the isfinite macro seems to be giving an error. 

    #define isfinite(x) (sizeof(x) == sizeof(long double) ? __isfinitel(x) : sizeof(x) == sizeof(double) ? __finite(x) : __finitef(x))

It seems like `__isfinitel(x)` should be `__finitel(x)` to fit with the pattern, and sure enough below in the file, `__isfinitel` is not declared but `__finitel` is.

Before I submit a patch, I'd just like to confirm: which is the proper prefix here?

liammachado:
I apologize for leaving this for so long. Yes, please submit a patch changing `__isfinitel(x)` to `__finitel(x)`.

PR #399, opened 2019-01-01
Fixed isfinite() macro

liammachado:
The macro currently calls __isfinitel, but the function is called __finitel.


liammachado:
Fixes [Issue #384](https://github.com/smackers/smack/issues/384).

Issue #308, opened 2018-03-17
Generate smack.c via a Python script


shaobo-he:
Currently smack.c is hand-crafted and uses macros to capture patterns in function declarations. Therefore, smack.c is not well organized. Using a Python script to automatically generate smack.c is a promising solution to this problem.

zvonimir:
This is related to the discussion started in this pull request: https://github.com/smackers/smack/pull/307
I thought more about this and I do think it is probably a good idea. However, how would this work with large projects where we manually include `smack.h` and manually link with `smack.c`?
Instead of generating these files dynamically every time SMACK is run, how about we think of it as a parser generator, and actually have a python script that generates these files, which we then commit into the repo. I hope this makes sense.

michael-emmi:
What is the workflow you have in mind with this “manual linking” of `smack.c`? Do you mean that the user invokes `llvm2bpl` directly, rather than invoking the Python `smack` script?

In the latter case, the user presumably provides LLVM bitcode to `smack`, in which case we can generate and link the corresponding `smack.c` module, just before running `llvm2bpl`.

The potential problems I see with the user maintaining a `smack.c` module are that (a) their module may be inconsistent with future updates to Smack, since the `smack.c` we generate can change across versions, and (b) since the generated `smack.c` module depends on options to `smack` — currently memory safety and choice of memory model, but future versions could add additional dependencies — fixing `smack.c` would limit the user to verification queries with the same options, e.g., cannot do both with and without memory safety with the same `smack.c`.

My intuition on best practices for software maintenance suggests that we should try to avoid encouraging users to store a generated `smack.c` module.

zvonimir:
Coming back to this. Why would we generate `smack.c` automatically, instead of generating Boogie code directly (via a python script invoked from top.py)?

michael-emmi:
Good point!

I’d say generating Boogie code directly should be preferred, so long as none of those Boogie-code declarations actually depend on the compiler or preprocessor. I assumed we are talking about the declarations inside `__SMACK_decls`, and it seems that they are independent.

Would there be any reason that the other `__SMACK_` and `__VERIFIER_` function definitions could not be generated in the same way?

It seems that the only remaining function definitions in `smack.c` are `exit` and `calloc`, and they probably don’t belong there in any case. Their definitions are both given in C code, and according to our current conventions, those should go into a `stdlib.c` file.

michael-emmi:
[...]

PR #410, opened 2019-01-29
Move prelude generation from Python to C++

liammachado:
In [a previous pull request](https://github.com/smackers/smack/pull/354), some of the prelude generation code was moved from `smack.c` to `prelude.py`. This pull request moves the prelude generation code to a separate `Prelude` class and removes `prelude.py`.


lib/smack/SmackRep.cpp
```diff
@@ -1162,6 +1172,68 @@ std::string SmackRep::getPrelude() {
   }
   s << "\n";
 
+  for (size_t s1 = 0; s1 < INTEGER_SIZES.size(); ++s1) {
```
zvonimir: I think this block should be moved into the `Prelude` class, maybe in a function called `generatePrelude`. That way prelude generation would be more encapsulated. Unless you have a good reason for putting it here.

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+FuncDecl* Prelude::inlineBinaryOp(int width, std::string name, const Expr* body) {
+  std::string type = "i" + std::to_string(width);
+  return Decl::function(SmackRep::indexedName(name, {type}), {{"i1", type}, {"i2", type}}, type, body, {Attr::attr("inline")});
+}
+
+FuncDecl* Prelude::builtinBinaryOp(int width, std::string builtinName, std::string name) {
+  std::string type = "i" + std::to_string(width);
+  return Decl::function(SmackRep::indexedName(name, {type}), {{"i1", type}, {"i2", type}}, type, NULL, {Attr::attr("builtin", builtinName)});
```
zvonimir: Break these long lines at 80 characters. See: https://llvm.org/docs/CodingStandards.html#source-code-width

lib/smack/SmackRep.cpp
```diff
@@ -60,15 +61,24 @@ const std::vector<unsigned> REF_CONSTANTS = {
   1024
 };
 
-std::string indexedName(std::string name, std::initializer_list<std::string> idxs) {
+const std::vector<std::string> BINARY_OPS {"add", "sub", "mul", "sdiv", "smod", "srem", "udiv", "urem", "shl", "lshr", "ashr", "and", "or", "xor", "nand"};
```
zvonimir: These should be moved to `Prelude` as well.

lib/smack/SmackRep.cpp
liammachado: I put it there because `SmackRep.cpp` already had a `getPrelude()` function, and I thought it would make more sense to put all prelude generation code in a single function rather than separating it into two different functions across different classes. Then again, creating the `Prelude.cpp` class results in this effect anyway. Do you think `SmackRep.cpp` should have a `getPrelude()` function if we're creating a dedicated `Prelude` class, or do you think it's fine if some of the prelude generation happens in `SmackRep.cpp`?

zvonimir: I see. I think that ideally all of the prelude generation should be moved into your `Prelude` class. I hope that there are not too many dependencies that have to be passed to it from `SmackRep`. I just looked at the code and it doesn't seem that bad. So `SmackRep` should not have `getPrelude` once this is done.

zvonimir:
Hey @liammachado. I just removed some axioms from `smack.c`, see here: https://github.com/smackers/smack/commit/0efc78490c9f3dc64fa56d510591f8ffb64b9a29
Are you generating these in this pull request already? Please double-check that and let me know.

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+FuncDecl* Prelude::inlineBinaryComp(int width, std::string name,
+  const Expr* body) {
+  std::string type = "i" + std::to_string(width);
+  return Decl::function(SmackRep::indexedName(name, {type, "bool"}),
+    {{"i1", type}, {"i2", type}}, "bool", body, {Attr::attr("inline")});
+}
+
+FuncDecl* Prelude::inlineBinaryBvComp(int width, std::string name,
+  const Expr* body) {
```
zvonimir: Sorry for bugging you about this, but please format arguments that spill into more than one line like this: https://llvm.org/doxygen/CodeExtractor_8cpp_source.html#l00510
Just to keep is consistent with LLVM source.

liammachado:
No, this pull request doesn't interfere with those axioms.

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,438 @@
...
+  BinExpr::Minus, BinExpr::Times};
+
+const std::vector<BinExpr::Binary> COMP_OPS_BODY {BinExpr::Lte, BinExpr::Lt,
+  BinExpr::Gte, BinExpr::Gt, BinExpr::Lte, BinExpr::Lt, BinExpr::Gte,
+  BinExpr::Gt};
+
+const std::vector<BinExpr::Binary> EQ_OPS_BODY {BinExpr::Eq, BinExpr::Neq};
+
+FuncDecl* Prelude::uninterpretedUnaryOp(int width, std::string name) {
+  std::string type = "i" + std::to_string(width);
```
shaobo-he: I think we can encapsulate this line into a function like `getIntegerTypeName`.

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,438 @@
...
+  return Decl::function(SmackRep::indexedName(name, {type}), {{"i1", type}},
```
shaobo-he: `"i1"` is like magic number in C. We can avoid it by using a function like `makeIntVar` or `makeIntVars` depending on the number of arguments. By this way, you can also make the argument names consistent in the function body.

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,438 @@
...
+    type, NULL, {});
```
shaobo-he: `NULL` to `nullptr`, maybe? I'm not sure but I saw a lot of occurrences of `nullptr` in C++ rather than `NULL`.

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,438 @@
...
+  return Decl::function(SmackRep::indexedName(name, {type}),
+    {{"i1", type}, {"i2", type}}, type, body, {Attr::attr("inline")});
+}
+
+FuncDecl* Prelude::builtinBinaryOp(int width, std::string builtinName,
+                                   std::string name) {
+  std::string type = "i" + std::to_string(width);
+  return Decl::function(SmackRep::indexedName(name, {type}),
+    {{"i1", type}, {"i2", type}}, type, NULL,
+    {Attr::attr("builtin", builtinName)});
```
shaobo-he: `builtin` or `bvbuiltin` are like magic numbers, too. Make the creation of `builtin` attributes inside a function. Imagine you have to change these attributes, e.g., per changes of Boogie spec. Then you only need to change the names in these functions rather than replace every string in their occurrences.

shaobo-he:
Along with my comments above, I think it'd be better if we can organize these functions more properly. For example, can we group the inline function declarations together as they all have inline attributes? But I'm not sure it's worth the efforts or how to implement it.

zvonimir:
Looking at these changes again, I have another request for @liammachado. It would be good to somehow document what each function in `Prelude.cpp` generates. I suggest you add comments to each function showing an example axiom/function/string that is generates. That would make it much easier to map them to Boogie source during debugging.

liammachado:
I've made changes that address both Shaobo's and Zvonimir's recent comments. Are there any more suggestions?

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,526 @@
+#include "smack/Prelude.h"
+#include "smack/Naming.h"
+#include "smack/SmackOptions.h"
+#include "smack/Regions.h"
+
+namespace smack {
+
+const std::vector<unsigned> INTEGER_SIZES = {1, 5, 6, 8, 16, 24, 32, 40, 48,
```
shaobo-he: Is there a particular reason that `INTEGER_SIZES` is a global variable? It appears to me that it's only used in one function.

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,526 @@
...
+  56, 64, 80, 88, 96, 128, 160, 256};
+
+const std::vector<unsigned> REF_CONSTANTS = {
```
shaobo-he: See the comment above.

include/smack/Prelude.h
```diff
@@ -0,0 +1,61 @@
...
+  FuncDecl* inlineBvbuiltinBinarySelect(int width, std::string name, std::string name2);
+  FuncDecl* inlineBvbuiltinBinaryPred(int width, std::string name);
+  FuncDecl* inlineConversion(int width, int width2, std::string name, const Expr* body);
+  FuncDecl* truncBvOp(int width, int width2);
+  FuncDecl* truncOp(int width, int width2);
+
+  std::string getIntegerTypeName(int width);
+  std::string getBvTypeName(int width);
+
+  std::string makeVarName(int i);
```
shaobo-he: Do these functions need to be members of `Prelude` class? Are they used by other classes?

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,526 @@
...
+    std::string offset;
+    if (rep.ptrSizeInBits == 32)
+      offset = "4294967296";
+    else if (rep.ptrSizeInBits == 64)
+      offset = "18446744073709551616";
+    else
+      llvm_unreachable("Unexpected pointer bit width.");
+    s << Decl::function(rep.indexedName("$bv2int",{rep.ptrSizeInBits}),
+      {{"i",bt}}, it, Expr::cond(
+        Expr::fn(rep.indexedName("$slt", {bt, "bool"}),
```
shaobo-he: This is my fault, `"bool"` should be `Naming::BOOL_TYPE`.

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,526 @@
...
+}
+
+//e.g.: function {:inline} $min.bv1(i1: bv1, i2: bv1) returns (bv1)
+//        { if $slt.bv1.bool(i1, i2) then i1 else i2 }
+FuncDecl* Prelude::inlineBvbuiltinBinarySelect(int width, std::string name,
+                                               std::string name2) {
+  std::string type = getBvTypeName(width);
+
+  const Expr* body = Expr::if_then_else(
+    Expr::fn(name2 + "." + type + ".bool", makeVarExprs(2)),
```
shaobo-he: Can this line be expressed using `indexedName`?

shaobo-he:
The RP has improved a lot. Now my major concern is that we should avoid using string literals in the `Prelude` class. In other words, we should try to use variables defined in the `Naming` class.  For example, we have,
```C++
const std::vector<std::string> BINARY_OPS {"add", "sub", "mul", "sdiv",
  "smod", "srem", "udiv", "urem", "shl", "lshr", "ashr", "and", "or", "xor",
  "nand"};
```
in `Prelude.cpp`, these string literals are more or less duplicated in the `Naming` class (https://github.com/smackers/smack/blob/b6d0ea75d8b879998e3fa1f9ecd4b67f5f32602e/lib/smack/Naming.cpp#L100-L112). Ideally, we want them in one place such that when we want to change them, we don't have to search their every occurrences.

Does it make sense?

liammachado:
There are some string literals that I don't think can be avoided. For example, I don't think LLVM has an equivalent instruction for `"smod"`; in other words, I don't think that `Instruction::SMod` exists.

liammachado:
Are you suggesting we make separate variables to store the strings corresponding to each operation? For example:
```
const std::string TRUNC_STR = "$trunc";
const std::string ZEXT_STR = "$zext";
const std::string SEXT_STR = "$sext";
.
.
.
```
And then we re-write [this declaration](https://github.com/smackers/smack/blob/b6d0ea75d8b879998e3fa1f9ecd4b67f5f32602e/lib/smack/Naming.cpp#L87) to be:
```
const std::map<unsigned,std::string> Naming::INSTRUCTION_TABLE {
  {Instruction::Trunc, TRUNC_STR},
  {Instruction::ZExt, ZEXT_STR},
  {Instruction::SExt, SEXT_STR},
  .
  .
  .
```

shaobo-he:
> Are you suggesting we make separate variables to store the strings corresponding to each operation? For example:
> 
> ```
> const std::string TRUNC_STR = "$trunc";
> const std::string ZEXT_STR = "$zext";
> const std::string SEXT_STR = "$sext";
> .
> .
> .
> ```
> And then we re-write [this declaration](https://github.com/smackers/smack/blob/b6d0ea75d8b879998e3fa1f9ecd4b67f5f32602e/lib/smack/Naming.cpp#L87) to be:
> 
> ```
> const std::map<unsigned,std::string> Naming::INSTRUCTION_TABLE {
>   {Instruction::Trunc, TRUNC_STR},
>   {Instruction::ZExt, ZEXT_STR},
>   {Instruction::SExt, SEXT_STR},
>   .
>   .
>   .
> ```

Yes, though I'd suggest to name them like `TRUNC_FUNC`. @zvonimir @michael-emmi any suggestions?

lib/smack/Prelude.cpp
zvonimir: Good catch @shaobo-he. Yes, these should all be changed into `nullptr`, which will make the code more aligned with LLVM style and C++11 (https://stackoverflow.com/questions/20509734/null-vs-nullptr-why-was-it-replaced).

zvonimir:
Regarding @shaobo-he's suggestion to move all names into global constants in the `Naming` class, I am of the opinion that we should probably not do that unless a particular name is shared by multiple classes or methods. If a name is for example used as a string in just one method, then there is probably no need to replace it with a global constant. If a name gets repeated as a string in several places (across classes, methods), then it would probably be a good idea to replace it with a string constant.
Just my 2 cents.

include/smack/Prelude.h
zvonimir: @shaobo-he is I think right here. Since these are "free" helper functions (they don't access members of `Prelude`), it would be better to just put their implementations in the cpp file, without having them be private members of the `Prelude` class. See: https://stackoverflow.com/questions/1145482/should-c-class-helper-functions-be-members-free-or-anon-namespace-free

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,534 @@
...
+
+std::string getBvTypeName(int width) {
+  return "bv" + std::to_string(width);
+}
+
+std::string makeVarName(int i) {
+  return "i" + std::to_string(i);
+}
+
+std::vector<std::string> makeVarNames(int numVarNames) {
```
shaobo-he: This function seems unnecessary because first it's only used in `makeVars` and `makeVarExprs` and these functions can be easily expressed using only `makeVarName` and `makeVarExpr`.

Moreover, it also appears to me these two functions can be generalized to something like this,
```C++
template <class T>
std::list<T> makeListOfElems(unsigned num, T (*makeElem)(int)) {
  std::list<T> ret;
  for (unsigned i = 1; i <= num; ++i)
     ret.emplace_back(makeElem(i));
  return ret;
} 
```

lib/smack/Prelude.cpp
```diff
@@ -0,0 +1,534 @@
...
+      {makeInlineAttr()});
+  } else
+    s << Decl::function(rep.indexedName("$bv2int",{rep.ptrSizeInBits}),
+      {{"i",bt}}, it, nullptr, {makeBuiltinAttr("bv2int")})
+      << "\n";
+  s << "\n";
+
+  if (SmackOptions::BitPrecise) {
+    // XXX TODO don't assume 64-bit pointers TODO XXX
+    s << "// Bytewise pointer storage" << "\n";
```
shaobo-he: We may want these strings to be generated from functions.

lib/smack/Prelude.cpp
liammachado: The only problem with what you propose is that `makeVars` additionally takes a `std::string type` variable. Because of this, I don't see how we can use a single argument function pointer to get the behavior of `makeVars`. So instead of using a function pointer to represent the `makeElem` function, I used another template variable. See [this commit](https://github.com/smackers/smack/pull/410/commits/e06330ca7144f4d3a216e9e037db36669ee89b13) for details.

lib/smack/Prelude.cpp
liammachado: What would be an appropriate name for such a function?

lib/smack/Prelude.cpp
shaobo-he: If you pass the closure you created to `makeListOfElems`, it should work without using the additional `U`.

lib/smack/Prelude.cpp
shaobo-he: Sorry, I confused you. I meant the function strings below this line.

lib/smack/Prelude.cpp
liammachado: According to [this answer](https://stackoverflow.com/questions/28746744/passing-capturing-lambda-as-function-pointer/), capturing lambdas can't be converted to function pointers.

lib/smack/Prelude.cpp
liammachado: Thanks, I've just made this change. Let me know if it's okay.

lib/smack/Prelude.cpp
shaobo-he: I didn't know that. Thank you for sharing this link.

include/smack/Prelude.h
```diff
@@ -0,0 +1,46 @@
...
+#include "smack/SmackRep.h"
+
+#include <string>
+
+namespace smack {
+
+class Prelude {
+  SmackRep &rep;
+
+  FuncDecl* uninterpretedUnaryOp(unsigned width, std::string name);
```
michael-emmi: It is not clear why any of these are member functions, since they do not access members of the Prelude class — i.e., `rep`. Shouldn’t they just be regular (non-member) functions?

include/smack/SmackRep.h
```diff
@@ -96,6 +97,9 @@ class SmackRep {
 
   bool isUnsafeFloatAccess(const llvm::Type* elemTy, const llvm::Type* resultTy);
 
+  static std::string indexedName(std::string name, std::initializer_list<std::string> idxs);
```
michael-emmi: It is not clear why any of these are member functions, since they do not access members of the SmackRep class. Shouldn’t they just be regular (non-member) functions?

liammachado: These functions are used by both SmackRep.cpp and Prelude.cpp. I can duplicate them in both files, but this requires placing the ones in Prelude.cpp outside of the smack namespace. This is what I've done in [this commit](https://github.com/smackers/smack/pull/410/commits/004fd28b2afab765d6c2307f35277a33e8922ea7).

michael-emmi: Why does this require duplication, or moving outside of the smack namespace? Why can’t SmackRep.h simply declare these functions in the smack namespace, let SmackRep.cpp define them, and then have Prelude.cpp refer to the functions declared in SmackRep.h?

Maybe I am missing something subtle about namespaces?

liammachado: No, it was my mistake. I just made [another commit](https://github.com/smackers/smack/pull/410/commits/0399e46b539414fe7778f67adfafd52c0a52b4d0) that implements your suggestion.

michael-emmi: Nice! I’ll add just a couple more comments.

include/smack/Prelude.h
```diff
@@ -0,0 +1,25 @@
+//
+// This file is distributed under the MIT License. See LICENSE for details.
+//
+
+#ifndef PRELUDE_H
+#define PRELUDE_H
+
+#include "smack/BoogieAst.h"
```
michael-emmi: Looks like we can avoid this include here, and include in Prelude.cpp instead, since this header does not use BoogieAst directly. Otherwise modules which include the Prelude header would have to be recompiled every time you changed the BoogieAst header.

zvonimir:
Closes #308

Issue #413, opened 2019-02-13
Dealing with C promoting floats into doubles in __SMACK_code


zvonimir:
So according to the C standard, floats are automatically promoted into doubles when passed into vararg functions. Here is a discussion covering this:
https://stackoverflow.com/questions/28097564/why-does-printf-promote-a-float-to-a-double

This almost certainly creates performance issues for us since whenever we invoke `__SMACK_code` and pass in floats, they will get cast to doubles by clang beforehand, and reasoning about doubles is painfully slow in SMT solvers. So we should think about how to deal with this problem.

I came up with two solutions with different pros/cons:

1. We add several special `__SMACK_code_floatN` functions, where `N` determines the number of float arguments each function takes. That is how we avoid vararg in the first place. In `llvm2bpl` these functions would have to be handled separately, but in a way almost identical to the regular `__SMACK_code`. The issue with this proposal is that we are introducing these special functions.

2. We add a piece of code in the processing of `__SMACK_code` that determines for each argument if a cast from float to double is happening just before the invocation. If it is, then we remove the cast. The issue with this proposal is that we don't really know if a user actually intentionally introduced such a cast, which might happen in some situations, and we would remove all such casts. Also, this might be a bit complicated/awkward to implement since for each cast we would have to look ahead to see if the value is being passed to `__SMACK_code`, in which case we would omit generating bpl code for it. Maybe it would be better to remove such instructions in LLVM IR instead. Not sure.

Thoughts? Other proposals?

shaobo-he:
We add a new function like `__SMACK_type_cast()` to signal if we want to use the original type or promoted type. For example, `__SMACK_code(__SMACK_type_cast(a))` means the original type of `a` is used instead of promoted type of `a`.

Or we can add substrings to `__SMACK_code` as hints that we shouldn't use the promoted types of arguments. For example, `__SMACK_code("^1,2,3^ blahblah")`.

keram88:
I benchmarked this, and these conversions are a big problem. Simple single precision programs like this:
```
float f = __VERIFIER_nondet_float();
assume(!__isnan(f));
float g = fabsf(f);
assert(g >= 0);
```
take about three times longer with the conversions than without.

Currently, we are producing code that looks like this:
```
procedure {:inline 1} fabsf($f0: bvfloat) returns ($r: bvfloat)
{
$bb0:
  $f2 := ftd($f0);
  $f4: = dtf($f2);
  $f3 := $abs.bvfloat($f4);
[...]

PR #435, opened 2019-05-04
add type annotation to __SMACK_code

shaobo-he:
This PR implements the proposal in issue #413.

TODO:
- [x] update doc
- [x] add regressions
- [x] change math.c (optional)


keram88:
One thing I realized shortly after we wrote this is that we should check that the source type of the casted argument matches the "format" type.

Also, maybe we should warn if an uncaught casted argument flows into __SMACK_code so we can nip these possibly expensive casts early.

docs/boogie-code.md
```diff
@@ -38,6 +38,28 @@ void assume(bool v) {
...
+original types should be `i8`, `i16`, `i32`, and `float`. For example, without using
+such annotations, the definition of the `floorf` function is:
+```C
+float floorf(float x) {
+  double ret = __VERIFIER_nondet_double();
+  __SMACK_code("@ := ftd($rmode, $round.bvfloat(RTN, dtf($rmode, @)));", ret, x);
+   return ret;
+ }
+```
+, where the `ftd` and `dtf` functions are conversions between `float` and `double` variables.
```
zvonimir: Remove this comma in the beginning of this line.

zvonimir: Also "where the `ftd` and `dtf` functions are conversions between `float` and `double` variables that are needed to deal with the promotion of the arguments."

lib/smack/SmackRep.cpp
```diff
@@ -1012,13 +1017,47 @@ std::string SmackRep::code(llvm::CallInst& ci) {
 
   std::string s = fmt;
   for (unsigned i=1; i<ci.getNumOperands()-1; i++) {
-    const Expr* a = arg(f, i, ci.getOperand(i));
+    Value* a = ci.getOperand(i);
     std::string::size_type idx = s.find('@');
     assert(idx != std::string::npos && "inline code: too many arguments.");
 
+    llvm::Type* target_type = a->getType();
```
zvonimir: The names of variables should be camel case. Fix `target_type`.

lib/smack/SmackRep.cpp
```diff
@@ -1012,13 +1017,47 @@ std::string SmackRep::code(llvm::CallInst& ci) {
 
   std::string s = fmt;
   for (unsigned i=1; i<ci.getNumOperands()-1; i++) {
-    const Expr* a = arg(f, i, ci.getOperand(i));
+    Value* a = ci.getOperand(i);
```
zvonimir: I guess `a` here represents an argument. Rename this into `arg`.

lib/smack/SmackRep.cpp
```diff
@@ -1012,13 +1017,47 @@ std::string SmackRep::code(llvm::CallInst& ci) {
...
     std::string::size_type idx = s.find('@');
     assert(idx != std::string::npos && "inline code: too many arguments.");
 
+    llvm::Type* target_type = a->getType();
+    bool is_cast = false;
```
zvonimir: The names of variables should be camel case.

lib/smack/SmackRep.cpp
```diff
@@ -1012,13 +1017,47 @@ std::string SmackRep::code(llvm::CallInst& ci) {
...
+	target_type = Type::getInt32Ty(a->getContext());
+	is_cast = true;
+	break;
+      }
+      default:
+	break;
+      }
+    }
+    if (a->getType() != target_type) {
+	assert(isa<CastInst>(a) && "Expected a cast expression.");
```
zvonimir: Wrong indentation.

lib/smack/SmackRep.cpp
```diff
@@ -1012,13 +1017,47 @@ std::string SmackRep::code(llvm::CallInst& ci) {
...
-    const Expr* a = arg(f, i, ci.getOperand(i));
+    Value* a = ci.getOperand(i);
     std::string::size_type idx = s.find('@');
     assert(idx != std::string::npos && "inline code: too many arguments.");
 
+    llvm::Type* target_type = a->getType();
+    bool is_cast = false;
+    if (idx + 1 < s.length()) {
+      switch(s[idx+1]) {
+      case 'c': {
```
zvonimir: LLVM code style does not prescribe opening and closing of curly braces for case statements. Remove those.

lib/smack/SmackRep.cpp
```diff
@@ -1012,13 +1017,47 @@ std::string SmackRep::code(llvm::CallInst& ci) {
...
+	target_type = Type::getInt16Ty(a->getContext());
+	is_cast = true;
+	break;
+      }
+      case 'i': {
+	target_type = Type::getInt32Ty(a->getContext());
+	is_cast = true;
+	break;
+      }
+      default:
```
zvonimir: I think we should have some light-weight error checking here. I think your code would currently silently accept something like `@x`, is that right? Maybe you could add some error checking code here that would throw an assertion violation if a wrong character is used.

test/basic/smack_code_annot.c
```diff
@@ -0,0 +1,43 @@
...
+
+// @expect verified
+// @flag --bit-precise --float
+
+typedef unsigned int u32;
+typedef unsigned short u16;
+typedef unsigned char u8;
+
+u32 u16_to_u32(u16 x)
+{
```
zvonimir: Wrong formatting of curly braces. Take a look at other regressions.

test/basic/smack_code_annot.c
```diff
@@ -0,0 +1,43 @@
+#include "smack.h"
+
+// @expect verified
+// @flag --bit-precise --float
```
zvonimir: I don't think we should be adding regressions that require `--float` flag into `basic` folder. I guess we should add those into `float` folder. The same goes for `--bit-precise`. Maybe a good idea would be to split the benchmark into the int and float parts, and put the into one into bit vectors folder, and the float one into the float folder. Or keep it as one regression and put it into float folder.

lib/smack/SmackRep.cpp
shaobo-he: I changed it into `argV` because there's a function called `arg`.

lib/smack/SmackRep.cpp
shaobo-he: This behavior is intended. `@x` is interpreted as `<arg-expr>x`.

test/basic/smack_code_annot.c
shaobo-he: The reason that these flags are added is because we don't distinguish different integer types or float/double types without these flags. Namely, in order to test this feature, we need these flags to perform meaningful type checking.  So I think that it's fine we add these flags to benchmarks in the `basic` folder as long as these flags are not used to test the features they enable.

lib/smack/SmackRep.cpp
keram88: I can't find any instances of `@[a-zA-Z]` in the existing code, but maybe it's useful.
Right now we allow a weird situation where `@x` becomes `<arg-expr>x` but `@f` is just `<precast-arg-expr>`. Perhaps a consistent solution would be to check in `default` if the second character is in `[a-zA-Z]` and if it's not on the list from https://docs.python.org/2.7/library/struct.html#format-characters we report an error. If we want the literal character there, we can escape it with something like `\`, so `@\x` is `<arg-expr>x`.

lib/smack/SmackRep.cpp
```diff
@@ -1012,13 +1017,44 @@ std::string SmackRep::code(llvm::CallInst& ci) {
...
-    const Expr* a = arg(f, i, ci.getOperand(i));
+    Value* argV = ci.getOperand(i);
     std::string::size_type idx = s.find('@');
     assert(idx != std::string::npos && "inline code: too many arguments.");
 
+    llvm::Type* targetType = argV->getType();
+    bool isCast = false;
+    if (idx + 1 < s.length()) {
+      switch(s[idx+1]) {
+        case 'c':
```
keram88: Does anyone think it would be bad to write this as:
```
case 'c':
case 'b':
case 'B':
```
LLVM doesn't distinguish between these types, but it might match up with the type in the surrounding C code.

lib/smack/SmackRep.cpp
shaobo-he: I think we don't want to make the design complicated even if it means a more complete design. For our purposes, the current characters are sufficient.

zvonimir: Ok, I get it. I am ok that it stays as is.

test/basic/smack_code_annot.c
zvonimir: We should not have any benchmarks in the `basic` folder that require the theory of bitvectors or floating-points. When you add `--float` and `--bit-precise` that implies you need both of these theories. No?

docs/boogie-code.md
```diff
@@ -38,6 +38,28 @@ void assume(bool v) {
...
+original types should be `i8`, `i16`, `i32`, and `float`. For example, without using
+such annotations, the definition of the `floorf` function is:
+```C
+float floorf(float x) {
+  double ret = __VERIFIER_nondet_double();
+  __SMACK_code("@ := ftd($rmode, $round.bvfloat(RTN, dtf($rmode, @)));", ret, x);
+   return ret;
+ }
+```
+where the `ftd` and `dtf` functions are conversions between `float` and `double` variables that are needed to deal with the promotion of the arguments.
```
zvonimir: This line is too long. Keep them under 80 characters.

lib/smack/SmackRep.cpp
```diff
@@ -1002,6 +1002,11 @@ const Stmt* SmackRep::call(llvm::Function* f, const llvm::User& ci) {
   return Stmt::call(procName(f, ci), args, rets);
 }
 
+// we use C-style format characters
+// (https://docs.python.org/2.7/library/struct.html#format-characters)
+// e.g., @f means the variable is a float
+// while @h means the variable is a short
+// absence of a format character means use the promoted type as it is
```
zvonimir: Replace `as it is` with `as is`

lib/smack/SmackRep.cpp
zvonimir: This is a good point. @shaobo-he any thoughts on this?

lib/smack/SmackRep.cpp
```diff
@@ -1012,13 +1017,44 @@ std::string SmackRep::code(llvm::CallInst& ci) {
...
   for (unsigned i=1; i<ci.getNumOperands()-1; i++) {
-    const Expr* a = arg(f, i, ci.getOperand(i));
+    Value* argV = ci.getOperand(i);
     std::string::size_type idx = s.find('@');
     assert(idx != std::string::npos && "inline code: too many arguments.");
 
+    llvm::Type* targetType = argV->getType();
+    bool isCast = false;
+    if (idx + 1 < s.length()) {
+      switch(s[idx+1]) {
```
zvonimir: Wrong formatting again for switch-case - indentation is wrong. See here for how to format it according to LLVM coding style that we also try to follow: https://github.com/llvm-mirror/llvm/blob/master/lib/IR/ConstantRange.cpp#L83

test/basic/smack_code_annot.c
```diff
@@ -0,0 +1,40 @@
...
+  __SMACK_code("@ := ftd(RNE, @f);", ret, x);
+  return ret;
+}
+
+int main(void) {
+  u8 x8 = __VERIFIER_nondet_unsigned_char();
+  u8 y8 = __VERIFIER_nondet_unsigned_char();
+  u16 x16 = __VERIFIER_nondet_unsigned_short();
+  u16 y16 = __VERIFIER_nondet_unsigned_short();
+  __VERIFIER_assume(x8 > y8);
```
zvonimir: Why not use just `assume`?

zvonimir:
@liammachado : Could you take care of updating math.c based on the new `__SMACK_code` syntax?

liammachado:
Yes, I'll take care of that.

test/basic/smack_code_annot.c
shaobo-he: I split the tests into two copies for each folder.

test/basic/smack_code_annot.c
shaobo-he: It's the same thing, right? I can change it to `assume`.

docs/boogie-code.md
```diff
@@ -38,6 +38,29 @@ void assume(bool v) {
 }
 ````
 SMACK interprets the `@` symbol by inserting the value of the C variable `v`.
+Since the arguments of variadic functions may be promoted, SMACK allows users
+to append a format character to the `@` symbol as an indicator that the argument,
+rather than its promoted form, should be used in the format string. There are four
+format characters: `'c'`, `'h'`, `'i'`, and `'f'`, which represent that the
```
zvonimir: This should be updated to match the changes @keram88 made below to the main switch-case.

test/basic/smack_code_annot.c
zvonimir: Semantically it's the same thing. Syntactically, it makes sense to use `assume` if you are using `assert`, or `__VERIFIER_assume` if you are also using `__VERIFIER_assert`. I think/hope that in our regressions we are using just `assume` and `assert` and so it would be good to update this as well.

lib/smack/SmackRep.cpp
```diff
@@ -1012,13 +1017,48 @@ std::string SmackRep::code(llvm::CallInst& ci) {
...
+      case 'c':
+      case 'b':
+      case 'B':
+        targetType = Type::getInt8Ty(argV->getContext());
+        isCast = true;
+        break;
+      case 'f':
```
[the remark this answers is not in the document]
keram88: It seems the rational comment is gone, but this is the gist of it.
The cases are meant to match up with the format characters in https://docs.python.org/2.7/library/struct.html#format-characters. Lowercase and capitals are unsigned and signed respectively. Even though LLVM doesn't differentiate between signed-ness, it might be nice to have the type in the format string match the type in the code.
I'll update the doc to reflect this.

docs/boogie-code.md
keram88: Done.

docs/boogie-code.md
```diff
@@ -38,6 +38,42 @@ void assume(bool v) {
...
 SMACK interprets the `@` symbol by inserting the value of the C variable `v`.
+Since the arguments of variadic functions may be promoted, SMACK allows users
+to append a format character to the `@` symbol as an indicator that the argument,
+rather than its promoted form, should be used in the format string. The format
+characters are given in the following table along with their corresponding C
+types and their effective types within SMACK:
+
+| Format | Original C Type | SMACK Type |
+|--------|-----------------|------------|
+| c      | char            | i8         |
```
keram88: I can also see this being written like:
- `'c'/'b'/'B'` correspond to the char/signed char/unsigned char types respectively and correspond to the i8 type.
- ...

I don't have a real preference between this and the table.

zvonimir: I don't think we should be exposing SMACK type here. For example, if you enable `--bit-precise`, it would be bv8 instead of i8. But we should have information regarding which character matches which type. I think though we should remove the table and have that information be in a sentence, such as: `We allow the following format characters for the respective C types: `c` (char), `b` (signed char)...`

zvonimir: Btw, do these characters match the typical C printf format characters? It would be good if they would. So then you should also say "We allow the following format characters (inspired by C) for the respective C types:..."

keram88: I think Python based these characters on the `printf` format characters. Direct `printf` versions would look like: `@c` -> `char`, `@h(d?)` -> `short`, `@d` ->`int` (`int` doesn't have a length specifier apparently) and `@f` -> `float` (technically `double` according to `printf`). The differences in our format are using `@i` instead of `@d` for `int` and `short` being `@h` instead of `@hd`. I think "inspired by C" would be fine for this. 

I can change the doc to a sentence.

zvonimir: Well, then maybe say "inspired by Python" :)

zvonimir: Or even better "inspired by C and Python"

PR #437, opened 2019-05-07
Update floating-point constant syntax

liammachado:
Boogie's floating-point constant syntax was updated in [this pull request](https://github.com/boogie-org/boogie/pull/107). This pull request makes the generation of floating-point literals consistent with the new syntax.


lib/smack/SmackRep.cpp
```diff
@@ -659,9 +659,55 @@ const Expr* SmackRep::lit(const llvm::Value* v, bool isUnsigned) {
...
+      }
+
+      APInt bias = APInt::getSignedMaxValue(expSize);
+      if (exp == 0) {
+        --bias;
+      }
+
+      bool ov;
+
+      APInt moveDec = exp.usub_ov(bias, ov);
```
zvonimir: Do you need to check for overflow here? Can it ever happen? Would it be really bad if it happens?
I guess I would consider adding `assert(!overflow && "Overflow in floating-point computation");` here.
And at the other places where `overflow` gets used.

bin/versions
```diff
@@ -1,8 +1,8 @@
 MONO_VERSION=5.0.0.100
 Z3_SHORT_VERSION=4.8.1
 Z3_FULL_VERSION=4.8.1.016872a5e0f6
-BOOGIE_COMMIT=cd0609f660
-CORRAL_COMMIT=d7d389f22d
+BOOGIE_COMMIT=b694618b08
```
zvonimir: I just merged your Boogie pull request and so you should update Boogie hash again to the latest one.

lib/smack/SmackRep.cpp
```diff
@@ -659,9 +659,55 @@ const Expr* SmackRep::lit(const llvm::Value* v, bool isUnsigned) {
       }
       const APInt API = APF.bitcastToAPInt();
       const APInt n_sign = API.trunc(expSize+sigSize-1);
```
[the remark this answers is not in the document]
liammachado: I'm guessing it stands for "not sign", or something like that. It represents all of the bits except the sign bit.

lib/smack/SmackRep.cpp
liammachado: The only other way to subtract two APInts is to use the -= operator. This would require me to write:
```
APInt moveDec = exp;
moveDec -= bias;
```
I wanted to keep it all on one line, so I used `usub_ov`. There's another function called usub_sat that doesn't take the overflow argument, but for some reason it's not visible to SMACK.

liammachado: To answer your question, overflow happens whenever `exp < bias`, which is quite often, but it doesn't matter at all, because we want the value to wrap around anyways.

lib/smack/SmackRep.cpp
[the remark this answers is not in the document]
liammachado: I removed this variable since it's not really used anywhere.

lib/smack/BoogieAst.cpp
```diff
@@ -452,7 +456,12 @@ void BvLit::print(std::ostream& os) const {
 }
 
 void FPLit::print(std::ostream& os) const {
-  os << (neg ? "-" : "") << sig << "e" << expo << "f" << sigSize << "e" << expSize;
+  if (specialValue == "") {
```
zvonimir: It would probably be a bit better to use `specialValue.empty()` here.

lib/smack/SmackRep.cpp
```diff
@@ -658,10 +658,48 @@ const Expr* SmackRep::lit(const llvm::Value* v, bool isUnsigned) {
...
+      }
+
+      bool overflow;
+      APInt moveDec = exp.usub_ov(bias, overflow);
+      moveDec &= APInt(expSize, 3);
+
+      APInt finalExp = exp.usub_ov(bias, overflow).usub_ov(moveDec, overflow)
+        .ashr(2);
+
+      int leftSize = 4 * (*moveDec.getRawData() / 4 + 1);
```
zvonimir: Do `moveDec->getRawData()` here, instead of `*moveDec.getRawData()`.

zvonimir: In fact, since `moveDec->getRawData()` is used 3 times, introduce a local variable for it.

liammachado: `*moveDec.getRawData()` is equivalent to `*(moveDec.getRawData())`, not `(*moveDec).getRawData()`, which is what `moveDec->getRawData()` is equivalent to.

zvonimir: Got it, my mistake.

zvonimir:
Please squash the commits @liammachado, and write a nice commit message. I'll merge it as soon as that is done.

PR #438, opened 2019-05-08
Updated math.c to use new __SMACK_code syntax

liammachado:
This updates math.c to use type annotations with calls to __SMACK_code, which was added in [this pull request](https://github.com/smackers/smack/pull/435). This results in a simpler implementation for many functions.

### DIFF
--- a/share/smack/lib/math.c
+++ b/share/smack/lib/math.c
@@ -5,8 +5,8 @@
 #include <smack.h>
 
 float fabsf(float x) {
-  double ret = __VERIFIER_nondet_double();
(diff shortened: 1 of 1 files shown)


shaobo-he:
LGTM. In the meanwhile, I don't see any regression tests for the functions in this file. If you can find some time, it would be great to add them (probably in another PR).

liammachado:
Actually, we do already have regression tests for these functions, which are located [in this directory](https://github.com/smackers/smack/tree/develop/test/mathc).

shaobo-he:
@liammachado I see. Did you run them and see the results?

PR #439, opened 2019-05-09
Enable SMACK to build a no-std Rust crate

keram88:
As the name suggests, this feature gates the standard Vec class to enable building in no standard Rust programs.


shaobo-he:
LGTM.

PR #443, opened 2019-05-09
fix floating-point conversions

shaobo-he:
We don't need functions like `dtf` so I removed them. And floating-point
conversions from/to half-precision numbers are also implemented.

### DIFF
--- a/test/float/smack_code_annot.c
+++ b/test/float/smack_code_annot.c
@@ -2,14 +2,14 @@
 
 // @expect verified
 
-double f16_to_f32(float x) {
+double f32_to_f64(float x) {
   double ret = __VERIFIER_nondet_double();
-  __SMACK_code("@ := ftd(RNE, @f);", ret, x);
+  __SMACK_code("@ := $fpext.bvfloat.bvdouble(RNE, @f);", ret, x);
   return ret;
 }
 
 int main(void) {
-  assert(f16_to_f32(2.0f) == 2.0);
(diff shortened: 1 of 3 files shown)


zvonimir:
How come we don't need `dtf` any more?

zvonimir:
It seems there are failing regressions in the float folder.

shaobo-he:
> How come we don't need `dtf` any more?

They shouldn't exist in the first place. These functions should be `fpext` and `fptrunc` functions.

> It seems there are failing regressions in the float folder.

I fixed them.

test/float/smack_code_annot.c
```diff
@@ -4,7 +4,7 @@
 
 double f16_to_f32(float x) {
   double ret = __VERIFIER_nondet_double();
-  __SMACK_code("@ := ftd(RNE, @f);", ret, x);
+  __SMACK_code("@ := $fpext.bvfloat.bvdouble(RNE, @f);", ret, x);
```
zvonimir: Why do you have bvfloat into bvdouble if the function is `f16_to_f32`?
It seems to be that the types do not match the function name.

shaobo-he: Good catch. Fixed.

zvonimir:
@shaobo-he Please squash commits and I'll merge it then. Thx!

shaobo-he:
> @shaobo-he Please squash commits and I'll merge it then. Thx!

Done. Please merge it.

PR #444, opened 2019-05-14
Upgraded LLVM version to 4.0.1

zvonimir:
This pull request upgrades the used LLVM version to 4.0.1. I cherry-picked most of the changes from the pull request #389. The pull request had to be thoroughly tested and evaluated on SVCOMP benchmarks before merging.


zvonimir:
There is a failing regression. It seems it has something to do with DSA since memory maps are messed up. I've been studying the major updates that were made to DSA but could not figure out what the problem is.

zvonimir:
This is the minimum failing program I managed to create:
```
#include <pthread.h>
#include <smack.h>

pthread_mutex_t lock;
pthread_mutexattr_t lockattr;

void* t1(void *arg) {
  pthread_mutex_lock(&lock);
  return 0;
}

int main(void) {
  lock.attr.type = PTHREAD_MUTEX_ERRORCHECK;
  pthread_mutex_lock(&lock);
  pthread_mutexattr_init(&lockattr);
  pthread_mutex_init(&lock, 0);
  pthread_t t;
  pthread_create(&t, 0, t1, 0);
  return 0;
}
```
When you invoke it like this: `smack --pthread --context-bound=1 address_of_global.c`, you get a false bug. This is likely due to some weird DSA issue caused by the updates.
Any help you can provide to debug this would be greatly appreciated.

zvonimir:
Ok, my last commit fixed the issue above. Running SVCOMP again, and then hopefully we'll be ok to merge.

zvonimir:
@liammachado and @shaobo-he, I would need your help. This branch for some reason breaks 3 floating-point benchmarks, in the sense that it returns a wrong result.

See here: http://node0.smackbenchwebserver.smack.emulab.net/results.py?runset=runs%2Fexec_2019.05.16_18.40.08.528737_ReachSafety-Floats&runset=runs%2Fexec_2019.05.12_23.46.43.049129_ReachSafety-Floats

This is strange and I don't know why it is happening. I would appreciate if you could get to the bottom of this sooner rather than later so that we can merge this. Thanks!

zvonimir:
Ok, if you take a look at the last run I did on floats that I mentioned above, you can notice that these benchmarks are not failing. So this is after all something that has to do with the LLVM 4.0.1 update. Please see if you can root cause it.

shaobo-he:
@zvonimir 

It turns out Clang translates the function `__finite(d)` to `one(llvm.fabs.f64(d), inf)` where llvm.fabs.f64 is an LLVM intrinsic (https://llvm.org/docs/LangRef.html#llvm-fabs-intrinsic). It appears the previous version of Clang directly calls the `__finite` function instead.

zvonimir:
Thanks @shaobo-he for root causing this. Ok, my take on this is that we merge this, and address the intrinsics issue in a separate pull request. SVCOMP benchmarks look good (some slowdown, some speedup), apart from this issue with 3 floating-point benchmarks.

Issue #436, opened 2019-05-04
Handle LLVM Half Precision Floating-Point Intrinsics


shaobo-he:
Please see: https://llvm.org/docs/LangRef.html#half-precision-floating-point-intrinsics.

PR #447, opened 2019-05-14
Add support for the LLVM half-conversion intrinsics

keram88:
This adds support for the LLVM conversion intrinsics for converting: `f64`->`f16`,  `f32`->`f16`,  `f16`->`f64` and `f16`->`f32`.
Since the normal mode of LLVM is to use half precision as an IEEE binary16 storage format, this should be all we need for now. (BFloat16 hardware is a ways off apparently.)
Addresses #436.

### DIFF
--- a/lib/smack/SmackInstGenerator.cpp
+++ b/lib/smack/SmackInstGenerator.cpp
@@ -144,6 +144,42 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
   nextInst++;
 }
 
+void SmackInstGenerator::processIntrinsicCall(llvm::IntrinsicInst* ii) {
+  static const auto f16UpCast = [&ii, this] {
+    std::string funcName;
+    auto tyID = ii->getFunctionType()->getReturnType()->getTypeID();
+    if (tyID == llvm::Type::FloatTyID)
+      funcName = "__SMACK_convert_from_fp16_f32";
+    else if (tyID == llvm::Type::DoubleTyID)
+      funcName = "__SMACK_convert_from_fp16_f64";
+    else
+      llvm_unreachable("Unexpected return type of half-precision upcast intrinsic.");
+    return Stmt::call(funcName, {rep->expr(ii->getArgOperand(0))},
+      {naming->get(*ii)});
+  };
+  static const auto f16DownCast = [&ii, this] {
+    std::string funcName;
+    auto tyID = ii->getArgOperand(0)->getType()->getTypeID();
+    if (tyID == llvm::Type::FloatTyID)
+      funcName = "__SMACK_convert_to_fp16_f32";
+    else if (tyID == llvm::Type::DoubleTyID)
+      funcName = "__SMACK_convert_to_fp16_f64";
+    else
+      llvm_unreachable("Unexpected operand type of half-precision downcast intrinsic.");
+    return Stmt::call(funcName, {rep->expr(ii->getArgOperand(0))},
+      {naming->get(*ii)});
+  };
+  static const std::map<llvm::Intrinsic::ID, std::function<const Stmt*()>> stmtMap {
+    {llvm::Intrinsic::convert_from_fp16, f16UpCast},
+    {llvm::Intrinsic::convert_to_fp16, f16DownCast}
+  };
+  auto it = stmtMap.find(ii->getIntrinsicID());
+  if (it != stmtMap.end())
+    return emit(it->second());
+  else
+    return emit(rep->call(ii->getCalledFunction(), *ii));
+}
+
 void SmackInstGenerator::visitBasicBlock(llvm::BasicBlock& bb) {
   nextInst = bb.begin();
   currBlock = getBlock(&bb);
@@ -622,6 +658,9 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
     // Semantically, this function simply returns the value v.
     Value* val = ci.getArgOperand(0);
     emit(Stmt::assign(rep->expr(&ci), rep->expr(val)));
+
+  } else if (auto ii = dyn_cast<IntrinsicInst>(&ci)) {
+    processIntrinsicCall(ii);
 
   } else if (name.find(Naming::RUST_ENTRY) != std::string::npos) {
     // Set the entry point for Rust programs
--- a/share/smack/frontend.py
+++ b/share/smack/frontend.py
@@ -74,6 +74,7 @@ def default_clang_compile_command(args, lib = False):
   if args.memory_safety: cmd += ['-DMEMORY_SAFETY']
   if args.integer_overflow: cmd += (['-ftrapv', '-fsanitize=shift'] if not lib else ['-DSIGNED_INTEGER_OVERFLOW_CHECK'])
   if args.float: cmd += ['-DFLOAT_ENABLED']
+  if args.bit_precise: cmd += ['-DBIT_PRECISE']
   if sys.stdout.isatty(): cmd += ['-fcolor-diagnostics']
   return cmd
 
@@ -239,7 +240,7 @@ def rust_frontend(input_file, args):
 def default_build_libs(args):
   """Generate LLVM bitcodes for SMACK libraries."""
   bitcodes = []
-  libs = ['smack.c', 'stdlib.c', 'errno.c']
+  libs = ['smack.c', 'stdlib.c', 'errno.c', 'intrinsics.c']
 
   if args.pthread:
     libs += ['pthread.c']
--- a/share/smack/lib/intrinsics.c
+++ b/share/smack/lib/intrinsics.c
@@ -1,0 +1,43 @@
+#include "smack.h"
+
+#if FLOAT_ENABLED && BIT_PRECISE
+float __SMACK_convert_from_fp16_f32(unsigned short x) {
+  float ret = __VERIFIER_nondet_float();
+  __SMACK_code("@f := $fpext.bvhalf.bvfloat($rmode, $bitcast.bv16.bvhalf(@H));", ret, x);
+  return ret;
+}
+
+double __SMACK_convert_from_fp16_f64(unsigned short x) {
+  double ret = __VERIFIER_nondet_double();
+  __SMACK_code("@ := $fpext.bvhalf.bvdouble($rmode, $bitcast.bv16.bvhalf(@H));", ret, x);
+  return ret;
+}
+
+unsigned short __SMACK_convert_to_fp16_f32(float x) {
+  unsigned short ret = __VERIFIER_nondet_unsigned_short();
+  __SMACK_code("assume($bitcast.bv16.bvhalf(@H) == $fptrunc.bvfloat.bvhalf($rmode, @f));", ret, x);
+  return ret;
+}
+
+unsigned short __SMACK_convert_to_fp16_f64(double x) {
+  unsigned short ret = __VERIFIER_nondet_unsigned_short();
+  __SMACK_code("assume $bitcast.bv16.bvhalf(@H) == $fptrunc.bvdouble.bvhalf($rmode, @);", ret, x);
+  return ret;
+}
+#else
+float __SMACK_convert_from_fp16_f32(unsigned short x) {
+  return __VERIFIER_nondet_float();
+}
+
+double __SMACK_convert_from_fp16_f64(unsigned short x) {
+  return __VERIFIER_nondet_double();
+}
+
+unsigned short __SMACK_convert_to_fp16_f32(float x) {
+  return __VERIFIER_nondet_unsigned_short();
+}
+
+unsigned short __SMACK_convert_to_fp16_f64(double x) {
+  return __VERIFIER_nondet_unsigned_short();
+}
+#endif
--- a/include/smack/SmackInstGenerator.h
+++ b/include/smack/SmackInstGenerator.h
@@ -40,6 +40,7 @@ private:
   void generateGotoStmts(llvm::Instruction& i,
                          std::vector<std::pair<const Expr*, llvm::BasicBlock*> > target);
   void processInstruction(llvm::Instruction& i);
+  void processIntrinsicCall(llvm::IntrinsicInst* ii);
   void nameInstruction(llvm::Instruction& i);
   void annotate(llvm::Instruction& i, Block* b);
 
--- a/test/float/half_intrinsics.c
+++ b/test/float/half_intrinsics.c
@@ -1,0 +1,16 @@
+#include "smack.h"
+#include "math.h"
+
+// @flag --bit-precise
+// @expect verified
+
+int main(void) {
+  union {unsigned short u; __fp16 h;} uh;
+  uh.u = 1U;
+  assert(isnormal((float)uh.h));
+  assert(isnormal((double)uh.h));
+  uh.h = 66504.0f;
+  assert(uh.u == 31744U);
+  uh.h = 66504.0;
+  assert(uh.u == 31744U);
+}
--- a/test/float/half_intrinsics_fail.c
+++ b/test/float/half_intrinsics_fail.c
@@ -1,0 +1,16 @@
+#include "smack.h"
+#include "math.h"
+
+// @flag --bit-precise
+// @expect error
+
+int main(void) {
+  union {unsigned short u; __fp16 h;} uh;
+  uh.u = 1U;
+  assert(isnormal((float)uh.h));
+  assert(isnormal((double)uh.h));
+  uh.h = 66504.0f;
+  assert(uh.u != 31744U);
+  uh.h = 66504.0;
+  assert(uh.u == 31744U);
+}


lib/smack/SmackInstGenerator.cpp
```diff
@@ -623,6 +623,22 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
     Value* val = ci.getArgOperand(0);
     emit(Stmt::assign(rep->expr(&ci), rep->expr(val)));
 
+  } else if (name.find("llvm.convert.from.fp16.f32") != std::string::npos) {
```
shaobo-he: It seems that we can merge these branches like `if (a || b || c ...)` since they have the same body.

share/smack/lib/smack.c
```diff
@@ -257,6 +257,30 @@ void __SMACK_dummy(int v) {
...
+
+double __SMACK_f16_to_f64(unsigned short x) {
+  double ret = __VERIFIER_nondet_float();
+  __SMACK_code("@ := $fpext.bvhalf.bvdouble(RNE, $bitcast.bv16.bvhalf(@H));", ret, x);
+  return ret;
+}
+
+unsigned short __SMACK_f32_to_f16(float x) {
+  unsigned short ret = __VERIFIER_nondet_unsigned_short();
+  __SMACK_code("assume($bitcast.bv16.bvhalf(@H) == $fptrunc.bvfloat.bvhalf(RNE, @f));", ret, x);
```
shaobo-he: My fault, the rounding mode argument should be current system rounding mode.

keram88: Yes.

zvonimir: Yes, this should be the current rounding mode, and not RNE (I would guess).
Moreover, why are you using these auxiliary C functions, instead of emitting Boogie code directly from `SmackRep.cpp`? I don't see an advantage of adding these C functions.

zvonimir:
Could you add a simple regression that tests this code?

zvonimir:
I discussed this pull request with @keram88. My take is that intrinsics should ideally either all be uniformly supported in C++ directly (by generating Boogie code), or using `smack.c`. If at all possible of course. It seems to me that mixing and matching (some in C++, some in `smack.c`) is probably not idea.

shaobo-he:
> I discussed this pull request with @keram88. My take is that intrinsics should ideally either all be uniformly supported in C++ directly (by generating Boogie code), or using `smack.c`. If at all possible of course. It seems to me that mixing and matching (some in C++, some in `smack.c`) is probably not idea.

Yes, I agree. I think I prefer using `smack.c` because it's generally more expressive and convenient than writing Boogie code programmatically.

shaobo-he:
@keram88 @zvonimir I refactored the code based on our discussions. Please take a look at it.

share/smack/lib/half_intrinsics.c
```diff
@@ -0,0 +1,29 @@
+#include "smack.h"
+
+// Now this guard is unnecessary but if we decide to use this file
+// to implement all the intrinsics, we will need it.
+#if FLOAT_ENABLED
+float __SMACK_convert_from_fp16_f32(unsigned short x) {
+  float ret = __VERIFIER_nondet_float();
+  __SMACK_code("@f := $fpext.bvhalf.bvfloat(RNE, $bitcast.bv16.bvhalf(@H));", ret, x);
```
zvonimir: Should rounding mode be hard-coded here?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -623,6 +623,16 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
     Value* val = ci.getArgOperand(0);
     emit(Stmt::assign(rep->expr(&ci), rep->expr(val)));
 
+  } else if (SmackOptions::BitPrecise && SmackOptions::FloatEnabled &&
```
zvonimir: What code will get generated (if any) if BitPrecise and FloatEnabled are not turned on?
Should we not generate some kind of uninterpreted functions in that cases?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -623,6 +623,16 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
...
+    (name.find("llvm.convert.from.fp16.f32") != std::string::npos ||
+    name.find("llvm.convert.from.fp16.f64") != std::string::npos ||
+    name.find("llvm.convert.to.fp16.f32") != std::string::npos ||
+    name.find("llvm.convert.to.fp16.f64") != std::string::npos)) {
+    std::replace(name.begin(), name.end(), '.', '_');
```
zvonimir: This way of generating function names is quite fragile. Maybe we could do something better here, like a constant map mapping LLVM intrinsic names into SMACK functions/models. Or maybe create a function `getIntrinsicModelName` that takes LLVM name and generated SMACK name for any intrinsic. Maybe that would be better.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -623,6 +623,16 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
     Value* val = ci.getArgOperand(0);
     emit(Stmt::assign(rep->expr(&ci), rep->expr(val)));
 
+  } else if (SmackOptions::BitPrecise && SmackOptions::FloatEnabled &&
+    (name.find("llvm.convert.from.fp16.f32") != std::string::npos ||
+    name.find("llvm.convert.from.fp16.f64") != std::string::npos ||
```
zvonimir: Could we use `getIntrinsicID()` to do this matching, instead of using strings?

lib/smack/SmackInstGenerator.cpp
zvonimir: Let's pull out handing of intrinsics into a separate function, so that it is not mixed with e.g. handing of annotations. I suggest we refactor this by introducing two private functions `processIntrinsicCall` and `processContractCall` (or something like that).

zvonimir: I would like that all intrinsics are then handled using models from a C header. So you should refactor `llvm.expect` above to to the same. Like I said, either we should model all intrinsics in a C header, or we should model all intrinsics directly in C++ (meaning here). Now are are doing both, which I don't like since it is not uniform.

share/smack/lib/half_intrinsics.c
```diff
@@ -0,0 +1,29 @@
+#include "smack.h"
```
zvonimir: Let's call this file `intrinsics.c` and have all intrinsics modeled here.

share/smack/lib/half_intrinsics.c
shaobo-he: All the lower precision floating-point representations can be precisely represented by higher precision ones. The rounding mode doesn't matter since no rounding would occur.

lib/smack/SmackInstGenerator.cpp
shaobo-he: Calls to procedures without bodies will be generated if these flags are not turned on. I think our implementations only make sense if these flags are enabled because if they are not, `i16` is unbounded, which means functions like `i16->half` are undefined.

shaobo-he: I agree. However, since the scope of this PR is limited to half-conversion intrinsics, maybe we want another PR to implement a file that handles all the LLVM intrinsics?

share/smack/lib/half_intrinsics.c
keram88: While no rounding is involved in these upcasts, it shouldn't hurt to be consistent and use the current rounding mode for all conversions.

lib/smack/SmackInstGenerator.cpp
keram88: We'd still be kind of inconsistent with intrinsics since we're writing some of them directly in boogie, while the integer overflow intrinsics are written as LLVM transformations. I suppose this is okay since most overflow checks are handled fairly differently, except for the division and left shift overflow checks, which sort of amount to what we're doing here. We also have the special functions (rust panic, etc.) which we probably want to organize as well.

share/smack/lib/half_intrinsics.c
zvonimir: Yes, I think what @keram88 says here is a good idea. Let's use the current rounding mode everywhere.

lib/smack/SmackInstGenerator.cpp
zvonimir: Ok, I am fine with another pull request that will refactor this code as I suggested.
Can you do that right after this one?

keram88: Yes. That way we can see what the new structure looks like.

lib/smack/SmackInstGenerator.cpp
shaobo-he: It turns out `getIntrinsicID` is tricky to use. Apart from that, it seems this function returns an ID for overloaded intrinsics. For example, all the conversions to `fp16` has one ID. We have to resort to the function names.

zvonimir: I see. I have to say I find it hard to review this pull request independently of suggesting more general refactoring changes to how we handle intrinsics. So let's maybe push this one true, and follow it with refactoring right away (before I forget everything I wrote here about refactoring).
Does that sound good?

shaobo-he: Sounds good to me. I implemented a new version that uses `getIntrinsicID` instead of function names.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -144,6 +144,33 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
   nextInst++;
 }
 
+void SmackInstGenerator::processIntrinsicCall(llvm::IntrinsicInst* ii) {
+  static auto f16Cast = [&ii, this](bool f16UpCast) {
```
zvonimir: Oh, man, this is maybe overly complicated just to avoid using the names of the intrinsics. Sorry about that.
Let's actually use their names in combination with `getIntrinsicID`. So you first switch based on `getIntrinsicID` to get into the right top-level intrinsic type, and then you switch based on names. The list of names will be exhaustive for a particular intrinsic ID, and there will be an assertion in the code checking that a name gets matches. This assertion will catch any changes in the names of the intrinsics.

share/smack/lib/half_intrinsics.c
zvonimir: Please rename this into `intrinsics.c`

lib/smack/SmackInstGenerator.cpp
keram88: Despite (or because of) how hard this was to get type-checked... I kind of agree.

shaobo-he: I actually prefer not to use function names now as they might not be a reliable and are subject to version changes. Using function names is essentially more or less equivalent to what we're doing now, which is finding the precise intrinsic associated with the overloaded ID by using function signatures.

shaobo-he:
@zvonimir @keram88 I fixed the failing regressions. If this PR looks reasonable to you now, I will squeeze the commits and then we can merge it.

lib/smack/SmackInstGenerator.cpp
keram88: @zvonimir do you mean the complexity of using the lambda map compared to that of using a switch statement?
I think the IDs in the enum are very unlikely to be changed.

share/smack/lib/intrinsics.c
```diff
@@ -0,0 +1,29 @@
+#include "smack.h"
+
+// Now this guard is unnecessary but if we decide to use this file
+// to implement all the intrinsics, we will need it.
+#if FLOAT_ENABLED
```
zvonimir: In C++ code, this is guarded by both float and bit-precise being enabled. Here, it is guarded just by float. Why is there this discrepancy?

share/smack/frontend.py
```diff
@@ -250,6 +250,8 @@ def default_build_libs(args):
   if args.float:
     libs += ['math.c']
     libs += ['fenv.c']
+    if args.bit_precise:
+      libs += ['intrinsics.c']
```
zvonimir: Why are we not adding `intrinsics.c` to `libs` always?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -144,6 +144,37 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+        funcName = f16UpCast? "__SMACK_convert_from_fp16_f64":
+          "__SMACK_convert_to_fp16_f64";
+      return Stmt::call(funcName, {rep->expr(ii->getArgOperand(0))},
+            {naming->get(*ii)});
+    } else
+      return rep->call(ii->getCalledFunction(), *ii);
+  };
+  static std::map<llvm::Intrinsic::ID, std::function<const Stmt*()>> stmtMap {
+    {llvm::Intrinsic::convert_from_fp16,
+      []() { return f16Cast(true); }},
```
zvonimir: Instead of passing in true/false to signal up/down cast, and then branching on this in `f16Cast`, why wouldn't you instead split f16Cast into two function, one for up and one for down casting?
That would make the code cleaner since you would avoid branching on `f16UpCast` boolean.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -144,6 +144,37 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+    {llvm::Intrinsic::convert_to_fp16,
+      []() { return f16Cast(false); }}
+  };
+  auto it = stmtMap.find(ii->getIntrinsicID());
```
zvonimir: I don't understand the purpose of `stmtMap`. It is generated dynamically every time `processIntrinsicCall` is called. So how come you are putting the generated statements in this map, instead of emitting them directly?

share/smack/lib/intrinsics.c
shaobo-he: The short answer is that we don't have a macro associated with the flag `--bit-precise`.  Currently we generate Boogie floating-point functions used in these C functions even when `--bit-precise` is not enabled. So we're good for now.

share/smack/frontend.py
shaobo-he: I think we can do it in the coming PR to refactor LLVM intrinsic handling.

lib/smack/SmackInstGenerator.cpp
shaobo-he: Having two functions was our previous implementation then I realized that the two functions share a lot of code like `if (SmackOptions::...)`. I think the code is actually cleaner with this switch.

lib/smack/SmackInstGenerator.cpp
shaobo-he: Actually this map is not named properly. It should be something like `stmtGenMap` which means that it maps an intrinsic ID to a procedure that generates Boogie statements. Such a map is generated only once as the `static` keyword in its definition suggests. The Boogie statement is generated dynamically via application to the mapped procedure. I guess the purpose of this map is because one-to-one mapping from intrinsic IDs to generated statements is not possible since the intrinsic IDs are overloaded. However, the one-to-one mapping from intrinsic IDs to the statement generation procedure is possible.

lib/smack/SmackInstGenerator.cpp
zvonimir: Actually, what has also been bugging me is the fact that we do not explicitly specify models for these intrinsics for the cases where `--bit-precise` and `--floats` is not set. I find the fact that these models are implicitly generated confusing.
So I propose that we include those models as well in the `intrinsics.c` file, in which case `if (SmackOptions::...)` conditions would disappear.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -144,6 +144,38 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
 
+void SmackInstGenerator::processIntrinsicCall(llvm::IntrinsicInst* ii) {
+  static const auto f16UpCast = [&ii, this] {
+    std::string funcName;
+    auto tyID = ii->getFunctionType()->getReturnType()->getTypeID();
+    if (tyID == llvm::Type::FloatTyID)
+      funcName = "__SMACK_convert_from_fp16_f32";
+    else if (tyID == llvm::Type::DoubleTyID)
+      funcName = "__SMACK_convert_from_fp16_f64";
+    return Stmt::call(funcName, {rep->expr(ii->getArgOperand(0))},
```
zvonimir: What happens if neither `if` nor `else if` are executed?
I suspect we want to add here as a sanity check in future LLVM versions:
```
else
  llvm_unreachable("Unexpected return type of half-precision upcast intrinsic.");
```

lib/smack/SmackInstGenerator.cpp
```diff
@@ -144,6 +144,38 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+      {naming->get(*ii)});
+  };
+  static const auto f16DownCast = [&ii, this] {
+    std::string funcName;
+    auto tyID = ii->getArgOperand(0)->getType()->getTypeID();
+    if (tyID == llvm::Type::FloatTyID)
+      funcName = "__SMACK_convert_to_fp16_f32";
+    else if (tyID == llvm::Type::DoubleTyID)
+      funcName = "__SMACK_convert_to_fp16_f64";
+    return Stmt::call(funcName, {rep->expr(ii->getArgOperand(0))},
```
zvonimir: What happens if neither `if` nor `else if` are executed?
I suspect we want to add here as a sanity check in future LLVM versions:
```
else
  llvm_unreachable("Unexpected operand type of half-precision downcast intrinsic.");
```

share/smack/lib/intrinsics.c
```diff
@@ -0,0 +1,45 @@
+#include "smack.h"
+
+// Now this guard is unnecessary but if we decide to use this file
```
zvonimir: I think these comments do not hold any more, and should hence be removed.

lib/smack/SmackInstGenerator.cpp
shaobo-he: The else branch should not be executed because according to LLVM language manual, the resulting types should only be either float or double. I added the sanity check.

lib/smack/SmackInstGenerator.cpp
shaobo-he: Same as the previous comment.

zvonimir:
Please squash and I'll merge it then. Thanks!

shaobo-he:
> Please squash and I'll merge it then. Thanks!

Done.

PR #451, opened 2019-05-28
Refactor/improve LLVM intrinsic handling

shaobo-he:
- [x] Refactor existing intrinsic handling
- [x] Add FP intrinsic handling
- [x] Add BV intrinsic handling
- [ ] Add Fixed Point intrinsic handling
- [ ] Add regressions


keram88:
I'm not sure what the intended scope of this is. I would like to move the Rust functions out into something like `isSpecialFunction` and `processSpecialFunction` as well.

shaobo-he:
> I'm not sure what the intended scope of this is. I would like to move the Rust functions out into something like `isSpecialFunction` and `processSpecialFunction` as well.

I think they are out of the scope of this PR.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -652,13 +662,6 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst& ci) {
     WARN("ignoring llvm.debug call.");
     emit(Stmt::skip());
 
-  } else if (name.find("llvm.expect.") != std::string::npos) {
```
zvonimir: Move handling of debug intrinsics above into `processIntrinsicCall` as well.

zvonimir:
@keram88 : Please create a separate pull request for that. I guess the function should be called `processSpecialRustCall` since currently they are all Rust-specific.

lib/smack/SmackInstGenerator.cpp
shaobo-he: > Move handling of debug intrinsics above into `processIntrinsicCall` as well.

Good catch. Done.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -166,18 +166,35 @@ void SmackInstGenerator::processIntrinsicCall(llvm::IntrinsicInst* ii) {
...
+    {llvm::Intrinsic::expect, expect}
   };
+
   auto it = stmtMap.find(ii->getIntrinsicID());
   if (it != stmtMap.end())
-    return emit(it->second());
+    it->second();
   else
-    return emit(rep->call(ii->getCalledFunction(), *ii));
+    emit(rep->call(ii->getCalledFunction(), *ii));
```
zvonimir: Could we issue a warning here to indicate that we encountered an intrinsic we are not currently handling/modeling?

zvonimir:
Please squash and then I'll merge it in.

shaobo-he:
> Please squash and then I'll merge it in.

Shouldn't we also add the handling of intrinsics like `llvm.fabs` and `llvm.ctlz`?

zvonimir:
I thought you wanted to do that as a separate pull request, which maybe would not be a bad idea.

shaobo-he:
> I thought you wanted to do that as a separate pull request, which maybe would not be a bad idea.

I think we should implement them in this PR so that we have a better general picture about what's a good way to implement them.

keram88:
I thought it was separate too.
Anyway, this is my (three-quarters baked?) Prelude.cpp implementation for bswap, so we can see what that approach looks like.
https://github.com/smackers/smack/commit/6ae5c93d6b9a775700f9b2ae0088961792598648

zvonimir:
When you started working on the intrinsics, I said we should try to handle them uniformly, meaning either put all of them into `intrinsics.c` or put all of them into `Prelude.cpp`. Now it seems that some will go into `intrinsics.c` and some into `Prelude.cpp`. Why is that so? Why could they not all be at the same place?
Wouldn't you agree that having all intrinsics be modeled at the same place would make sense?

shaobo-he:
> When you started working on the intrinsics, I said we should try to handle them uniformly, meaning either put all of them into `intrinsics.c` or put all of them into `Prelude.cpp`. Now it seems that some will go into `intrinsics.c` and some into `Prelude.cpp`. Why is that so? Why could they not all be at the same place?
> Wouldn't you agree that having all intrinsics be modeled at the same place would make sense?
> 
> Looking at the bvswap implementation, to me it seems it would be hard to implement it nicely in `intrinsics.c` without introducing some crazy macros or something like that.
> 
> Hence, my proposal for refactoring is as follows:
> 
> 1. Move what is currently in `intrinsics.c` into `Prelude.cpp`. I would guess that should be pretty easy.
> 2. Delete `intrinsics.c`.
> 3. Implement a function called `generateIntrinsics` or something like that in `Prelude.cpp`. All models for intrinsics should be generated from that functions. Be careful about `float` and `bit-precise` flags.

By talking about handling LLVM intrinsics uniformly, I thought we meant to choose one between the two design choices,

1. Replace the function call to an LLVM intrinsic with a call to the intrinsic handling function which is what we implemented.
2. Replace the function call to an LLVM intrinsic with Boogie commands directly. For example, a call to `llvm.expect` is translated to an assignment.

I voted for the first option because I thought models of LLVM intriniscs can be much more easily expressed using C. For example, there are a lot of resources online about how to implement counting the leading zeros in loop-free C code. But Mark figures out ways to implement these models using Boogie functions, which is more efficient than C code. So the second option is more promising because we do not need `intrinsics.c`.

The bottom line is that there is nothing in `intrinsics.c` that could be moved to `Prelude.cpp` because the former only contains functions using `__SMACK_code`. Its existence is due to our adoption of option 1.

shaobo-he:
In spite of our ongoing discussion, I squashed the commits so it's ready to be merged.

zvonimir:
> > When you started working on the intrinsics, I said we should try to handle them uniformly, meaning either put all of them into `intrinsics.c` or put all of them into `Prelude.cpp`. Now it seems that some will go into `intrinsics.c` and some into `Prelude.cpp`. Why is that so? Why could they not all be at the same place?
> > Wouldn't you agree that having all intrinsics be modeled at the same place would make sense?
> > Looking at the bvswap implementation, to me it seems it would be hard to implement it nicely in `intrinsics.c` without introducing some crazy macros or something like that.
> > Hence, my proposal for refactoring is as follows:
> > 
> > 1. Move what is currently in `intrinsics.c` into `Prelude.cpp`. I would guess that should be pretty easy.
> > 2. Delete `intrinsics.c`.
> > 3. Implement a function called `generateIntrinsics` or something like that in `Prelude.cpp`. All models for intrinsics should be generated from that functions. Be careful about `float` and `bit-precise` flags.
> 
> By talking about handling LLVM intrinsics uniformly, I thought we meant to choose one between the two design choices,
> 
> 1. Replace the function call to an LLVM intrinsic with a call to the intrinsic handling function which is what we implemented.
> 2. Replace the function call to an LLVM intrinsic with Boogie commands directly. For example, a call to `llvm.expect` is translated to an assignment.
> 
> I voted for the first option because I thought models of LLVM intriniscs can be much more easily expressed using C. For example, there are a lot of resources online about how to implement counting the leading zeros in loop-free C code. But Mark figures out ways to implement these models using Boogie functions, which is more efficient than C code. So the second option is more promising because we do not need `intrinsics.c`, although it's unclear that we can do this to all the intriniscs.
> 
> The bottom line is that there is nothing in `intrinsics.c` that could be moved to `Prelude.cpp` because the former only contains functions using `__SMACK_code`. Its existence is due to our adoption of option 1.

Well, note that your option (1) can be implemented in the following two ways:
1. Implement models for intrinsics as C functions in `intrinsics.c`, which is what we did.
2. Implement models for intrinsics as Boogie functions in `Prelude.cpp`, which is what Mark did.

So I think our options for implementing models for intrinsics are as follows:
1. Generate Boogie code that models an intrinsic directly and inline it. In this case there would be no functions or function calls being generated.
2. Implement models for intrinsics as C functions in `intrinsics.c`, and then invoke the appropriate function every time an intrinsic is encountered.
3. Implement models for intrinsics as Boogie functions/procedures in `Prelude.cpp`, and then invoke the appropriate function every time an intrinsic is encountered.
4. Some combination of the above depending on an intrinsic.

My main goal has been uniformity, meaning to avoid (4).

I actually don't have a strong preference among options (1)-(3). I like (2), which is what @shaobo-he implemented. But then it seems that `bwswap` would be hard to implement cleanly using that approach, which is why @keram88 went with (3). Is that right @keram88?

I have my doubts about (1). On one hand, it generates Boogie code directly and already inlined, and so there are no extraneous procedure invocations and definitions. That might be good for performance. On the other hand, we lose procedure boundaries that might be leveraged by Corral to abstract away intrinsics that are not needed to prove a property, which might be bad for performance.

Ultimately, I would choose between (1)-(3) based on which approach we think will lead to uniform and clean implementation of models of intrinsics.

shaobo-he:
> So I think our options for implementing models for intrinsics are as follows:
> 
>     1. Generate Boogie code that models an intrinsic directly and inline it. In this case there would be no functions or function calls being generated.
> 
>     2. Implement models for intrinsics as C functions in `intrinsics.c`, and then invoke the appropriate function every time an intrinsic is encountered.
> 
>     3. Implement models for intrinsics as Boogie functions/procedures in `Prelude.cpp`, and then invoke the appropriate function every time an intrinsic is encountered.
> 
>     4. Some combination of the above depending on an intrinsic.
> 
Let's use a concrete example to illustrate the differences between the four approaches. For example, we want to handle `lvm.bswap.i16`, my understanding of 1 is listed as follows,

| Option | Translation                              |
|--------|------------------------------------------|
| 1      | res := i[8:0]++i[16:8] |
| 2      | call res := __SMACK_bswap_i16(i)         |
| 3      | res := $bswap.bv16(i)                    |
| 4      | Not sure                                 |

Am I right here? It so, isn't option 1 and option 3 essentially the same since `$bswap.bv16` is Boogie function with inline attribute. Option 2 could also use this Boogie function via `__SMACK_code`, which means both `intrinsic.c` and `Prelude.cpp` are both modified.

zvonimir:
You are comparing them based on their Boogie syntax. I am more interested in comparing them based on their SMACK implementations. Meaning, I suggest we pick whichever solution is the easiest and cleanest to implement in SMACK, as opposed to picking them based on how the generated code looks like at the Boogie level. I hope this makes sense.

shaobo-he:
> You are comparing them based on their Boogie syntax. I am more interested in comparing them based on their SMACK implementations. Meaning, I suggest we pick whichever solution is the easiest and cleanest to implement in SMACK, as opposed to picking them based on how the generated code looks like at the Boogie level. I hope this makes sense.

I see. Let me push a version that I thought is ideal.

shaobo-he:
I pushed a new version. We will follow the option 3.

CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ if (NOT WIN32 OR MSYS OR CYGWIN)
   string(REPLACE "-Wl," "" LLVM_CXXFLAGS "${LLVM_CXXFLAGS}")
   string(REPLACE "-gsplit-dwarf" "" LLVM_CXXFLAGS "${LLVM_CXXFLAGS}")
   string(REGEX REPLACE "-O[0-9]" "" CMAKE_CXX_FLAGS "${LLVM_CXXFLAGS}")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-exceptions -fno-rtti -Wno-undefined-var-template")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-exceptions -fno-rtti -Wno-undefined-var-template -Wno-braced-scalar-init")
```
zvonimir: Why did you have to add this flag?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -145,39 +145,71 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+            Naming::HALF_TYPE, Expr::id(Naming::RMODE_VAR),
+            rep->expr(ii->getArgOperand(0))),
+          INDEXED_FUNC_APP(BitCast, "bv16", Naming::HALF_TYPE, rep->expr(ii)))));
+    } else
+      OVER_APPROXIMATE
+  };
+  static const auto expect = [&ii, this] {
+    // The llvm.expect.* function has two arguments: a value v and an expected
+    // value that is supposed to be used by optimizers.
+    // Semantically, this function simply returns the value v.
```
zvonimir: Add translation comment like you did for the other intrinsics.

zvonimir:
@keram88 and @shaobo-he : could we do `bvswap` using this pattern as well? This intrinsic is much more complicated, and so I am curious how its implementation will look like.

CMakeLists.txt
shaobo-he: We don't need it now.

shaobo-he:
> @keram88 and @shaobo-he : could we do `bvswap` using this pattern as well? This intrinsic is much more complicated, and so I am curious how its implementation will look like.

Yes, we can. The models for `bvswap` is inductively defined like this,
```
function {:inline} $bswap.bv16(i1: bv16) returns (bv16) { i1[8:0]++i1[16:8] }
function {:inline} $bswap.bv32(i1: bv32) returns (bv32) { i1[8:0]++$bswap.bv16(i1[24:8])++i1[32:24] }
function {:inline} $bswap.bv48(i1: bv48) returns (bv48) { i1[8:0]++$bswap.bv32(i1[40:8])++i1[48:40] }
function {:inline} $bswap.bv64(i1: bv64) returns (bv64) { i1[8:0]++$bswap.bv48(i1[56:8])++i1[64:56] }
function {:inline} $bswap.bv80(i1: bv80) returns (bv80) { i1[8:0]++$bswap.bv64(i1[72:8])++i1[80:72] }
function {:inline} $bswap.bv96(i1: bv96) returns (bv96) { i1[8:0]++$bswap.bv80(i1[88:8])++i1[96:88] }
```

lib/smack/Prelude.cpp
```diff
@@ -253,6 +258,23 @@ FuncDecl* truncBvOp(unsigned width, unsigned width2) {
     makeVars(1, type), type2, body, {makeInlineAttr()});
 }
 
+FuncDecl* bswapBvOp(unsigned width) {
+  std::string type = getBvTypeName(width);
+  std::string var = makeVarName(1);
+  const Expr* body;
+  if (width == 16) {
+    body = Expr::id(makeBvIndex(var,8,0) + "++" + makeBvIndex(var,16,8));
+  }
```
zvonimir: Fix code formatting.

lib/smack/Prelude.cpp
```diff
@@ -280,6 +302,17 @@ std::string getBytewisePointerStorageStr() {
   return s.str();
 }
 
+void generateIntrinsics(std::stringstream& prelude,
+			  const std::vector<unsigned> integer_sizes) {
+  const unsigned MAX_INT = *max_element(std::begin(integer_sizes), std::end(integer_sizes));
+  if (SmackOptions::BitPrecise) {
+    // bswap functions
+    for (size_t size = 16; size <= MAX_INT; size += 16) {
```
zvonimir: Could we not generate bvswap functions only for the relevant bit-widths that actually appear in `INTEGER_SIZES` instead of for all that are divisible by 16?
The idea would I guess be to generate bvswap only for bit-widths that are present in `INTEGER_SIZES` and are divisible by 16.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -145,39 +145,79 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
 }
 
 void SmackInstGenerator::processIntrinsicCall(llvm::IntrinsicInst* ii) {
+  #define OVER_APPROXIMATE \
```
michael-emmi: There must be a way to avoid preprocessor macros here…

shaobo-he: I added more macros :)

Why would we want to avoid them?

shaobo-he:
@keram88 It seems LLVM even has fixed point arithmetic intrinsics: https://llvm.org/docs/LangRef.html#id1924. You can put your fixed point encoding here.

lib/smack/SmackInstGenerator.cpp
michael-emmi: Search the web for “macros are evil” — there are several well-known reasons, and I don’t think I should have to explain them.

If anything, I would implore you to justify why you should use macros here and not functions. Sometimes there are valid reasons for doing so.

keram88:
Just saw this:
```
If src == 0 then the result is the size in bits of the type of src if is_zero_undef == 0 and undef otherwise.
```
in:
https://llvm.org/docs/LangRef.html#llvm-ctlz-intrinsic
How would we want to handle this?

lib/smack/SmackInstGenerator.cpp
shaobo-he: I removed all the macros defined in this function.

shaobo-he:
> Just saw this:
> 
> ```
> If src == 0 then the result is the size in bits of the type of src 
> if is_zero_undef == 0 and undef otherwise.
> ```
> 
> in:
> https://llvm.org/docs/LangRef.html#llvm-ctlz-intrinsic
> How would we want to handle this?

Does it mean we can also let the result be the size of bits even if `is_zero_undef==1`?

keram88:
> > Just saw this:
> > ```
> > If src == 0 then the result is the size in bits of the type of src 
> > if is_zero_undef == 0 and undef otherwise.
> > ```
> > 
> > 
> > in:
> > https://llvm.org/docs/LangRef.html#llvm-ctlz-intrinsic
> > How would we want to handle this?
> 
> Does it mean we can also let the result be the size of bits even if `is_zero_undef==1`?

Pretty much. Apparently some architectures return garbage when the input is zero. Modern x86 and ARM evidently work as expected. Maybe we should just leave it alone since I can only imagine it appearing with higher optimization levels.

lib/smack/Prelude.cpp
keram88: I changed the implementation so that the version for the bit-width 16 bits lower doesn't need to exist. The idea was to make the correctness of the definitions self evident, but we have tests :).

zvonimir:
Why do they have both `fptrunc` instruction and `trunc` intrinsic?
Their semantics seems identical to me.

zvonimir:
> > > Just saw this:
> > > ```
> > > If src == 0 then the result is the size in bits of the type of src 
> > > if is_zero_undef == 0 and undef otherwise.
> > > ```
> > > 
> > > 
> > > in:
> > > https://llvm.org/docs/LangRef.html#llvm-ctlz-intrinsic
> > > How would we want to handle this?
> > 
> > 
> > Does it mean we can also let the result be the size of bits even if `is_zero_undef==1`?
> 
> Pretty much. Apparently some architectures return garbage when the input is zero. Modern x86 and ARM evidently work as expected. Maybe we should just leave it alone since I can only imagine it appearing with higher optimization levels.

We should implement the semantics as fateful as we can, based on the LLVM documentation. So I think that in these situations this function should return an unconstrained value.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -145,39 +145,165 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+    errs() << "warning: over-approximate call to intrinsic: "
+      << ii->getCalledFunction()->getName() << "\n";
+    emit(rep->call(ii->getCalledFunction(), *ii));
+  };
+  static const auto assignFuncApp = [&ii, this]
+    (bool flag,
+     std::string funcNameBase,
+     std::initializer_list<std::string> funcNameArgs,
+     std::list<const Expr*> args) {
+      if (flag)
```
zvonimir: Why do you have this `flag` here, instead of calling `overApproximate` directly from `stmtMap`?
Just use `overApproximate` directly in `stmtMap`. It will make the code cleaner.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -145,39 +145,165 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+    // translation: $res := $<func>.bv*($arg1, $arg2);
+    (std::string funcNameBase) {
+    auto argT = rep->type(ii->getArgOperand(0)->getType());
+    assignFuncApp(SmackOptions::FloatEnabled,
+      funcNameBase,
+      {argT},
+      {rep->expr(ii->getArgOperand(0)), rep->expr(ii->getArgOperand(1))});
+  };
+  static const auto assignRoundFPFuncApp = [&ii, this]
+    // translation: $res := $round.bv*(rmode, $arg1);
```
zvonimir: It is weird that comment comes before the signature is finished. Fix this formatting.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -145,39 +145,165 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+        Expr::eq(
+          Expr::fn(indexedName("$fptrunc", {argT, Naming::HALF_TYPE}),
+            Expr::id(Naming::RMODE_VAR),
+            rep->expr(ii->getArgOperand(0))),
+          Expr::fn(indexedName("$bitcast", {retT, Naming::HALF_TYPE}),
+            rep->expr(ii)))));
+    } else
+      overApproximate(ii);
+  };
+  static const auto assignBVFuncApp = [&ii, this] (std::string funcNameBase){
```
zvonimir: Missing space in the end.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -145,39 +145,165 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+    // translation: $res := $bswap.bv*($arg1);
+    auto argT = rep->type(ii->getArgOperand(0)->getType());
+    assignFuncApp(SmackOptions::BitPrecise,
+      funcNameBase,
+      {argT},
+      {rep->expr(ii->getArgOperand(0))});
+  };
+  static const auto assignUnFPFuncApp = [&ii, this]
```
zvonimir: Why do we need this function?

zvonimir:
I left my review. Just as a reminder, our goal is to write code that is easy to understand and follow by others. That often does not equate to having the smallest implementation or having the smallest number of branches or using fancy C++ features or anything like that. In fact, it is often quite the opposite. So please think about this when implementing features in SMACK.

lib/smack/SmackInstGenerator.cpp
zvonimir: The more I think about this, to me it seems we should remove `assignFuncApp`.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -145,39 +145,165 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+    // translation: $res := $<func>.bv*($arg1);
+    (std::string funcNameBase) {
+    auto argT = rep->type(ii->getArgOperand(0)->getType());
+    assignFuncApp(SmackOptions::FloatEnabled,
+      funcNameBase,
+      {argT},
+      {rep->expr(ii->getArgOperand(0))});
+  };
+  static const auto assignBinFPFuncApp = [&ii, this]
```
zvonimir: Let's add new line after each one of these.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -145,39 +145,165 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+    // translation: $res := $<func>.bv*($arg1, $arg2);
+    (std::string funcNameBase) {
+    auto argT = rep->type(ii->getArgOperand(0)->getType());
+    assignFuncApp(SmackOptions::FloatEnabled,
+      funcNameBase,
+      {argT},
+      {rep->expr(ii->getArgOperand(0)), rep->expr(ii->getArgOperand(1))});
+  };
+  static const auto assignRoundFPFuncApp = [&ii, this]
```
zvonimir: Why do we need this function?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -145,39 +145,165 @@ void SmackInstGenerator::processInstruction(llvm::Instruction& inst) {
...
+    auto argT = rep->type(ii->getArgOperand(0)->getType());
+    assignFuncApp(SmackOptions::FloatEnabled,
+      funcNameBase,
+      {argT},
+      {rep->expr(ii->getArgOperand(0))});
+  };
+  static const auto assignBinFPFuncApp = [&ii, this]
+    // translation: $res := $<func>.bv*($arg1, $arg2);
+    (std::string funcNameBase) {
+    auto argT = rep->type(ii->getArgOperand(0)->getType());
```
zvonimir: No need for introducing `argT` local variable.

lib/smack/SmackInstGenerator.cpp
shaobo-he: And replace it with?

shaobo-he:
> Why do they have both `fptrunc` instruction and `trunc` intrinsic?
> Their semantics seems identical to me.

More precisely are we talking about these two?
https://llvm.org/docs/LangRef.html#fptrunc-to-instruction and https://llvm.org/docs/LangRef.html#llvm-trunc-intrinsic.

The former truncates a higher precision floating-point type to a lower precision one such as double -> float. The latter rounds a floating-point value to an integer value that is less than or equal to the floating-point value. So they are different.

lib/smack/SmackInstGenerator.cpp
shaobo-he: > Why do you have this `flag` here, instead of calling `overApproximate` directly from `stmtMap`?
> Just use `overApproximate` directly in `stmtMap`. It will make the code cleaner.

How can `overApproximate` be used in `stmtMap`?

lib/smack/SmackInstGenerator.cpp
shaobo-he: > Why do we need this function?

It's used to produce assignments where the rhs is a unary floating-point function application.

lib/smack/Prelude.cpp
```diff
@@ -253,6 +253,20 @@ FuncDecl* truncBvOp(unsigned width, unsigned width2) {
     makeVars(1, type), type2, body, {makeInlineAttr()});
 }
 
+FuncDecl* bswapBvOp(unsigned width) {
+  std::string type = getBvTypeName(width);
+  std::string var = makeVarName(1);
+  const Expr* body = Expr::bvConcat(Expr::bvExtract(var,width/2,width/2-8),
+                                    Expr::bvExtract(var,width/2+8,width/2));
+  for (unsigned i = 8; i < width/2; i += 8) {
```
zvonimir: Does this work for *any* bit-width, for example i37?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -882,4 +835,150 @@ void SmackInstGenerator::visitMemSetInst(llvm::MemSetInst& msi) {
...
+    {llvm::Intrinsic::sqrt,
+      [] { assignUnFPFuncApp("$sqrt"); }},
+    {llvm::Intrinsic::maxnum,
+      [] { assignBinFPFuncApp("$max"); }},
+    {llvm::Intrinsic::minnum,
+      [] { assignBinFPFuncApp("$min"); }},
+    {llvm::Intrinsic::ceil,
+      [] { assignRoundFPFuncApp("RTP"); }},
+    {llvm::Intrinsic::floor,
+      [] { assignRoundFPFuncApp("RTN"); }},
```
zvonimir: You should use `RModeLit` here and at other place where rounding mode is hard coded. See:
https://github.com/smackers/smack/blob/develop/include/smack/BoogieAst.h#L90

lib/smack/Prelude.cpp
```diff
@@ -253,6 +253,20 @@ FuncDecl* truncBvOp(unsigned width, unsigned width2) {
     makeVars(1, type), type2, body, {makeInlineAttr()});
 }
 
+FuncDecl* bswapBvOp(unsigned width) {
+  std::string type = getBvTypeName(width);
```
zvonimir: Why is this piece of code not generated directly in `SmackInstGenerator`?
These functions are not calling each other any more, and so I am not sure why we would introduce them in the prelude.

zvonimir: @keram88 or @shaobo-he - please take a look at this question and maybe address it (or answer it).

lib/smack/Prelude.cpp
zvonimir: @keram88 or @shaobo-he - please take a look at this question and maybe address it (or answer it).

lib/smack/Prelude.cpp
shaobo-he: > Why is this piece of code not generated directly in `SmackInstGenerator`?
> These functions are not calling each other any more, and so I am not sure why we would introduce them in the prelude.

Do you mean we put the expression into SmackInstGenerator which has the same effect of inlining these functions?

lib/smack/Prelude.cpp
shaobo-he: > Does this work for _any_ bit-width, for example i37?

Not sure I understand this question. These functions are generated only when the bit-width can be divided by 16.

lib/smack/Prelude.cpp
zvonimir: Yes. I think @keram88 just did that.

zvonimir: Yes. I think @keram88 just did that.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -882,4 +832,179 @@ void SmackInstGenerator::visitMemSetInst(llvm::MemSetInst& msi) {
   emit(rep->memset(msi));
 }
 
+void SmackInstGenerator::generateUnModeledCall(llvm::CallInst* ci) {
+  if (ci->getFunctionType()->getReturnType()->isVoidTy()) {
+    WARN("ignoring call to "+ci->getCalledFunction()->getName().str());
```
zvonimir: Why is this `WARN` and the one below is `errs`?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -882,4 +832,179 @@ void SmackInstGenerator::visitMemSetInst(llvm::MemSetInst& msi) {
...
+              Expr::fn(indexedName("$fptrunc", {argT, Naming::HALF_TYPE}),
+                Expr::id(Naming::RMODE_VAR),
+                rep->expr(ci->getArgOperand(0))),
+              Expr::fn(indexedName("$bitcast", {retT, Naming::HALF_TYPE}),
+                rep->expr(ci)))));
+    } else
+      generateUnModeledCall(ci);
+  };
+
+  static const auto assignBVFuncApp = [this] (CallInst* ci, std::string fnBase) {
```
zvonimir: This function is not being use anywhere any more it seems. We should probably remove it.

lib/smack/Prelude.cpp
keram88: Only bit-widths that are a multiple of 16 are allowed by LLVM. Putting other widths in appears to crash Smack when the LLVM code gets first processed. We could still put a check here.

zvonimir: I get it. All good. No need for a check here.

lib/smack/SmackInstGenerator.cpp
keram88: It appears errs is being used to print a message while SMACK is running to warn of unmodeled behavior. @shaobo-he wrote this so he might be able to give his reasoning.

lib/smack/SmackInstGenerator.cpp

lib/smack/SmackInstGenerator.cpp
shaobo-he: If there's no return type, then calling a procedure without an implementation is harmless otherwise we could have false alarms.

zvonimir: Ah, so there are no intrinsics that have side-effects? In other words, you are saying that if there's no return type, then calling a procedure without an implementation is sound?

shaobo-he: > Ah, so there are no intrinsics that have side-effects? In other words, you are saying that if there's no return type, then calling a procedure without an implementation is sound?

There are some intrinsics like `mem.set` has side-effects but all of what I've seen are handled. Currently we handle `llvm.dbg` intrinsics using this function. If we give warnings printed on the screen, there will be a lot.

zvonimir:
Please squash this and I'll merge it once @michael-emmi approves it.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -882,4 +832,167 @@ void SmackInstGenerator::visitMemSetInst(llvm::MemSetInst& msi) {
...
+  // meaning we can use __builtin_* to implement math.c which is mostly
+  // modeled using __SMACK_code.
+
+  static const std::map<llvm::Intrinsic::ID, std::function<void(CallInst*)>> stmtMap {
+    {llvm::Intrinsic::convert_from_fp16, f16UpCast},
+    {llvm::Intrinsic::convert_to_fp16, f16DownCast},
+    {llvm::Intrinsic::bswap, bswap},
+    {llvm::Intrinsic::expect, identity},
+    {llvm::Intrinsic::fabs,
+      [] (CallInst* ci){ assignUnFPFuncApp(ci,"$abs"); }},
```
michael-emmi: Seems like you could simplify the following cases quite a bit by currying. For instance, rather than defining the `assignUnFPFuncApp` closure above, define the following function:
```C++
std::function<void(CallInst*)> SmackInstGenerator::assignUnFPFuncApp(std::string fnBase) {
  return [this, fnBase] (CallInst* ci) {
    // translation: $res := $<func>.bv*($arg1);
    if (SmackOptions::FloatEnabled)
      emit(Stmt::assign(
            rep->expr(ci),
            Expr::fn(indexedName(fnBase,
                {rep->type(ci->getArgOperand(0)->getType())}),
              rep->expr(ci->getArgOperand(0)))));
    else
      generateUnModeledCall(ci);
  };
}
```
Then you could replace the current line with simply:
```C++
assignUnFPFuncApp("$abs")},
```

shaobo-he: Good suggestion. Is there a particular reason that this function is made as a member method instead of a local variable?

michael-emmi: That’s probably just stylistic preferences; either way resolves the issue.

Issue #406, opened 2019-01-21
Clean up or document SMACK warnings


zvonimir:
Warnings such as:
- collapsing DSA node
- memory intrinsic length exceeds threshold (0); adding quantifiers
- ignoring inline asm passed as argument

are repeatedly confusing our users. We should think about not printing them out by default maybe, and we should also rethink if they are all really needed.

zvonimir:
Ok, so let's maybe discuss this a little bit. I think that the warnings I listed below should not be printed by default. We should have some kind of a command line option that you have to enable to see these warnings.

zvonimir:
Actually, now that I think more about this, maybe we should leave this warning `ignoring inline asm passed as argument`. Since that one I think can affect soundness and it would be good for users to see it.

shaobo-he:
I talked to Mark about this. I propose we use multi-level verbose/debug output.

zvonimir:
Yes, I agree with you guys. Could you please propose an implementation for this functionality?

shaobo-he:
@keram88 Can you take a look at this?

shaobo-he:
In light of recent development on handling LLVM intrinsics, we should probably give warnings for over-approximations such as bitwise operations when the `--bit-precise` flag is not enabled.

shaobo-he:
In the current develop branch, we have the following warnings printed to the console,

1. inline asm
2. memory intrinsic length
3. those generated by the `DSMonitor` class such as `collapsing DSA node`
4. over-approximations in intrinsic handling

I also propose we extend 4. I propose we use a two-tier warning outputs. The first tier includes 1 and 4 which is important due to its influences on soundness. The second tier includes 2 and 3. 

@zvonimir @michael-emmi @keram88 what do you think?

keram88:
I think 4 should be as informative as possible. We could go so far as to say something like `(try --bit-precise)`. This would cut down on surprises for users (and occasional debugging holes for us).

shaobo-he:
Good point. We should do this.

keram88:
For the fourth item. I'm thinking of the following:
We create an `enum class`
`enum class Requirements { BitPrecise, Float, Strings, ... };`

Then we have a requirements predicate:
```
bool meetsRequirements(std::initializer_list<Requirements> reqs) {
  bool success = true;
  for (req : reqs) {
    switch(req) {
    case (BitPrecise): true &= SmackOptions::BitPrecise; break;
     ...
    }
  }  
}
```

[...]

Issue #321, opened 2018-06-05
Adding ll files as regressions to SMACK


zvonimir:
This came up several times in the past several months. As the usage scenarios of SMACK are expanding beyond just C/C++ programs, we are running into situations where we need regressions for LLVM IR programs that are not easy (or are even impossible) to generate from C/C++.
I think the right move would be to add ll files to our regression suite for such situations. SMACK already supports taking ll files as input, and it probably would not be hard to update `regtest.py` to support ll files.
The main drawback that I can see is that LLVM IR sometimes changes between LLVM versions, and so such regressions would have to be updated with LLVM versions sometimes.
Thoughts?

michael-emmi:
Why not just run the regression conditionally, with the frontend that generates the code, if that frontend is available — e.g., in the path — and otherwise ignore the regression? Is this too much trouble?

zvonimir:
As we are slowly moving towards multi-language verification, what you suggested would I guess make sense in the long run. I mean, we already have Rust regressions somewhere in @keram88 Rust branch.
How should we structure the regressions in that case? Should we have top-level folders indicating languages (c,cpp,ll,rust,etc.), and then subfolders within these with regression features? Or something else?

michael-emmi:
On the other hand, perhaps maintaining LLVM-code regression tests is not as bad as we think. After all, how does LLVM handle this? Surely they don’t have to rewrite hundreds/thousands of regression tests every time they add/update some language feature?

michael-emmi:
The filesystem organization that you propose sounds reasonable to me.

keram88:
I investigated this a bit ago and found their AutoUpgrade functionality, but wasn't sure about how committed they were to it based on some mailing list posts. 
However, I finally found what looks like LLVM's policy:
> IR Backwards Compatibility
> 
> When the IR format has to be changed, keep in mind that we try to maintain some backwards compatibility. The rules are intended as a balance between convenience for llvm users and not imposing a big burden on llvm developers:
> 
> - The textual format is not backwards compatible. We don’t change it too often, but there are no specific promises.
> - Additions and changes to the IR should be reflected in test/Bitcode/compatibility.ll.
> - The current LLVM version supports loading any bitcode since version 3.0.
> - After each X.Y release, compatibility.ll must be copied to compatibility-X.Y.ll. The corresponding bitcode file should be assembled using the X.Y build and committed as compatibility-X.Y.ll.bc.
[...]

PR #464, opened 2019-06-11
[WIP] Add LLVM bit intrinsics

keram88:
This pull request aims to add several of the LLVM bit-wise intrinsics such as count leading zeros.
Optionally, we can add vectorized versions of these operations as well.
- [x] bswap - Already in develop - here for completeness.
- [x] bitreverse
- [x] ctlz
- [x] ctpop
- [x] cttz
- [x] ~~fshl~~ Not available in LLVM 4
- [x] ~~fshr~~

I think we should focus on those that have been seen in real programs such as ctlz, and those that are easy to generate the intrinsic through compiler builtins.

This is intended to be merged after the `improve-prelude` pull request since there is some functionality I want to use.

### DIFF
--- a/lib/smack/SmackInstGenerator.cpp
+++ b/lib/smack/SmackInstGenerator.cpp
@@ -889,26 +889,131 @@ void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst& ii) {
     else
       generateUnModeledCall(ci);
   };
+  
+  static const auto bitreverse = [this](Value* arg){
+    auto width = arg->getType()->getIntegerBitWidth();
+    auto var = rep->expr(arg);
 
-  static const auto bswap = [this](CallInst* ci){
+    // Swap the bits to the right and left of the middle
+    const Expr* body;
+    if (width % 2 == 0) {
+      body = Expr::bvConcat(Expr::bvExtract(var,width/2,width/2-1),
+                            Expr::bvExtract(var,width/2+1,width/2));
+    }
+    else {
+      body = Expr::bvExtract(var, width/2+1, width/2);
+    }
+    // Swap the bits to the right and the left of the already swapped portion.
+    unsigned offset = width&1;
+    for (unsigned i = width%2==0 ? 1 : 0; i < width/2; ++i) {
+      body = Expr::bvConcat(Expr::bvConcat(Expr::bvExtract(var,width/2-i,width/2-i-1),
+                                           body),
+                            Expr::bvExtract(var, width/2+i+1+offset, width/2+i+offset));
+    }
+    return body;
+  };
+
+  static const auto bswap = [this](Value* arg){
+    auto width = arg->getType()->getIntegerBitWidth();
+    auto var = rep->expr(arg);
+
+    // Swap the bytes to the right and left of the middle
+    const Expr* body = Expr::bvConcat(Expr::bvExtract(var,width/2,width/2-8),
+                                      Expr::bvExtract(var,width/2+8,width/2));
+
+    // Swap the bytes to the right and the left of the already swapped portion.
+    for (unsigned i = 8; i < width/2; i += 8) {
+      body = Expr::bvConcat(Expr::bvConcat(Expr::bvExtract(var,width/2-i,width/2-i-8),
+                                           body),
+                            Expr::bvExtract(var, width/2+i+8, width/2+i));
+    }
+    return body;
+  };
+
+  // Count leading zeros
+  static const auto ctlz = [this](CallInst* ci) {
     if (SmackOptions::BitPrecise) {
       auto width = ci->getArgOperand(0)->getType()->getIntegerBitWidth();
       auto var = rep->expr(ci->getArgOperand(0));
 
-      // Swap the bytes to the right and left of the middle
-      const Expr* body = Expr::bvConcat(Expr::bvExtract(var,width/2,width/2-8),
-                                        Expr::bvExtract(var,width/2+8,width/2));
-
-      // Swap the bytes to the right and the left of the already swapped portion.
-      for (unsigned i = 8; i < width/2; i += 8) {
-        body = Expr::bvConcat(Expr::bvConcat(Expr::bvExtract(var,width/2-i,width/2-i-8),
-                                             body),
-                              Expr::bvExtract(var, width/2+i+8, width/2+i));
+      // e.g., if v[32:31] == 1 then 0bv32 else if v[31:30] == 1 then 1bv32 else
+      // ... else if v[1:0] == 1 then 31bv32 else 32bv32
+      const Expr* body = Expr::lit(width, width);
+      for (unsigned i = 0; i < width; ++i) {
+        body = Expr::if_then_else(Expr::eq(Expr::bvExtract(var, i+1, i),
+                                           Expr::lit(1,1)),
+                                  Expr::lit(width-i-1, width),
+                                  body);
       }
+
+      // Handle the is_zero_undef case, i.e. if the flag is set and the argument
+      // is zero, then the result is undefined.
+      auto isZeroUndef = rep->expr(ci->getArgOperand(1));
+      body = Expr::if_then_else(Expr::and_(Expr::eq(isZeroUndef, Expr::lit(1, 1)),
+                                           Expr::eq(var, Expr::lit(0,width))),
+                                rep->expr(ci), // The result is undefined
+                                body);
+      emit(Stmt::havoc(rep->expr(ci)));
       emit(Stmt::assign(rep->expr(ci), body));
     }
     else
       generateUnModeledCall(ci);
+  };
+
+  // Count trailing zeros
+  static const auto cttz = [this](CallInst* ci) {
+    if (SmackOptions::BitPrecise) {
+      auto width = ci->getArgOperand(0)->getType()->getIntegerBitWidth();
+      auto arg = rep->expr(ci->getArgOperand(0));
+
+      // e.g., if v[1:0] == 1 then 0bv32 else if v[2:1] == 1 then 1bv32 else
+      // ... else if v[32:31] == 1 then 31bv32 else 32bv32
+      const Expr* body = Expr::lit(width, width);
+      for (unsigned i = width; i > 0; --i) {
+        body = Expr::if_then_else(Expr::eq(Expr::bvExtract(arg, i, i-1),
+                                           Expr::lit(1,1)),
+                                  Expr::lit(i-1, width),
+                                  body);
+      }
+
+      // Handle the is_zero_undef case, i.e. if the flag is set and the argument
+      // is zero, then the result is undefined.
+      auto isZeroUndef = rep->expr(ci->getArgOperand(1));
+      body = Expr::if_then_else(Expr::and_(Expr::eq(isZeroUndef, Expr::lit(1, 1)),
+                                           Expr::eq(arg, Expr::lit(0,width))),
+                                rep->expr(ci), // The result is undefined
+                                body);
+      emit(Stmt::havoc(rep->expr(ci)));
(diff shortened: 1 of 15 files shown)


keram88:
My ctpop (count 1s) implementation is very slow. The 8-bit version looks like this in pseudo-code:
`$sext.bv1.bv8(v[8:7])+$sext.bv1.bv8(v[7:6])+$sext.bv1.bv8(v[6:5])+...+$sext.bv1.bv8(v[1:0])`
This is basically what Nikolaj Bjorner recommended in a StackOverflow answer, so I'm not sure if we can do better.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -910,6 +910,84 @@ void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst& ii) {
...
+                                  Expr::lit(i-1, width),
+                                  body);
+      }
+
+      // Handle the is_zero_undef case, i.e. if the flag is set and the argument
+      // is zero, then the result is undefined.
+      auto isZeroUndef = rep->expr(ci->getArgOperand(1));
+      body = Expr::if_then_else(Expr::and_(Expr::eq(isZeroUndef, Expr::lit(1, 1)),
+                                           Expr::eq(arg, Expr::lit(0,width))),
+                                rep->expr(ci), // The result is undefined
```
keram88: I couldn't find a way to create a nondet value directly using our functionality. This assigns the variable to itself, which is fine so long as it hasn't been assigned previously. I'd like something more predictable.

keram88: Inserted a havoc to make sure the value is unknown.

keram88:
I think this is complete for the non-vector bit intrinsics.

shaobo-he:
It appears that we can simplify the code like this,

```C++
static const auto assignBvExpr = [this] (std::function<const Expr*(Value*)> exprGenFunc) {
    return [this, exprGenFunc] (CallInst* ci) {
      if (SmackOptions::BitPrecise)
        emit(Stmt::assign(
              rep->expr(ci),
              exprGenFunc(ci->getArgOperand(0))));
      else
        generateUnModeledCall(ci);
    };
  };
```
Then functions like `bitreverse` becomes,
```C++
  static const auto bitreverse = [this](Value* arg){
    auto width = arg->getType()->getIntegerBitWidth();
    auto var = rep->expr(arg);

    // Swap the bits to the right and left of the middle
    const Expr* body;
    if (width % 2 == 0) {
      body = Expr::bvConcat(Expr::bvExtract(var,width/2,width/2-1),
          Expr::bvExtract(var,width/2+1,width/2));
    }
    else {
      body = Expr::bvExtract(var, width/2+1, width/2);
    }
    // Swap the bytes to the right and the left of the already swapped portion.
    for (unsigned i = 1; i < width/2; i += 1) {
      body = Expr::bvConcat(Expr::bvConcat(Expr::bvExtract(var,width/2-i,width/2-i-1),
            body),
          Expr::bvExtract(var, width/2+i+1, width/2+i));
    }
    return body;
  };
```
Finally, its dispatch becomes `assignBvExpr(bitreverse)`.

keram88:
Yes. I didn't see that. I'll fix it once Travis finishes so I don't have to update the regressions again.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -904,7 +904,7 @@ void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst& ii) {
       body = Expr::bvExtract(var, width/2+1, width/2);
     }
     // Swap the bits to the right and the left of the already swapped portion.
-    unsigned offset = width%2==0 ? 0 : 1;
+    unsigned offset = width&1;
     for (unsigned i = width%2==0 ? 1 : 0; i < width/2; ++i) {
```
shaobo-he: Isn't `width%2==0 ? 1 : 0` equivalent to `offset^1`?

keram88: Yes, but I want to keep the emphasis on its relation to the bit-width.

keram88: `width^1` would also work, but `width%2==0 ...` is more obvious to me because of the parity.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -889,28 +889,133 @@ void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst& ii) {
...
-                                             body),
-                              Expr::bvExtract(var, width/2+i+8, width/2+i));
+      // Handle the is_zero_undef case, i.e. if the flag is set and the argument
+      // is zero, then the result is undefined.
+      auto isZeroUndef = rep->expr(ci->getArgOperand(1));
+      body = Expr::if_then_else(Expr::and_(Expr::eq(isZeroUndef, Expr::lit(1, 1)),
+                                           Expr::eq(var, Expr::lit(0,width))),
+                                rep->expr(ci), // The result is undefined
+                                body);
+      emit(Stmt::havoc(rep->expr(ci)));
```
zvonimir: Why do you have this havoc before the assignment?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -889,28 +889,133 @@ void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst& ii) {
...
       }
+
+      // Handle the is_zero_undef case, i.e. if the flag is set and the argument
+      // is zero, then the result is undefined.
+      auto isZeroUndef = rep->expr(ci->getArgOperand(1));
+      body = Expr::if_then_else(Expr::and_(Expr::eq(isZeroUndef, Expr::lit(1, 1)),
+                                           Expr::eq(arg, Expr::lit(0,width))),
+                                rep->expr(ci), // The result is undefined
+                                body);
+      emit(Stmt::havoc(rep->expr(ci)));
```
zvonimir: Same question - why havoc before the assignment?

test/bits/bitreverse.c
```diff
@@ -0,0 +1,54 @@
...
+  result |= (x&0x04000000) >> 21; // 26->5
+  result |= (x&0x08000000) >> 23; // 27->4
+  result |= (x&0x10000000) >> 25; // 28->3
+  result |= (x&0x20000000) >> 27; // 29->2
+  result |= (x&0x40000000) >> 29; // 30->1
+  result |= (x&0x80000000) >> 31; // 31->0
+  return result;
+}
+
+int main() {
```
zvonimir: This should be `int main(void)`, and `main` should also return an integer.

test/bits/bitreverse.c
```diff
@@ -0,0 +1,54 @@
...
+  uint16_t x = __VERIFIER_nondet_unsigned_short();
+  uint32_t y = __VERIFIER_nondet_unsigned_int();
+  __VERIFIER_assert(__builtin_bitreverse16(x) == bitreverse16(x));
```
zvonimir: We use just `assert` in our regressions.

test/bits/countpop32.c
```diff
@@ -0,0 +1,46 @@
...
+  result += x&1;  x >>= 1;
+  result += x&1;  x >>= 1;
+  result += x&1;  x >>= 1;
+  result += x&1;  x >>= 1;
+  result += x&1;  x >>= 1;
+  result += x&1;
+  return result;
+}
+
+int main() {
```
zvonimir: This should be `int main(void)`, and main should also return an integer.

test/bits/countpop32_fail.c
```diff
@@ -0,0 +1,46 @@
...
+  result += x&1;  x >>= 1;
+  result += x&1;  x >>= 1;
+  result += x&1;  x >>= 1;
+  result += x&1;  x >>= 1;
+  result += x&1;  x >>= 1;
+  result += x&1;
+  return result;
+}
+
+int main() {
```
zvonimir: This should be `int main(void)`, and main should also return an integer.

lib/smack/SmackInstGenerator.cpp
keram88: Basically what this is doing is this:
```
$i4 = if (($is_zero_undef == 1bv1) &&
              ($i1 == 0bv32))
          then $i4 else
          ctlz($i4)
```
The havoc is there to make `$i4` nondeterministic since LLVM says this is undefined. This comes up if this is being done in a loop where `$i4` gets an old value. I added an example which passes if it's not havoc'd first.

Perhaps it would be better to have an assertion failure instead?

keram88: Related to #406. The `is_zero_undef` argument to `llvm.ctz.*(%arg, %is_zero_undef)` is a constant, so we could emit a warning if `is_zero_undef==true` that the user may want to use another flag (something like `--undefined`) if they get weird results.

zvonimir: Ah, ok, I get it. What do you mean by "to have an assertion failure instead"?
I also don't quite get your last comment.

keram88: What I'm thinking of is that someone gets an assertion violation from code using this function with zero undefined, the cause may not be very obvious. I'm thinking we can give a simple warning that this can happen and use a general undefined behavior assertion category for catching these errors. As it is, I think this is well outside the scope of this PR.

zvonimir:
Why does this pull request modify `Prelude.cpp`?
I see Shaobo's commits in this pull request, which I think are not related to it. I think that something got messed up. Could you maybe force-push to clean this up somehow?
Oh, or maybe this branch was created starting from Shaobo's prelude branch?
I am confused here.

keram88:
This was created with the Prelude branch, because there was some functionality I intended to use, and some modifications I anticipated making. It turned out I only needed to make a new `Stmt::havoc` function. I think I'll rebase it back to develop, though I am okay if the merge order is Prelude then this.

keram88:
I checked out from `develop` the files changed from the improve Prelude branch.
The changes in this PR are:
- Models for the bit-intrinsics outlined at the top of the PR.
- Regression tests for the new intrinsics.
- Small changes to `BoogieAST` to make implementation easier.

The one issue is the regression tests for `lllvm.ctpop` which are long-running. I doubt we'll be able to make them fast, but I did manage to get them to run on Travis in at most 80 seconds. If this is a problem, we may want to move it to `timeouts`, or another long-running test folder.

shaobo-he:
> I checked out from `develop` the files changed from the improve Prelude branch.
> The changes in this PR are:
> 
> * Models for the bit-intrinsics outlined at the top of the PR.
> * Regression tests for the new intrinsics.
> * Small changes to `BoogieAST` to make implementation easier.
> 
> The one issue is the regression tests for `lllvm.ctpop` which are long-running. I doubt we'll be able to make them fast, but I did manage to get them to run on Travis in at most 80 seconds. If this is a problem, we may want to move it to `timeouts`, or another long-running test folder.
> 
> I think as it stands, this is okay for merging.

Maybe we can reduce the bit-widths of these timedout regressions.

keram88:
> Maybe we can reduce the bit-widths of these timedout regressions.

I tried. Clang only has builtins for 32bits and 64bits.

keram88:
Perhaps we can approach the countpop timeout by adding an LLVM test a la issue #321. We may need to choose our priorities here, so maybe for now we can move the countpop tests to timeouts and record it in an issue.

zvonimir:
Yes, I would be ok if you put a regression into timeouts.

keram88:
I got the run time on popcount down to a (relatively) reasonable level. I made it test inputs `<2<sup>16</sup>` and `>=2<sup>16</sup>` separately. Travis consistently finishes `countpop32.c` in 20-40 seconds.

shaobo-he:
LGTM

PR #466, opened 2019-06-18
Add a more consistent and sophisticated warning system

shaobo-he:


lib/smack/SmackWarnings.cpp
```diff
@@ -0,0 +1,34 @@
...
+    std::string flags = "";
+    bool sat = true;
+    for (auto rf : requiredFlags) {
+      if (!*rf) {
+        sat = false;
+        flags += (rf->ArgStr.str() + ",");
+      }
+    }
+    if (!sat) {
+      std::string msg = "warning: over-approximating " + name + "; try add flag(s): ";
```
zvonimir: It should be `try adding flag(s): `

lib/smack/SmackWarnings.cpp
```diff
@@ -0,0 +1,34 @@
+#include "llvm/Support/Debug.h"
+#include "smack/SmackWarnings.h"
+
+
+namespace smack {
+using namespace llvm;
+void SmackWarnings::warnWithType(const char* type, std::string msg) {
+  if (isCurrentDebugType(type))
```
zvonimir: Why do you need this if condition? Could you not use the LLVM macro instead?

lib/smack/SmackWarnings.cpp
```diff
@@ -0,0 +1,34 @@
...
+    dbgs() << msg << "\n";
+}
```
zvonimir: Could we add these warnings into the bpl file as well?

lib/smack/SmackWarnings.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+void SmackWarnings::warnIfUnsound(std::string name, std::initializer_list<const cl::opt<bool>*> requiredFlags) {
+    std::string flags = "";
+    bool sat = true;
```
zvonimir: Please rename this variable into something more descriptive.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -847,36 +849,47 @@ void SmackInstGenerator::generateUnModeledCall(llvm::CallInst* ci) {
 void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst& ii) {
   processInstruction(ii);
 
-  static const auto f16UpCast = [this] (CallInst* ci) {
+//(CallInst -> Void) -> [Flags] -> (CallInst -> Void)
+  static const auto conditionalModel = [this] (std::function<void(CallInst*)> modelGenFunc,
+    std::initializer_list<const cl::opt<bool>*> requiredFlags) {
+    return [this, requiredFlags, modelGenFunc] (CallInst* ci) {
+      bool sat = true;
+      for (auto rf : requiredFlags)
```
zvonimir: Do you have to iterate over `requiredFlags` both here and in `warnIfUnsound`?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -847,36 +849,47 @@ void SmackInstGenerator::generateUnModeledCall(llvm::CallInst* ci) {
...
+        if (!*rf)
```
zvonimir: Instead of searching for `false` using this for loop, couldn't you use some kind of a built-in function for searching over C++ list?

lib/smack/SmackWarnings.cpp
shaobo-he: It appears that the development process suggests what @keram88 proposed as opposed to using LLVM debug system. I think macros like `DEBUG_WITH_TYPE` creates names (i.e., the first arguments) that may conflict with other names, which may not be checked at compile time. W.r.t this condition, it's purpose is to bypass the message generation statements if the current warning level is too low.

zvonimir:
Are you always printing warnings as comments into bpl file?

shaobo-he:
> Are you always printing warnings as comments into bpl file?

No, they are printed only when the warning level is greater than or equal to unsound.

lib/smack/SmackWarnings.cpp
```diff
@@ -0,0 +1,95 @@
...
+  UnsetFlagsT ret;
+  std::copy_if(requiredFlags.begin(), requiredFlags.end(), std::inserter(ret, ret.begin()),
+    [](FlagT* flag) {return !*flag;});
+  return ret;
+}
+
+std::string SmackWarnings::getFlagStr(UnsetFlagsT flags) {
+  std::string ret = "";
+  for (auto f : flags)
+    ret += (f->ArgStr.str() + ",");
```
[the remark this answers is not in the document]
shaobo-he: Yes.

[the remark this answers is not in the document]
keram88: I guess we should drop the separating comma too.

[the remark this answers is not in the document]
shaobo-he: Done.

shaobo-he:
We need to do some refactoring after the bv intrinsic handling PR is merged.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -834,61 +841,66 @@ void SmackInstGenerator::visitMemSetInst(llvm::MemSetInst& msi) {
 }
 
 void SmackInstGenerator::generateUnModeledCall(llvm::CallInst* ci) {
-  if (ci->getFunctionType()->getReturnType()->isVoidTy()) {
-    WARN("ignoring call to "+ci->getCalledFunction()->getName().str());
+  if (ci->getType()->isVoidTy()) {
+    //SmackWarnings::warnUnsound(ci->getCalledFunction()->getName(), currBlock, ci, true);
```
zvonimir: Why is this commented out?

lib/smack/SmackOptions.cpp
```diff
@@ -14,8 +14,12 @@ const llvm::cl::list<std::string> SmackOptions::EntryPoints(
   llvm::cl::value_desc("PROCS")
 );
 
-const llvm::cl::opt<bool> SmackOptions::Warnings(
-  "warnings", llvm::cl::desc("Enable warnings.")
+const llvm::cl::opt<std::string> SmackOptions::WarningType(
```
zvonimir: What are the available warning types? I thought you can create an option with preset available types.

lib/smack/SmackWarnings.cpp
```diff
@@ -0,0 +1,95 @@
...
+    warningLevel = WarningLevel::Silent;
+  else
+    llvm_unreachable("invalid warning type.");
+}
+
+bool SmackWarnings::isSufficientWarningLevel(WarningLevel level) {
+  return warningLevel >= level;
+}
+
+SmackWarnings::UnsetFlagsT SmackWarnings::getUnsetFlags(RequiredFlagsT requiredFlags) {
```
zvonimir: What does this do?

lib/smack/SmackWarnings.cpp
```diff
@@ -0,0 +1,95 @@
...
+  if (unsetFlags.size())
+    warnUnsound(name, unsetFlags, currBlock, i, ignore);
+}
+
+void SmackWarnings::warnUnsound(std::string unmodeledOpName, Block* currBlock, const Instruction* i, bool ignore) {
+  warnUnsound("unmodeled operation " + unmodeledOpName, UnsetFlagsT(), currBlock, i, ignore);
+}
+
+void SmackWarnings::warnUnsound(std::string name, UnsetFlagsT unsetFlags, Block* currBlock, const Instruction* i, bool ignore) {
+  if (isSufficientWarningLevel(WarningLevel::Unsound)) {
```
zvonimir: You should instead exit early, which is what LLVM coding standards recommend:
```
  if (!isSufficientWarningLevel(WarningLevel::Unsound)) return;
```

lib/smack/SmackWarnings.cpp
```diff
@@ -0,0 +1,95 @@
...
+    std::string beginning = std::string("llvm2bpl: ") + buildDebugInfo(i);
+    std::string end = (ignore? "unsoundly ignoring " : "over-approximating ") + name + ";";
+    if (currBlock)
+      currBlock->addStmt(Stmt::comment(beginning + "warning: " + end));
+    std::string hint = "";
+    if (unsetFlags.size())
+      hint = (" try adding flag(s): " + getFlagStr(unsetFlags));
+    errs() << beginning;
+    bool hasColors = SmackOptions::ColoredWarnings;
```
zvonimir: No need to introduce this local variable I don't think.

share/smack/top.py
```diff
@@ -110,6 +110,9 @@ def arguments():
   noise_group.add_argument('--debug-only', metavar='MODULES', default=None,
     type=str, help='limit debugging output to given MODULES')
 
+  noise_group.add_argument('--warn', metavar='OPTION', default="unsound",
```
zvonimir: What are the available options? You should restrict them, which is possible in python command line parser.

lib/smack/SmackInstGenerator.cpp
shaobo-he: This is relevant to this comment: https://github.com/smackers/smack/pull/451#discussion_r290532783. We don't want to show a lot of `llvm.dbg` intrinsics.

lib/smack/SmackWarnings.cpp
shaobo-he: It tests the list of required flags are set or not. For example, handling the half-precision intrinsics requires both `--float` and `--bit-precise`.  If one of the flags is set, we only need to remind the users to set the other.

zvonimir:
I have a more high-level concern about this pull request. I know we already discussed this to some extent, but it is still not completely clear to me how come we could not simply use LLVM's `DEBUG` macro functionality to achieve this.

What is `DEBUG` missing that you could not implement this using it?

Also, we use `DEBUG` macros a fair amount in the code. So if this gets merge, we will have a combination of `DEBUG` macros and calls to `SmackWarnings` in the code. How should then I think about them? Do they serve a different purpose? Or is the plan to switch to `SmackWarnings` completely and remove `DEBUG` macros?

shaobo-he:
> I have a more high-level concern about this pull request. I know we already discussed this to some extent, but it is still not completely clear to me how come we could not simply use LLVM's `DEBUG` macro functionality to achieve this.
> 
> What is `DEBUG` missing that you could not implement this using it?
> 
> Also, we use `DEBUG` macros a fair amount in the code. So if this gets merge, we will have a combination of `DEBUG` macros and calls to `SmackWarnings` in the code. How should then I think about them? Do they serve a different purpose? Or is the plan to switch to `SmackWarnings` completely and remove `DEBUG` macros?

First of all, the existing `DEBUG` macros serve a different purpose than `SmackWarnings`. The former is used to print information used for debugging as its name stands for while the latter is to print warnings messages to the users. From this perspective, we don't have to worry about mixing the two implementations.

Second, when I first implemented this PR, I used LLVM's `DEBUG` macros, which turned out to be very cumbersome. The current implementation is much cleaner than the one using macros. First, it creates a very clear and informative interfaces: arguments passed to llvm2bpl are meaningful such as `-warn-type=unsound` as opposed to `--debug-only=unsound` if we use LLVM's macros. Second, handling warning message level is intuitive and simple in the current implementation: we simply compare the enum values as opposed to comparing the macro strings.

lib/smack/SmackInstGenerator.cpp
zvonimir: Ok. But then why is is commented out, instead of completely erased?

lib/smack/SmackOptions.cpp
```diff
@@ -14,8 +14,20 @@ const llvm::cl::list<std::string> SmackOptions::EntryPoints(
...
+const llvm::cl::opt<SmackWarnings::WarningLevel> SmackOptions::WarningLevel(
+  "warn-type",
+  llvm::cl::desc("Enable certain type of warning messages."),
+  llvm::cl::values(
+    clEnumValN(SmackWarnings::WarningLevel::Silent,
+      "silent", "No warning messages"),
+    clEnumValN(SmackWarnings::WarningLevel::Unsound,
+      "unsound", "Enable warnings about unsoundness"),
+    clEnumValN(SmackWarnings::WarningLevel::Info,
+      "info", "Enable warnings about unsoundness and debug information"))
```
zvonimir: It is strange to call this `debug information`. In our recent discussion, you said that we are still using `DEBUG` macro for printing debug information. So this should be named different. What does `info` enable?

share/smack/top.py
```diff
@@ -110,6 +110,13 @@ def arguments():
   noise_group.add_argument('--debug-only', metavar='MODULES', default=None,
     type=str, help='limit debugging output to given MODULES')
 
+  noise_group.add_argument('--warn', default="unsound",
+    choices=['silent', 'unsound', 'info'],
+    help='''enable certain type of warning messages
+            (silent: no warning messages;
+            unsound: warnings about unsoundness;
+            info: warnings about unsoundness and debug information) [default: %(default)s]''')
```
zvonimir: Same comment as above about `debug information`. We already have a flag `--debug` that turns on debug information.

share/smack/top.py
```diff
@@ -322,7 +329,8 @@ def llvm_to_bpl(args):
   """Translate the LLVM bitcode file to a Boogie source file."""
 
   cmd = ['llvm2bpl', args.linked_bc_file, '-bpl', args.bpl_file]
-  cmd += ['-warnings']
+  cmd += ['-warn-type='+args.warn]
```
zvonimir: Add spaces before and after `+`.

lib/smack/SmackOptions.cpp
shaobo-he: It enables memcpy/memset length warning and collapsing DSA nodes. I'm not sure what is a good name for these.

share/smack/top.py
shaobo-he: Done.

lib/smack/SmackInstGenerator.cpp
shaobo-he: I fixed it.

lib/smack/SmackOptions.cpp
zvonimir: Maybe `translation information`. So `Enable warnings about unsoundness and translation information`. You can fix this below as well, and then maybe squash commits. Thanks!

PR #468, opened 2019-07-03
Minor fix of Prelude.cpp

shaobo-he:
This commit fixes the declarations of load/store functions as follows,

1. type-safe load/store functions should always be generated.
2. type-unsafe load/store functions are generated according to the pointer size.

### DIFF
--- a/test/bits/pointer_memops_fail.c
+++ b/test/bits/pointer_memops_fail.c
@@ -1,0 +1,18 @@
+#include "smack.h"
+#include <stdlib.h>
+
+// @expect error
+// @flag --clang-options=-m32 --bit-precise-pointers
+
+int* a[2];
+
+int main(void) {
+  int b = 1;
+  int** c = (int**)malloc(sizeof(int*));
+  a[0] = &b;
+  a[1] = &b;
+  *c = a[0];
+  assert(**c == 1);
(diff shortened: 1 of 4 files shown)


zvonimir:
How did you find out about this problem? Could you write a small regression that triggers it?

shaobo-he:
> How did you find out about this problem? Could you write a small regression that triggers it?

I found it when I tested the Prelude merge on SV-COMP benchmarks. I'm not sure why it is not triggered by our regressions.

shaobo-he:
I added two tests for this PR.

test/bits/pointer_memops_fail.c
```diff
@@ -0,0 +1,18 @@
+#include "smack.h"
+#include <stdlib.h>
+
+// @expect error
+// @flag --clang-options=-m32 --bit-precise-pointers
```
zvonimir: Why do you need `-m32`? Does it time out otherwise?

shaobo-he: No, this is because we have a switch between 32 bit and 64 bit points and I want to test the 32 bit version.

shaobo-he:
> Looks good to me. Please squash the commits.

Done.

PR #470, opened 2019-07-04
Improve the build system of SMACK

shaobo-he:
This PR intends to do the following two things,

- [x] Enable docker build
- [x] Use Ninja
- [x] Update Windows installation info
- [ ] Decouple the build process with the LLVM version used by SMACK


zvonimir:
Building using ninja is still experimental even in LLVM 9. So why would we switch to it so early?

shaobo-he:
> Building using ninja is still experimental even in LLVM 9. So why would we switch to it so early?

Ninja has been working well on SMACK so far. So we don't lose anything. Personally, I want to use it because it automatically utilizes all the available cores whereas I have to specify `-j` when using `make`. This can be very useful for building SMACK on Emulab, Vagrant VM, and Docker image.

zvonimir:
Ok, fair enough. But please test this on Mac OS as well.
It seems you have one more item to address in this pull request. Maybe add me as a reviewer only once you are done in the future.

shaobo-he:
> Ok, fair enough. But please test this on Mac OS as well.
> It seems you have one more item to address in this pull request. Maybe add me as a reviewer only once you are done in the future.

I'm not sure if the additional item is necessary. Basically I want to let users to use the existing C++ compiler rather than the specific version of Clang SMACK uses to generate IR files.

zvonimir:
> > Ok, fair enough. But please test this on Mac OS as well.
> > It seems you have one more item to address in this pull request. Maybe add me as a reviewer only once you are done in the future.
> 
> I'm not sure if the additional item is necessary. Basically I want to let users to use the existing C++ compiler rather than the specific version of Clang SMACK uses to generate IR files.

I see. Maybe you could clarify that in the documentation (https://github.com/smackers/smack/blob/master/docs/installation.md#general-system-requirements)?

Otherwise, this pull request looks good to me, except that I would like that you test it on a Mac OS machine to check whether ninja creates any problems there (maybe it is not installed by default or something like that, not sure).

shaobo-he:
@zvonimir I have tested both Ninja and the build process of the Docker image on Mac OS.

docs/installation.md
```diff
@@ -55,13 +55,35 @@ session, and halt, suspend, or destroy the virtual machine:
...
+### Quick Setup 2: Docker
+SMACK can also be run in a [Docker][] container. We tested the Dockerfile on
+the following configurations:
+
+* Ubuntu 16.04, docker-ce version 18.09.7
+* OS X 10.14.5, Docker Desktop with Docker engine version 18.09.2
+* Windows 10, Docker Desktop with Docker engine version 18.09.2
+
+Once Docker is successfully installed, build the Docker image by running the
+following command in SMACK's root directory which containes `Dockerfile`:
```
zvonimir: in the SMACK's root directory that contains `Dockerfile`

docs/installation.md
```diff
@@ -55,13 +55,35 @@ session, and halt, suspend, or destroy the virtual machine:
...
+```Shell
+docker build . -t smack
+```
+After the image is successuflly built, invoke a Docker container by running the
```
zvonimir: typo - successfully

docs/installation.md
```diff
@@ -55,13 +55,35 @@ session, and halt, suspend, or destroy the virtual machine:
...
+following command:
+```Shell
+docker run -it smack
+```
+For more advanced usages of Docker (e.g., mount host directories), please refer
```
zvonimir: to mount

docs/installation.md
```diff
@@ -55,13 +55,35 @@ session, and halt, suspend, or destroy the virtual machine:
...
+to Docker's official documents.
```
zvonimir: the Docker's official documentation

docs/installation.md
```diff
@@ -121,15 +144,22 @@ installers rather than built from source.
 and [Clang][] is problematic on some [Cygwin][] configurations. Please consult
 [LLVM][] documentation in case of any issues.
 
+#### Windows Subsystem for Linux (Recommended)
+
+SMACK can be installed on the Windows Subsystem for Linux (WSL) following the
```
zvonimir: by following

docs/installation.md
```diff
@@ -121,15 +144,22 @@ installers rather than built from source.
...
+same procedure of Linux installation (i.e., via the build script [build.sh][]).
```
zvonimir: procedure as the Linux installation

shaobo-he:
@zvonimir The installation instruction has been fixed.

shaobo-he:
> Looks good to me. Please squash and I'll merge it.

Done.

keram88:
Looks good to me. I tested the ninja build on Mac OS. llvm2bpl does build and run but I don't have boogie or corral to completely test it.

shaobo-he:
This is what I got from [hadolint](https://github.com/hadolint/hadolint),

```
Dockerfile:2 DL4000 MAINTAINER is deprecated
Dockerfile:6 DL3008 Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
Dockerfile:6 DL3009 Delete the apt-get lists after installing something
Dockerfile:6 DL3015 Avoid additional packages by specifying `--no-install-recommends`
Dockerfile:16 DL4006 Set the SHELL option -o pipefail before RUN with a pipe in it
Dockerfile:24 DL3020 Use COPY instead of ADD for files and folders
Dockerfile:30 DL3004 Do not use sudo as it leads to unpredictable behavior. Use a tool like gosu to enforce root
```

Not sure if we need to address them.

zvonimir:
I read more about ninja vs make comparisons. The main (only?) advantage of switching to ninja is to speed up the build process. Is that right? But we are adding a dependency, which sucks and I see it as a major drawback.
So why is the speed of building SMACK not satisfactory already? Even if I use just one core it builds pretty fast, and I would say it is not a bottleneck when it comes to booting emulab machines for example.

shaobo-he:
> I read more about ninja vs make comparisons. The main (only?) advantage of switching to ninja is to speed up the build process. Is that right? But we are adding a dependency, which sucks and I see it as a major drawback.
Why is it a drawback? Ninja can be easily installed either via as a Ubuntu package or a downloaded binary.

> So why is the speed of building SMACK not satisfactory already? Even if I use just one core it builds pretty fast, and I would say it is not a bottleneck when it comes to booting emulab machines for example.
On chimera (eight core i7-6700K), a fresh build using one core takes 1m34s while Ninja reduces it to 27s. The build time takes much longer on Emulab machines since their CPUs are much slower. The setup script on Emulab machines modifies `bin/build.sh`, which replaces `make` with `make -j64`. This is why it builds so fast.

PR #472, opened 2019-07-08
Clang format

zvonimir:
Using clang-format to enforce the LLVM coding style in SMACK.


shaobo-he:
Is it possible that we can split this PR to two PRs, one for adding clang-format to SMACK and the other for reformatting files? It's hard to review it because too many files are changed.

zvonimir:
I updated contribution guidelines already to note that we are using clang-format for formatting, which is super easy to use. I guess if there'll be issues I can further refine the guidelines.

PR #475, opened 2019-07-22
[WIP] Reformat the Python scripts

shaobo-he:
This PR intends to reformat SMACK's Python scripts following the outputs of pylint.


zvonimir:
Could we add checking into CI somehow, like what I did with clang-format?
Otherwise the code will soon again not be properly formatted.

zvonimir:
Ok, I am not sure any more about adding this into CI.
However, one thing I am definitely sure about is that I would like to stick with 2 spaces for indentation, and it seems that this switches it to 4 spaces.

shaobo-he:
> Ok, I am not sure any more about adding this into CI.
> However, one thing I am definitely sure about is that I would like to stick with 2 spaces for indentation, and it seems that this switches it to 4 spaces.

The 4 space indentation is specified by PEP 8 (https://www.python.org/dev/peps/pep-0008/#indentation). Personally I think we should follow this guide.

share/smack/top.py
```diff
@@ -680,21 +678,21 @@ def smackd_output(corral_output):
 
 def main():
     try:
-        global ARGS
-        ARGS = arguments()
+        global smack_args
```
zvonimir: Why would you replace more descriptive names (like `smack_args`) with less descriptive names (like just `ARGS`)?

share/smack/doctor.py
```diff
@@ -74,35 +74,35 @@ def check_headers(prefix):
...
 
     parser = argparse.ArgumentParser(description='Diagnose SMACK configuration issues.')
     parser.add_argument('-q', '--quiet', dest='quiet', action="store_true", default=False,
                         help='only show failed diagnostics')
     parser.add_argument('--prefix', metavar='P', dest='prefix', type=str, default='',
                         help='point to the installation prefix')
-    ARGS = parser.parse_args()
-    COUNT = 0
+    smack_doctor_args = parser.parse_args()
+    issue_count = 0
```
zvonimir: Why would you replace more descriptive names (like `issue_count`) with less descriptive names (like just `COUNT`)?
There are several global names that you truncated like that to make them harder to understand.

shaobo-he: I think your comments refer to old changes. The upper case names are removed.

zvonimir:
Please split this request into two separate pull requests: simple syntactic improvements and other improvements. I find it hard to review a pull request when they are mixed. Thx!

shaobo-he:
Close it for now.

Issue #53, opened 2014-10-19
Leverage llvm.assume intrinsic in SMACK


zvonimir:
http://llvm.org/docs/LangRef.html#llvm-assume-intrinsic

So our current assume should become something like:
void assume(int x) {
    __builtin_assume(x);
}

zvonimir:
I don't remember any more what I meant by this :). So I am closing it.

michael-emmi:
I think the idea was that LLVM optimizations could potentially make use of source-level assume statements. And taking advantage of that might be as simple as defining `__VERIFIER_assume` as `__builtin_assume` or whatever it’s called nowadays. So long as LLVM optimizations preserve these calls.

zvonimir:
Ah, ok, makes sense, thanks for the reminder. This should be pretty easy to try out on SVCOMP benchmarks and see if it makes a difference at all.

shaobo-he:
Fixed via #478

PR #478, opened 2019-08-01
[WIP] Optionally model the LLVM assume statement in Boogie

keram88:
Currently we don't do anything with LLVM's assume statement. This is because we
need to be sure that the assumption provided to the function is correct. At best
it might help with verification, but it could also lead SMACK to report erroneous
bugs (or no bugs!). For now, this removes this intrinsic, preventing a warning about an unsound
call to this function.

I could add the closure in the stmtMap, but I think it's valuable to have the rationale somewhere in this function.

### DIFF
--- a/lib/smack/SmackInstGenerator.cpp
+++ b/lib/smack/SmackInstGenerator.cpp
@@ -902,6 +902,28 @@ void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst &ii) {
         };
       };
 
+  // Optionally generate a boogie assume statement from assume statements in
+  // LLVM. Currently this behavior is experimental and must be enabled by
+  // passing the -llvm-assumes flag. The default behavior of this
+  // function is to ignore the assume statement, specified by the "none"
+  // argument. If the check argument is given, an additional assertion is
+  // generated to check the validity of the assumption.
+  static const auto assume = [this](CallInst *ci) {
+    if (SmackOptions::LLVMAssumes != LLVMAssumeType::none) {
+      auto arg = rep->expr(ci->getArgOperand(0));
+      auto llvmTrue =
+          SmackOptions::BitPrecise ? Expr::lit(1, 1) : Expr::lit(1LL);
+      auto chkStmt = Expr::eq(arg, llvmTrue);
+      if (SmackOptions::LLVMAssumes == LLVMAssumeType::check)
+        emit(Stmt::assert_(chkStmt));
+      else
+        emit(Stmt::assume(chkStmt));
+    } else {
+      // Skip assume statements
+      return;
+    }
+  };
+
   static const auto f16UpCast = conditionalModel(
       [this](CallInst *ci) {
         // translation: $f := $fpext.bvhalf.*($rmode, $bitcast.bv16.bvhalf($i));
@@ -1130,6 +1152,7 @@ void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst &ii) {
 
   static const std::map<llvm::Intrinsic::ID, std::function<void(CallInst *)>>
       stmtMap{
+          {llvm::Intrinsic::assume, assume},
           {llvm::Intrinsic::bitreverse, assignBvExpr(bitreverse)},
           {llvm::Intrinsic::bswap, assignBvExpr(bswap)},
           {llvm::Intrinsic::convert_from_fp16, f16UpCast},
--- a/lib/smack/SmackOptions.cpp
+++ b/lib/smack/SmackOptions.cpp
@@ -66,6 +66,17 @@ const llvm::cl::opt<bool>
 const llvm::cl::opt<bool> SmackOptions::IntegerOverflow(
     "integer-overflow", llvm::cl::desc("Enable integer overflow checks"));
 
+const llvm::cl::opt<LLVMAssumeType> SmackOptions::LLVMAssumes(
+    "llvm-assumes",
+    llvm::cl::desc(
+        "Optionally enable generation of Boogie assumes from LLVM assumes"),
+    llvm::cl::values(clEnumValN(LLVMAssumeType::none, "none",
+                                "disable generation of assume statements"),
+                     clEnumValN(LLVMAssumeType::use, "use",
+                                "enable generation of assume statements"),
+                     clEnumValN(LLVMAssumeType::check, "check",
+                                "enable checking of assume statements")));
+
 bool SmackOptions::isEntryPoint(std::string name) {
   for (auto EP : EntryPoints)
     if (name == EP)
--- a/share/smack/top.py
+++ b/share/smack/top.py
@@ -197,6 +197,10 @@ def arguments():
   translate_group.add_argument('--integer-overflow', action='store_true', default=False,
     help='enable integer overflow checks')
 
+  translate_group.add_argument('--llvm-assumes', choices=['none', 'use', 'check'], default='none',
+    help='optionally enable generation of Boogie assume statements from LLVM assume statements ' +
+         '(none=no generation [default], use=generate assume statements, check=check assume statements)')
+
   translate_group.add_argument('--float', action="store_true", default=False,
     help='enable bit-precise floating-point functions')
 
@@ -347,6 +351,7 @@ def llvm_to_bpl(args):
   if args.no_memory_splitting: cmd += ['-no-memory-splitting']
   if args.memory_safety: cmd += ['-memory-safety']
   if args.integer_overflow: cmd += ['-integer-overflow']
+  if args.llvm_assumes: cmd += ['-llvm-assumes=' + args.llvm_assumes]
   if args.float: cmd += ['-float']
   if args.modular: cmd += ['-modular']
   try_command(cmd, console=True)
--- a/test/special/assume.c
+++ b/test/special/assume.c
@@ -1,0 +1,12 @@
+#include "smack.h"
+
+// @expect verified
+// @flag --llvm-assumes=use
+
+int main(void) {
+  unsigned int y = 2 * (unsigned int)__VERIFIER_nondet_unsigned_short();
+  // This assumption is used for verification, even though bit-precise
+  // is not enabled, the assertion will pass.
+  __builtin_assume((y & 1) == 0);
+  __VERIFIER_assert((y & 1) == 0);
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ env:
     - TRAVIS_ENV="--exhaustive --folder=memory-safety"
     - TRAVIS_ENV="--exhaustive --folder=pthread"
     - TRAVIS_ENV="--exhaustive --folder=strings"
+    - TRAVIS_ENV="--exhaustive --folder=special"
 
 before_install:
   - sudo rm -rf /usr/local/clang-7.0.0
--- a/include/smack/SmackOptions.h
+++ b/include/smack/SmackOptions.h
@@ -10,6 +10,8 @@
 #include "smack/SmackWarnings.h"
 
 namespace smack {
+enum class LLVMAssumeType { none, use, check };
+
 class SmackOptions {
 public:
   static const llvm::cl::list<std::string> EntryPoints;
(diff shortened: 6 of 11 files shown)


keram88:
My primary impetus to add this is that Rust generates these in just about every program.

keram88:
Travis is failing this because of `clang-format` something in `lib/smack/Regions.cpp`. As far as I can tell the failing code is removed in LLVM 7.1. Regardless, this PR is unrelated to this file, and all regressions pass.

michael-emmi:
I’m curious about knowing more; is there a reference on this?

Overall, I wonder whether it might instead be a good idea to take advantage of these assumptions by modeling them as boogie assumes. Depending on why they’re added. For instance, if they correspond to things the rust type checker has already proven, then why not use them?

keram88:
@michael-emmi That's a good point, I should have discussed how Rust uses this function.

Rust doesn't appear to do anything interesting with `llvm.assume`. In all the Rust benchmarks, the uses of `llvm.assume` follow this pattern:
```
i1 x = (some boolean expression);
llvm.assume(x <= 1);
```
Basically Rust is telling LLVM its booleans are `0` or `1`, but LLVM should know this since their type is `i1`. 

The strongest thing I think Rust has ever done with respect to its typechecker is add the `noalias` attribute to function arguments. This issue goes into more detail about some of the problems with this: https://github.com/rust-lang/rust/issues/53105.

Another approach is to make generating Boogie assumptions opt-in or opt-out through a command line argument. Something like `--use-llvm-assume` or `--no-llvm-assume`. @shaobo-he @zvonimir What do you think about this?

We should update #53 regardless.

shaobo-he:
> @michael-emmi That's a good point, I should have discussed how Rust uses this function.
> 
> Rust doesn't appear to do anything interesting with `llvm.assume`. In all the Rust benchmarks, the uses of `llvm.assume` follow this pattern:
> 
> ```
> i1 x = (some boolean expression);
> llvm.assume(x <= 1);
> ```
> 
> Basically, Rust is telling LLVM its booleans are `0` or `1`, but LLVM should know this since their type is `i1`.
> 
> The strongest thing I think Rust has ever done with respect to its type-system is add the `noalias` attribute to function arguments. This issue goes into more detail about some of the problems with this: [rust-lang/rust#53105](https://github.com/rust-lang/rust/issues/53105).
> 
> Another approach is to make generating Boogie assumptions opt-in or opt-out through a command line argument. Something like `--use-llvm-assume` or `--no-llvm-assume`. @shaobo-he @zvonimir What do you think about this?
> 
> We should update #53 regardless.

I think without evident benefits of leveraging the LLVM assume intrinsic, the current PR is the better option.

zvonimir:
Please fix formatting so that CI is passing. It is an easy fix. Also, please address this compilation warning if possible:
```
../lib/smack/SmackInstGenerator.cpp:896:31: warning: lambda capture 'this' is not used [-Wunused-lambda-capture]
1984  static const auto assume = [this](CallInst *ci) { return; };
```

zvonimir:
I see. Formatting got broken elsewhere. I'll fix it. You still have to take care of the warning though.

michael-emmi:
Could we at least understand whether the assumes here represent facts that have been “proven” by previous analysis, or whether e.g., they represent speculative optimization hints that may be invalidated and require unoptimized fallback?

zvonimir:
From the LLVM language spec (https://llvm.org/docs/LangRef.html#llvm-assume-intrinsic):
```
The llvm.assume allows the optimizer to assume that the provided condition is true. This information can then be used in simplifying other parts of the code.
```
and also
```
If the condition is violated during execution, the behavior is undefined.
```
So to me it seems that the purpose of `assume` intrinsics is to provide "proven" facts.

I think we should ultimately have a command line switch `--leverage-llvm-assumes` that turns on using them for verification, and we drop them otherwise. Then we can experiment with this more.

keram88:
From https://llvm.org/docs/LangRef.html#llvm-assume-intrinsic:
```
The intrinsic allows the optimizer to assume that the provided condition is
always true whenever the control flow reaches the intrinsic call. No code is
generated for this intrinsic, and instructions that contribute only to the
provided condition are not used for code generation. If the condition is
violated during execution, the behavior is undefined.

Note that the optimizer might limit the transformations performed on values used
by the llvm.assume intrinsic in order to preserve the instructions only used to
form the intrinsic’s input argument. This might prove undesirable if the extra
information provided by the llvm.assume intrinsic does not cause sufficient
overall improvement in code quality. For this reason, llvm.assume should not be
used to document basic mathematical invariants that the optimizer can otherwise
deduce or facts that are of little use to the optimizer.
```
I think the uses of `llvm.assume` by Rust I gave earlier are of the kind that LLVM says not to generate, but this doesn't affect the correctness of SMACK. One of the most powerful uses of the `assume` statement is documentation of loop invariants from the compiler. LLVM probably can't discover these facts due to the loss of semantic information. This information is likely very helpful for SMACK.

I'm starting to fall on the side of generating boogie assumes for these; they can't hurt verification time and might help. My biggest concern is that the compiler will generate an `assume` which is false. However, we already assume that the compiler reasonably transforms the source. If erroneous assumes were being routinely generated, then incorrect optimizations in LLVM would occur and compiler bugs would be reported. My other consideration is that the compiler provides a correct assumption, but an undefined behavior invalidates it. I'm not too worried about this because the program is already ill-formed, and SMACK has an expanding set of checks for undefined behaviors to help diagnose this.

keram88:
> So to me it seems that the purpose of `assume` intrinsics is to provide "proven" facts.
> 
> I think we should ultimately have a command line switch `--leverage-llvm-assumes` that turns on using them for verification, and we drop them otherwise. Then we can experiment with this more.

This seems reasonable. We should try generating assumes to see how they behave before enabling or disabling them generally. I want to update this PR with your proposal.

michael-emmi:
Yea, I agree that experimenting with whether these generated `llvm.assume`s provide useful information to the verifier is the right thing to do. My hunch would also be that they could help significantly, e.g., if they carried higher-level program invariants from, e.g., `rustc` to LLVM.

One small note: it’s not obvious to me that additional assumes “can’t hurt verification time”, but perhaps that is another conversation altogether :-)

keram88:
> One small note: it’s not obvious to me that additional assumes “can’t hurt verification time”, but perhaps that is another conversation altogether :-)

This is worth investigating. I know some of our floating-point axioms slowed down verification. Solvers are complex beasts and it's hard to know what hurts or helps.

keram88:
I implemented Zvonimir's suggestion. I witnessed one interesting case along the lines of this (translated from Rust):
```
int x = __VERIFIER_nondet_int();
int y = 2*x;
...
int z = y&1;
...
__VERIFIER_assert(z == 1);
```
SMACK is able to produce an error here with `--leverage-llvm-assumes` despite `--bit-precise` being off.

Rust also puts in a useful assume statement, namely that its pointers are not null.

keram88:
What do you think of also adding a "strict" assume mode? Clang has `__builtin_assume`:
https://clang.llvm.org/docs/LanguageExtensions.html#builtin-assume
This seems to always emit an `@llvm.assume` statement. We could emit an assert to check that the assumption always holds.

zvonimir:
I see. So there should be a command line argument with 3 options: ignore assumes, assume assumes, and assert assumes. I like that. Please add the assert assumes option as well then. Good suggestion.

zvonimir:
Also, could you create a regression that checks for this?

zvonimir:
@keram88 : What's the status of this pull request? Any chance you could finish it this week?

keram88:
I have implemented the following:
* `--llvm-assumes=none`: This ignores LLVM assume statements as SMACK currently does, and is the default behavior.
* `--llvm-assumes=use`: This uses LLVM assumes as boogie assumes.
* `--llvm-assumes=check`: This uses LLVM assumes and additionally checks their validity.

I added the regressions here: https://github.com/smackers/smack/tree/no-warn-assume/test/special.
I wasn't sure where to put them since they are kind of weird tests. I think it covers the range of odd behaviors that could be verified using `__builtin_assume`, such as confirming that `1=0`.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -904,6 +904,27 @@ void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst &ii) {
...
+  // argument. If the check argument is given, an additional assertion is
+  // generated to check the validity of the assumption.
+  static const auto assume = [this](CallInst *ci) {
+    if (SmackOptions::LLVMAssumes != LLVMAssumeType::none) {
+      auto arg = rep->expr(ci->getArgOperand(0));
+      auto llvmTrue =
+          SmackOptions::BitPrecise ? Expr::lit(1, 1) : Expr::lit(1LL);
+      auto chkStmt = Expr::eq(arg, llvmTrue);
+      if (SmackOptions::LLVMAssumes == LLVMAssumeType::check)
+        emit(Stmt::assert_(chkStmt));
```
zvonimir: There should be `else` here. I understand what you wanted to do there - assume it if it gets asserted - but there is no need for that.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -904,6 +904,27 @@ void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst &ii) {
...
+      emit(Stmt::assume(chkStmt));
+    } else {
+      // Empty model
```
zvonimir: Change this comment into `Skip assume statements`. I think that would make it more descriptive.

lib/smack/SmackOptions.cpp
```diff
@@ -66,6 +66,18 @@ const llvm::cl::opt<bool>
...
+    "llvm-assumes",
+    llvm::cl::desc(
+        "Optionally enable generation of Boogie assumes from LLVM assumes"),
+    llvm::cl::values(
+        clEnumValN(LLVMAssumeType::none, "none",
+                   "disable generation of assume statements"),
+        clEnumValN(LLVMAssumeType::use, "use",
+                   "enable generation of assume statements"),
+        clEnumValN(LLVMAssumeType::check, "check",
+                   "enable checking of generated assume statments")));
```
zvonimir: You have a typo here. Also, change this into `enable checking of assume statements`.

share/smack/top.py
```diff
@@ -197,6 +197,9 @@ def arguments():
   translate_group.add_argument('--integer-overflow', action='store_true', default=False,
     help='enable integer overflow checks')
 
+  translate_group.add_argument('--llvm-assumes', choices=['none', 'use', 'check'], default='none',
+    help='optionally enable generation of Boogie assume statements from LLVM assume statements (none=no generation [default], use=generate assume statements, check=check validity of assume statemets)')
```
zvonimir: You have a typo here. Also, change the last option into `check assume statements`.

share/smack/top.py
```diff
@@ -347,6 +350,7 @@ def llvm_to_bpl(args):
   if args.no_memory_splitting: cmd += ['-no-memory-splitting']
   if args.memory_safety: cmd += ['-memory-safety']
   if args.integer_overflow: cmd += ['-integer-overflow']
+  if args.llvm_assumes: cmd += ['-llvm-assumes='+args.llvm_assumes]
```
zvonimir: Put space before and after `+`.

test/special/assume.c
```diff
@@ -0,0 +1,12 @@
+#include "smack.h"
```
zvonimir: You should add these new regressions into our travis CI script. That should be easy to do.

share/smack/top.py
keram88: Done.

share/smack/top.py
keram88: Done.

keram88:
I addressed @zvonimir's comments. I don't like force pushing, so if this looks good, I can squash the commits.

lib/smack/SmackInstGenerator.cpp

test/special/assume.c
keram88: Updated .travis.yml.

keram88:
Done. Thanks, @michael-emmi for driving this toward what I think is a much better solution.

PR #479, opened 2019-08-02
CVC4 support

liammachado:
This adds basic support for CVC4. It can be used instead of Z3 by adding the flag `--verifier-options=/bopt:proverOpt:SOLVER=cvc4` when using Corral, and `--verifier-options=/proverOpt:SOLVER=cvc4` when using Boogie. Adding support for CVC4 required making the "srem" and "urem" functions uninterpreted, as CVC4 does not currently support a "rem" function for integers. Additionally, CVC4 supports "bv2nat" in place of "bv2int".

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -23,6 +23,7 @@
 # Set these flags to control various installation options
 INSTALL_DEPENDENCIES=1
 INSTALL_Z3=1
+INSTALL_CVC4=1
 BUILD_BOOGIE=1
 BUILD_CORRAL=1
 BUILD_SYMBOOGLIX=1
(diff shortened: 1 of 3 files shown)


michael-emmi:
Nice! 👍 

Any plans to include CVC4 in the regression tests?

zvonimir:
@liammachado this looks good to me. I think you should add one more thing though. Please add `--solver` command line option to SMACK python script (similar to `--verifier`). The available options should be `z3` and `cvc4`, and `z3` should be the default one. Then users would not have to worry about manually setting verifier options.

I also do wonder if we should be running all regressions with CVC4 as well, if they are passing already. We would need to split it for travis, but that would not be too hard. I think we should maybe first merge this one in though, and then think about regressions.

PR #482, opened 2019-08-09
Remove duplicate conditional AST node

keram88:
This removes the `CondExpr` class in favor of the more descriptively named `IfThenElseExpr` class. The printing of `IfThenElseExpr` now encloses the expression in parentheses to prevent precedence issues.

### DIFF
--- a/lib/smack/BoogieAst.cpp
+++ b/lib/smack/BoogieAst.cpp
@@ -28,16 +28,16 @@ const Expr *Expr::or_(const Expr *l, const Expr *r) {
   return new BinExpr(BinExpr::Or, l, r);
 }
 
-const Expr *Expr::cond(const Expr *c, const Expr *t, const Expr *e) {
-  return new CondExpr(c, t, e);
-}
-
 const Expr *Expr::eq(const Expr *l, const Expr *r) {
   return new BinExpr(BinExpr::Eq, l, r);
 }
 
 const Expr *Expr::lt(const Expr *l, const Expr *r) {
   return new BinExpr(BinExpr::Lt, l, r);
+}
+
+const Expr *Expr::ifThenElse(const Expr *c, const Expr *t, const Expr *e) {
+  return new IfThenElseExpr(c, t, e);
 }
 
 const Expr *Expr::fn(std::string f, std::list<const Expr *> args) {
@@ -113,10 +113,6 @@ const Expr *Expr::sel(std::string b, std::string i) {
 
 const Expr *Expr::upd(const Expr *b, const Expr *i, const Expr *v) {
   return new UpdExpr(b, i, v);
-}
-
-const Expr *Expr::if_then_else(const Expr *c, const Expr *t, const Expr *e) {
-  return new IfThenElseExpr(c, t, e);
 }
 
 const Expr *Expr::bvExtract(const Expr *v, const Expr *u, const Expr *l) {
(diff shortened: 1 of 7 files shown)


keram88:
Do we want to remove the `cond` or `if_then_else` methods or leave them as is?

lib/smack/BoogieAst.cpp
```diff
@@ -510,7 +506,8 @@ void CodeExpr::print(std::ostream &os) const {
 }
 
 void IfThenElseExpr::print(std::ostream &os) const {
-  os << "if " << cond << " then " << true_value << " else " << false_value;
+  os << "(if " << cond << " then " << true_value << " else " << false_value
```
zvonimir: Rename `true_value` and `false_value` into `trueValue` and `falseValue`.

lib/smack/BoogieAst.cpp
```diff
@@ -29,7 +29,7 @@ const Expr *Expr::or_(const Expr *l, const Expr *r) {
 }
 
 const Expr *Expr::cond(const Expr *c, const Expr *t, const Expr *e) {
-  return new CondExpr(c, t, e);
+  return if_then_else(c, t, e);
```
zvonimir: Rename `if_then_else` into `ifThenElse`.

zvonimir: I also think we should rename this `cond` generator into `ifThenElse` to make it consistent.

zvonimir:
> Do we want to remove the `cond` or `if_then_else` methods or leave them as is?

Could you please clarify this question of yours. I don't quite get it.

zvonimir:
I think once you address my feedback we'll be ready to merge it in. Squash all the commits and write a nice commit message. Make sure to use `Fixes` to refer to the issue you are fixing. Thanks @keram88!

lib/smack/BoogieAst.cpp
keram88: I removed `cond` in favor of `ifThenElse`. They are semantically equivalent now, and there's no reason to keep both around.

keram88:
> > Do we want to remove the `cond` or `if_then_else` methods or leave them as is?
> 
> Could you please clarify this question of yours. I don't quite get it.

I meant that we had `cond` and `if_then_else` generators, they both produced an `IfThenElseExpr` node, and with the change to `IfThenElseExpr` printing, they were equivalent. I was asking if we wanted to keep both around, but you suggested we rename `cond`, so that answered my question. `ifThenElse` is now the only generator for conditional expressions.

I'm going to let the CI check clang-format since I got some nonsense locally, then I'll squash.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -1030,11 +1031,11 @@ void SmackInstGenerator::visitIntrinsicInst(llvm::IntrinsicInst &ii) {
...
         // is zero, then the result is undefined.
         auto isZeroUndef = rep->expr(ci->getArgOperand(1));
-        body = Expr::if_then_else(
-            Expr::and_(Expr::eq(isZeroUndef, Expr::lit(1, 1)),
-                       Expr::eq(var, Expr::lit(0, width))),
-            rep->expr(ci), // The result is undefined
-            body);
+        body =
+            Expr::ifThenElse(Expr::and_(Expr::eq(isZeroUndef, Expr::lit(1, 1)),
+                                        Expr::eq(var, Expr::lit(0, width))),
```
keram88: Whitespace is to make clang-format happy.

zvonimir:
Yo, please squash the commits and then I'll merge. Thx!

keram88:
Done.

PR #334, opened 2018-06-29
Add multi-language documentation

jjgarzella:
Update the README to reflect that SMACK now supports
many languages

Add a short tutorial on how to compile a new language
into SMACK.

PR #491, opened 2019-10-22
Upgrade Rust support for LLVM 8

### LINKED ISSUE #489: Cleaning up after Rust installation
It seems that our build script (`build.sh`) leaves the following in the root folder after installing Rust:
```
rust-nightly-x86_64-unknown-linux-gnu
rust.tar.gz
```
I wonder if those are really needed?
If Rust is being installed system-wide, then maybe it would be a good idea to clean this up by deleting these after the Rust installation is done.

keram88:
This updates Rust support for the version of LLVM that SMACK uses.
* Install the most recent Rust compiler to support LLVM 8
* Improve entry point detection
* Streamline Rust compiler installation. Closes #489


keram88:
I plan to document how to use Smack on Rust programs, possibly as an update to #334.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -646,10 +646,8 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst &ci) {
   } else if (name.find(Naming::RUST_ENTRY) != std::string::npos) {
     // Set the entry point for Rust programs
     auto castExpr = ci.getArgOperand(0);
-    if (auto CE = dyn_cast<const Constant>(castExpr)) {
-      auto mainFunc = CE->getOperand(0);
-      emit(Stmt::call(mainFunc->getName(), {}, {}));
-    }
+    auto mainFunction = dyn_cast<const Function>(castExpr);
```
shaobo-he: Why not use `cast<>` if we don't check for possible nullptr?

zvonimir: Yes, this could be changed to `cast`.

PR #493, opened 2019-10-23
Remove redundant work from integer overflow checker

keram88:
Don't create a bound check for zero for unsigned integers.


keram88:
This isn't ready for merging yet. This is from an ancient branch and I want to see what its status is.

keram88:
The first else corresponds to isSigned. The second else depends on if integer overflow checking is enabled.

keram88:
Actually I think this is wrong. The check on the lower bound should still happen on subtraction. I knew there was a reason I didn't do a PR last year.

PR #494, opened 2019-10-23
Add Rust regressions to test directory

keram88:


.travis.yml
```diff
@@ -37,6 +37,13 @@ env:
     - TRAVIS_ENV="--exhaustive --folder=c/pthread"
     - TRAVIS_ENV="--exhaustive --folder=c/strings"
     - TRAVIS_ENV="--exhaustive --folder=c/special"
+    - TRAVIS_ENV="--exhaustive --folder=rust/basic --languages rust"
```
zvonimir: Minor point, but to make these consistent, please change this to `--languages=rust`.

test/regtest.py
```diff
@@ -20,6 +20,9 @@
 OVERRIDE_FIELDS = ['verifiers', 'memory', 'time-limit', 'memory-limit', 'skip']
 APPEND_FIELDS = ['flags', 'checkbpl', 'checkout']
 
+LANGUAGES = {'c': {'*.c', '*.cpp'},
```
zvonimir: Add `cplusplus` as a separate language since it has its own subfolder already.

test/regtest.py
```diff
@@ -117,6 +120,7 @@ def process_test(cmd, test, memory, verifier, expect, checkbpl, checkout, log_fi
   t0 = time.time()
   p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
   out, err  = p.communicate()
+
```
zvonimir: No need for this new line.

test/regtest.py
```diff
@@ -186,12 +207,17 @@ def main():
                       help="sets the logging level (DEBUG, INFO, WARNING)")
   parser.add_argument("--output-log", action="store", dest="log_path", type=str,
                       help="sets the output log path. (std out by default)")
+  parser.add_argument("--languages", action="store", default="c",
```
zvonimir: This should be implemented using `choices`: https://docs.python.org/2/library/argparse.html#choices
Then you don't have to do the manual checking of this argument above.

zvonimir: See this one for how to do multiple choices: https://stackoverflow.com/questions/44342813/multiple-choices-command-line-argument

test/rust/basic/add_fail.rs
```diff
@@ -0,0 +1,9 @@
+#[macro_use]
+mod smack;
+use smack::*;
+// @expect error
```
zvonimir: To make these consistent with C regressions, leave an empty line before and after `@expect` comment and other flags.

test/rust/basic/add_overflow.rs
```diff
@@ -0,0 +1,11 @@
+#[macro_use]
+mod smack;
+use smack::*;
+// @flag --integer-overflow
```
zvonimir: Add empty line before this line. Make all of the regressions consistent like that.

PR #499, opened 2019-10-26
Model C integer remainder operations

### LINKED ISSUE #496: Re-support remainder operations
The CVC support PR made remainder functions uninterpreted, which causes a number of false alarms in SV-COMP benchmarks. It'd be better that we add the support again. Since `rem` is not defined in the SMT lib, we can express it using the existing operations like `mod` (see https://cs.nyu.edu/pipermail/smt-lib/2014/000823.html).

shaobo-he:
Fixes #496

### DIFF
--- a/lib/smack/Prelude.cpp
+++ b/lib/smack/Prelude.cpp
@@ -358,6 +358,44 @@ struct IntOpGen::IntArithOp : public IntOp {
     auto pred = MIN ? Expr::lt(a1, a2) : Expr::lt(a2, a1);
     return Expr::ifThenElse(pred, a1, a2);
   }
+
+  // generate inlined `srem(i1, i2)` function body like
+  // if (mod(i1,i2) != 0 && i1 < 0)
+  // then mod(i1,i2) - abs(i2) else mod(i1,i2), where
+  // `i1` is the dividend and `i2` is the divisor
+  // `mod` is the Euclidean function defined in the SMT lib
+  // therefore its result is always positive
+  // the `srem` operation defined by LLVM follows the same
+  // definition as modern C standards
+  // (https://en.wikipedia.org/wiki/Modulo_operation#In_programming_languages)
+  // its result has the same sign as i1 because the division rounds to 0
+  // therefore, the only case `mod` and `srem` differ is when i1 is negative
+  // and the remainder is not 0
+  // for this case, the result of `srem` is the result of `mod` minus |i2|
+  static const Expr *sremExpr(unsigned size) {
+    std::string type = getIntTypeName(size);
+    const Expr *dividend = makeIntVarExpr(1);
+    const Expr *divisor = makeIntVarExpr(2);
+    const Expr *zero = Expr::lit((unsigned long long)0);
+    const Expr *mod = Expr::fn(indexedName("$smod", {type}), dividend, divisor);
+    const Expr *modNeZero =
+        Expr::fn(indexedName("$ne", {type, "bool"}), mod, zero);
+    const Expr *dividendLtZero =
+        Expr::fn(indexedName("$slt", {type, "bool"}), dividend, zero);
+    const Expr *negRemainder = Expr::fn(
+        indexedName("$sub", {type}), mod,
+        Expr::fn(indexedName("$smax", {type}), divisor,
+                 Expr::fn(indexedName("$sub", {type}), zero, divisor)));
+    return Expr::ifThenElse(Expr::and_(modNeZero, dividendLtZero), negRemainder,
+                            mod);
+  }
+
+  // generate inlined `urem` function body like
+  // $smod.i32(i1,i2), where `$smod` is a wrapper to SMT's `mod` function
+  static const Expr *uremExpr(unsigned size) {
+    return Expr::fn(indexedName("$smod", {getIntTypeName(size)}),
+                    makeIntVarExpr(1), makeIntVarExpr(2));
+  }
 };
 
 void IntOpGen::generateArithOps(std::stringstream &s) const {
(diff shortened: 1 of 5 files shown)


lib/smack/Prelude.cpp
```diff
@@ -358,6 +358,44 @@ struct IntOpGen::IntArithOp : public IntOp {
...
+  // (https://en.wikipedia.org/wiki/Modulo_operation#In_programming_languages)
+  // its result has the same sign as i1 because the division rounds to 0
+  // therefore, the only case `mod` and `srem` differ is when i1 is negative
+  // and the remainder is not 0
+  // for this case, the result of `srem` is the result of `mod` minus |i2|
+  static const Expr *sremExpr(unsigned size) {
+    std::string type = getIntTypeName(size);
+    const Expr *a1 = makeIntVarExpr(1);
+    const Expr *a2 = makeIntVarExpr(2);
+    const Expr *zero = Expr::id("0");
```
zvonimir: This is a weird way to create an integer constant. Why are you doing it like that (meaning, why are you creating a variable instead of constant)?

lib/smack/Prelude.cpp
```diff
@@ -358,6 +358,44 @@ struct IntOpGen::IntArithOp : public IntOp {
...
+  // the `srem` operation defined by LLVM follows the same
+  // definition as modern C standards
+  // (https://en.wikipedia.org/wiki/Modulo_operation#In_programming_languages)
+  // its result has the same sign as i1 because the division rounds to 0
+  // therefore, the only case `mod` and `srem` differ is when i1 is negative
+  // and the remainder is not 0
+  // for this case, the result of `srem` is the result of `mod` minus |i2|
+  static const Expr *sremExpr(unsigned size) {
+    std::string type = getIntTypeName(size);
+    const Expr *a1 = makeIntVarExpr(1);
```
zvonimir: Why don't you call these pointer expressions `i1` and `i2` to match your comments?

lib/smack/Prelude.cpp
```diff
@@ -358,6 +358,44 @@ struct IntOpGen::IntArithOp : public IntOp {
...
+    const Expr *a2 = makeIntVarExpr(2);
+    const Expr *zero = Expr::id("0");
+    const Expr *mod = Expr::fn(indexedName("$smod", {type}), a1, a2);
+    const Expr *ne0 = Expr::fn(indexedName("$ne", {type}), mod, zero);
+    const Expr *lt0 = Expr::fn(indexedName("$slt", {type}), a1, zero);
```
zvonimir: Why don't you use `slt.type.bool` here, so that below you can use Boogie's `&&` operator?

lib/smack/Prelude.cpp
```diff
@@ -358,6 +358,44 @@ struct IntOpGen::IntArithOp : public IntOp {
...
+    const Expr *cond = Expr::fn(
+        indexedName("$ne", {getIntTypeName(1), "bool"}),
+        Expr::fn(indexedName("$and", {getIntTypeName(1)}), ne0, lt0), zero);
+    const Expr *thenExpr =
+        Expr::fn(indexedName("$sub", {type}), mod,
+                 Expr::fn(indexedName("$smax", {type}), a2,
+                          Expr::fn(indexedName("$sub", {type}), zero, a2)));
+    const Expr *elseExpr = mod;
```
zvonimir: No need to introduce the `elseExpr` variable here.

test/c/basic/srem.c
```diff
@@ -0,0 +1,46 @@
+#include "smack.h"
+
+// @expect verified
+
+int main(void) {
+  int x;
+  x = 62951;
+  assert(x % (-37) == 14);
```
zvonimir: Clang often does constant propagation and then evaluates expressions where it can. Maybe it does that even with `O0`. Could you check if the generated Boogie actually contains these remainder operations just to make sure they are actually being generated?

lib/smack/Prelude.cpp
shaobo-he: I think I need to refresh my memory with Boogie AST APIs. Now it's changed to a better form.

lib/smack/Prelude.cpp
shaobo-he: I renamed them to a better name.

lib/smack/Prelude.cpp
```diff
@@ -358,6 +358,43 @@ struct IntOpGen::IntArithOp : public IntOp {
...
+  // therefore, the only case `mod` and `srem` differ is when i1 is negative
+  // and the remainder is not 0
+  // for this case, the result of `srem` is the result of `mod` minus |i2|
+  static const Expr *sremExpr(unsigned size) {
+    std::string type = getIntTypeName(size);
+    const Expr *dividend = makeIntVarExpr(1);
+    const Expr *divisor = makeIntVarExpr(2);
+    const Expr *zero = Expr::lit((unsigned long long)0);
+    const Expr *mod = Expr::fn(indexedName("$smod", {type}), dividend, divisor);
+    const Expr *modNe0 =
```
zvonimir: I don't particularly like `0` in the variable name. Please use `Zero` instead.

lib/smack/Prelude.cpp
```diff
@@ -358,6 +358,43 @@ struct IntOpGen::IntArithOp : public IntOp {
...
+        Expr::fn(indexedName("$ne", {type, "bool"}), mod, zero);
+    const Expr *dividendLt0 =
```
zvonimir: I don't particularly like `0` in the variable name. Please use `Zero` instead.

test/c/basic/srem.c
```diff
@@ -0,0 +1,47 @@
+#include "smack.h"
+
+// @expect verified
+// @checkbpl grep ':= \$srem'
```
zvonimir: Wait, isn't this always going to be true since there is always `$srem` in the prelude?
Am I missing something here?

shaobo-he: No, this pattern matches assignments that have `$srem` in the right-hand side.

shaobo-he:
> Looks good to me. Please squash and then I'll merge it.

Done. Please merge it once the CI passes.

michael-emmi:
Looks like I’m coming late to this weekend-work party (!) but I am wondering why the current approach would be preferable to having per-prover definitions, i.e., different definitions depending on whether CVC or Z3 is selected as the prover.

zvonimir:
My take on this is that if there is no difference in performance (we still have to check this on SVCOMP benchmarks) then it is nice to be SMTLIB-compliant and not case-split on provers in SMACK.

PR #501, opened 2019-10-28
Inline `isExternal`

### LINKED ISSUE #500: Failing floating-point regressions when CVC4 is used

shaobo-he:
Partially fix #500


zvonimir:
Why does this change anything?

shaobo-he:
> Why does this change anything?

It seems Boogie will add some Z3-specific attributes when dealing with functions that have a body but are not inlined. These attributes are not supported by CVC4.

zvonimir:
I see, makes sense. If I remember correctly, functions with bodies that don't have `inline` attribute are turned into quantified axioms, which we want to avoid anyways.
Please write a more elaborate commit message describing what you just wrote above, and then force-push I guess. Then I will merge this.

shaobo-he:
> I see, makes sense. If I remember correctly, functions with bodies that don't have `inline` attribute are turned into quantified axioms, which we want to avoid anyways.
> Please write a more elaborate commit message describing what you just wrote above, and then force-push I guess. Then I will merge this.

Done.

PR #513, opened 2019-11-18
Upgraded to Python 3

zvonimir:
This mostly boiled down to simple syntax updates.
I also updated our installation documentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ addons:
     packages:
       - git
       - cmake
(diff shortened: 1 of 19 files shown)


zvonimir:
@michael-emmi it would be helpful if you could test this branch on MacOS to see if this broke anything. Thanks!

michael-emmi:
Just adding a small fix to that `./test/regrest.py` doesn’t have to be run from the `test` directory.

PR #519, opened 2019-12-10
Modular deps

### LINKED ISSUE #518: Modularize dependencies.
Currently Smack is fairly heavy-handed when it comes to installation: besides requiring install from source (see #516), Smack insists on an odd anti-convention for its Boogie and Corral dependencies: installing its own `boogie` and `corral` scripts into the path which refer to `$BOOGIE` and `$CORRAL` environment variables which must be set.

Besides going against conventions — `boogie` and `corral` should just show up in the `$PATH` — this makes packaging (see #516) inconvenient. Ideally, we would declare package dependencies on Boogie and Corral, allowing the usual modularity afforded by package management systems.

Of course, currently there is only one release of Boogie and Corral packages, respectively:
* https://www.nuget.org/packages/Boogie/2.4.1-netcore-alpha
* https://www.nuget.org/packages/Corral/2.4.1-netcore-alpha

but versioning and release automation is in the works.

Ultimately, we could declare dependencies on these packages.

As a first step, I propose we adopt the usual convention of assuming `boogie` and `corral` are in the `$PATH`, rather than doing this funky environment variable and script gymnastics.

michael-emmi:
Fixes #518 by avoiding installation of `boogie` and `corral` scripts, preferring the use of Boogie/Corral’s proper scripts instead.

Since this relies on boogie-org/corral#103 we will have to wait for that to merge, and update our corral version in `bin/versions`.


michael-emmi:
Extending to lockpwn and Symbooglix waiting on these PRs:
smackers/lockpwn#2
boogie-org/symbooglix#35

michael-emmi:
@zvonimir are you for or against this change?

michael-emmi:
> Installation documentation should be updated to reflect these changes, in particular the part that talks about setting BOOGIE and CORRAL environment variables.

Agreed. Will push a commit for that later today.

michael-emmi:
@zvonimir installation doc updated.

PR #523, opened 2019-12-12
Fix smack pthread.c when using --bit-precise

### LINKED ISSUE #520: Bit-precise pthread failure
The following pthread program fails when checked with `--bit-precise` or `--bit-precise-pointers`. The problem is the hardcoded int boogie types in `share/smack/lib/pthread.c`.

```
cat fail.c && echo --- && smack fail.c --pthread --bpl out.bpl --bit-precise
#include "smack.h"
#include <pthread.h>
#include <stdlib.h>

pthread_t tid;

void *thd(void *arg) {
    return NULL;
}

int main() {
    pthread_create(&tid, 0, thd, NULL);
    return 0;
}
---
SMACK program verifier version 2.4.0
Traceback (most recent call last):
[snip]
Exception: out.bpl(2188,7): Error: invalid type for out-parameter 0 in call to corral_getThreadID: bv32 (expected: int)
out.bpl(2225,17): Error: invalid type for argument 0 in map assignment: bv32 (expected: int)
out.bpl(2228,17): Error: invalid type for argument 0 in map assignment: bv32 (expected: int)
out.bpl(2232,17): Error: invalid type for argument 0 in map assignment: bv32 (expected: int)
out.bpl(2296,7): Error: invalid type for out-parameter 0 in call to corral_getChildThreadID: bv32 (expected: int)
5 type checking errors in out.bpl
[snip]
```

nchong-at-aws:
Fix #520 by switching between Boogie types in `pthread.c`. Previously the file used Boogie int everywhere.

### DIFF
--- a/share/smack/frontend.py
+++ b/share/smack/frontend.py
@@ -78,6 +78,7 @@ def default_clang_compile_command(args, lib = False):
   if args.memory_safety: cmd += ['-DMEMORY_SAFETY']
   if args.integer_overflow: cmd += (['-fsanitize=signed-integer-overflow,shift'] if not lib else ['-DSIGNED_INTEGER_OVERFLOW_CHECK'])
   if args.float: cmd += ['-DFLOAT_ENABLED']
+  if args.bit_precise: cmd += ['-DBIT_PRECISE']
   if sys.stdout.isatty(): cmd += ['-fcolor-diagnostics']
   return cmd
 
--- a/share/smack/lib/pthread.c
+++ b/share/smack/lib/pthread.c
@@ -4,10 +4,16 @@
 #include "pthread.h"
 #include "smack.h"
 
+#if BIT_PRECISE
+#define TID_TYPE "bv32"
+#else
+#define TID_TYPE "int"
+#endif
+
 void __SMACK_init_func_corral_primitives() {
   // Declare these, so bpl parsing doesn't complain
-  __SMACK_top_decl("procedure corral_getThreadID() returns (x:int);");
-  __SMACK_top_decl("procedure corral_getChildThreadID() returns (x:int);");
+  __SMACK_top_decl("procedure corral_getThreadID() returns (x:"TID_TYPE");");
+  __SMACK_top_decl("procedure corral_getChildThreadID() returns (x:"TID_TYPE");");
   __SMACK_top_decl("procedure corral_atomic_begin();");
   __SMACK_top_decl("procedure corral_atomic_end();");
 }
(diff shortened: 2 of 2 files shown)


michael-emmi:
Wondering whether the thread identifier type should be promoted to a Boogie type alias, and decided internally by Smack implementation, rather than decided at Clang-compile time. Thoughts @zvonimir ?

zvonimir:
Thanks @nchong-at-aws for the pull request!
Yep, a slightly cleaner way would probably be:
```
#if BIT_PRECISE
__SMACK_top_decl("type $tidtype = bv32;");
#else
__SMACK_top_decl("type $tidtype = i32;");
#endif
```
And then you can use `$tidtype` everywhere. Should be an easy fix. Thanks!

michael-emmi:
@zvonimir but is the header really the right place to declare the type alias? Wouldn’t SmackRep internally have a better idea, e.g., of the right bit-width for the corresponding int type?

nchong-at-aws:
Hi both, thanks for the comments. I tried the type alias as you suggested Zvonimir, but this gives a clang parse error (below). Can I use `__SMACK_top_decl` at the top-level? All other uses of this definition are in a function body.

```
[...]/share/smack/lib/pthread.c:7:1: error: conflicting types for '__SMACK_top_decl'
[...]/share/smack/include/smack.h:31:6: note: previous declaration is here
```

zvonimir:
@nchong-at-aws : sorry, I should have been more clear. Yep, you have to place `__SMACK_top_decl` within something starting with `__SMACK_init` such as `__SMACK_init_func_corral_primitives`.

@michael-emmi : I think it would be good to avoid having pthreads-specific code within `SmackRep`. 
A more general solution I was thinking about as well is to enhance `__SMACK_code` functionality with hooks for types. So something like `#int` would be compiled by SMACK code into the appropriate Boogie type, depending on whether bit-vectors are selected or not, bit-width of integers, etc. And we could also have `#short`, `#char`, etc.

Oh, one more thing. To avoid having to depend on the platform-dependent bit-width of `int` type in C, we should maybe switch pthreads header to use `int32_t`, which is not platform dependent.

nchong-at-aws:
Ah, so I have figured out that I can simply put the `__SMACK_top_decl` in a fresh function and it gets pushed into the Boogie file as expected. Strangely this is causing me some corral problems (an unhandled exception), but I'm not seeing this problem when I run corral directly on the bpl file.

nchong-at-aws:
Here's the corral error I'm seeing (when using the boogie type alias from commit f8cfc91). The problem disappears when I run corral directly.

```
$ smack --pthread equal.c --bpl out.bpl
SMACK program verifier version 2.4.0
Traceback (most recent call last):
[snip] LB: Took 0.00 s
Verifying program while tracking: {}
Program has a potential bug: False bug
Verifying program while tracking: {$M.0}

Unhandled Exception:
CoreLib.InsufficientDetailsToConstructCexPath: StratifiedInliningErrorReporter: Must find a successor
[snip]

$ corral out.bpl /tryCTrace /noTraceOnDisk /printDataValues:1 /k:1 /useProverEvaluate /timeLimit:1200 /cex:1 /maxStaticLoopBound:1 /recursionBound:1 /bopt:proverOpt:OPTIMIZE_FOR_BV=true /bopt:boolControlVC
LB: Took 0.00 s
Verifying program while tracking: {}
Program has a potential bug: False bug
Verifying program while tracking: {$M.0}
Program has no bugs
[snip]
```

zvonimir:
I am ok with the proposed changes. Let's see if CI passes, and in particular the `clang-format` check.

The error you are seeing is due to incompleteness of Z3. We see it only very rarely when doing sequential program verification. Not sure if doing context-bounded exploration somehow triggers it more frequently. We'll have to explore that a bit more.

Could you share the file that triggers this (in a separate issue)?

PR #527, opened 2019-12-16
Fix type errors in the svcomp scripts when using Python3

shaobo-he:


share/smack/svcomp/toSVCOMPformat.py
```diff
@@ -118,7 +118,7 @@ def buildEmptyXmlGraph(args, hasBug):
     programfile = os.path.abspath(args.orig_files[0])
     addKey(graph, "programfile", programfile)
     with open(programfile, 'r') as pgf:
-      addKey(graph, "programhash", hashlib.sha256(pgf.read()).hexdigest())
+      addKey(graph, "programhash", hashlib.sha256(pgf.read().encode('utf-8')).hexdigest())
```
zvonimir: Wait, to me it seems like you should be using `decode` here and `encode` below. No?
Since `encode` goes from string to bytes, and `decode` goes the other way around.

I also have a more basic question. Why this this fix even needed?

shaobo-he: It's actually the other way around because Python3 hash functions take bytes as inputs while `read` function returns a string. Without this change, I will get an error like `TypeError: Unicode-objects must be encoded before hashing`.

PR #529, opened 2019-12-17
Fix #528: memsafe pthread_join stub

### LINKED ISSUE #528: smack stub `pthread_join` is not memory safe
The stub for `pthread_join` in `share/smack/lib/pthread.c` dereferences `__thread_return` before checking it is non-null.

A program that is checked with `--memory-safety` using `pthread_join` with `__thread_return==NULL` will have an `invalid pointer dereference.`

```
// smack prog.c --pthread --memory-safety
#include "smack.h"
#include <pthread.h>
#include <stdlib.h>

void *t1(void *arg) {
  return NULL;
}

int main() {
  pthread_t tid1;
  pthread_create(&tid1, 0, t1, 0);
  pthread_join(tid1, NULL); //< __thread_return==NULL
  return 0;
}
```

nchong-at-aws:
Check for NULL retval parameter before deref

### DIFF
--- a/test/c/pthread/join_null_retval.c
+++ b/test/c/pthread/join_null_retval.c
@@ -1,0 +1,15 @@
+#include "smack.h"
+#include <pthread.h>
+#include <stdlib.h>
+
+// @expect verified
+// @flag --memory-safety
+
+void *t1(void *arg) { return NULL; }
+
+int main(void) {
+  pthread_t tid1;
+  pthread_create(&tid1, 0, t1, 0);
+  pthread_join(tid1, NULL);
+  return 0;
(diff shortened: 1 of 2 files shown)


test/c/pthread/join_null_retval.c
```diff
@@ -0,0 +1,17 @@
...
+#include <stdlib.h>
+
+// @expect verified
+// @flag --memory-safety
+
+void *t1(void *arg) {
+  return NULL;
+}
+
+int main() {
```
zvonimir: To maintain consistency, this should be `int main(void)`.

PR #533, opened 2019-12-31
Updated Travis CI to Ubuntu 18.04 (bionic)

zvonimir:


michael-emmi:
This is looking alright to me, but I’m not understanding why the following happened:
* CI was unable to add the LLVM bionic apt registry, even though that URL responds to my browser.
* Build config validation didn't happen, even though the validation setting is enabled.

zvonimir:
Related to your first point, I think my latest commit fixes that.
Related to your second point, I fixed warnings reported by the Travis configuration validator, and so I think it is in fact running. I guess they just don't have a nice message when everything is ok and there are no warnings.

michael-emmi:
Ah hah; looks good now!

PR #538, opened 2020-01-17
Renamed Rust nondet into verifier_nondet

zvonimir:
This slightly better matches SVCOMP style for nondet functions.


zvonimir:
Note that I also introduced `verifier_assume` and `verifier_assert` for compatibility with other tools.

Issue #187, opened 2016-11-22
Incorrect encoding for negative operands of unsigned integer comparison instructions


shaobo-he:
Consider the following program,

```C
#include "smack.h"

int main()
{
  unsigned x = 10;
  __VERIFIER_assert(x < ((unsigned)-2));
}
```

The assertion should not fail while SMACK reports a bug. It seems that SMACK treat the negative operand of `ult` instruction as it is. So instruction `%1 = icmp ult i32 10, -2, !dbg !853` is translated to `$i0 := $ult.i32(10, $sub.i32(0, 2));`, which results in the assertion failure.

zvonimir:
Maybe an even bigger problem (but same root cause) is the fact that `UINT_MAX` gets translated as -1.

michael-emmi:
This is a limitation of encoding C integer types with Boogie int types. Notice that things work as you’d expect when bit-precision is enabled.

A general fix which maintains our current integer encoding can be seen in the `integer-encoding-fix` which wraps integers used as unsigned types in the `$tou` function:
````
function $tou.i32(i: int) returns (int) {
  if i >= 0 && i < 4294967296 then i else $mod(i,4294967296)
}
````
You can verify that the fix works by wrapping the `$sub.i32(0, 2)` value:
````
$i0 := $ult.i32(10, $tou.i32( $sub.i32(0, 2) ));
````
The problem with this fix is that it is generally very expensive since modulo operations are sprinkled all over the place.

A less general fix that would work for this example would be to translate integer literals according to their context. For example, in this case we could see that the literal `-2` is used as an unsigned argument, and thus translate it to `UINT_MAX - 1`. This could work for argument literals, but would require more elaborate analysis beyond that, of questionable value.

michael-emmi:
Another thing to consider — not a quick fix — would be to basically do what the `integer-encoding-fix` does, yet run some analysis, e.g., interval analysis, to eliminate as many `$tou` operations as possible. The worst case would still be bad, of course.

Guoanshisb:
@michael-emmi Thanks for your input. I have a quick question, is there a problem with our current integer encoding for arithmetic? I can see that unsigned overflow is one.

michael-emmi:
@Guoanshisb not sure I understand your question, because your example demonstrates that there is a problem with our integer encoding: we do not give the correct unsigned interpretation to integer types.

[...]

PR #540, opened 2020-02-05
Add a regression for issue #187

shaobo-he:
These two programs are based on the one in the issue description of #187.


test/c/basic/sign_cast.c
```diff
@@ -0,0 +1,13 @@
+#include <limits.h>
+#include "smack.h"
```
zvonimir: Ah, formatting check is failing. include smack.h should be before limits.h it seems.

PR #548, opened 2020-02-18
Made memory allocation atomic

zvonimir:


share/smack/lib/smack.c
```diff
@@ -301,10 +303,12 @@ void __SMACK_decls(void) {
...
     "{\n"
+    "  call corral_atomic_begin();\n"
     "  if ($ne.ref.bool(n, $0.ref)) {\n"
     "    $allocatedCounter := $allocatedCounter + 1;\n"
     "  }\n"
     "  call p := $$alloc(n);\n"
+    "  call corral_atomic_end();\n"
     "}\n");
 
 #if MEMORY_MODEL_NO_REUSE_IMPLS
```
michael-emmi: I mean to comment on the following implementation of `$galloc`: should this also be made atomic?

zvonimir: Procedure `$galloc` is I believe only invoked to allocate globals before the entry point of the program is even reached. So it will always be executed sequentially. Unless I am missing something.

michael-emmi: Hmm ok then perhaps we should update something to reflect this so that this kind of knowledge doesn’t remain hidden in our heads :-)

zvonimir: Ok. I added a comment clarifying this. Looks good?

michael-emmi: Looks great!

PR #550, opened 2020-02-25
Replace LLVM's DSA with sea-dsa

shaobo-he:


.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "sea-dsa"]
+	path = sea-dsa
```
zvonimir: Let's avoid using tabs, so please turn them into spaces.

bin/build.sh
```diff
@@ -60,7 +60,7 @@ CONFIGURE_INSTALL_PREFIX=
 CMAKE_INSTALL_PREFIX=
 
 # Partial list of dependencies; the rest are added depending on the platform
-DEPENDENCIES="git cmake python3-yaml python3-psutil unzip wget ninja-build"
+DEPENDENCIES="git cmake python3-yaml python3-psutil unzip wget ninja-build libboost-all-dev subversion"
```
zvonimir: Why does it require subversion? That is really weird.

include/utils/Devirt.h
```diff
@@ -38,7 +41,9 @@ namespace llvm {
   class Devirtualize : public ModulePass, public InstVisitor<Devirtualize> {
```
zvonimir: Question. Why are we using this old devirtualizator from DSA? I thought that seadsa has its own, no?

michael-emmi:
Currently this PR adds many sections of comments, including old code, general uncertainties, as well as legitimate documentation. Shouldn’t we get rid of the first two categories of comments?

tools/llvm2bpl/llvm2bpl.cpp
```diff
@@ -153,7 +152,8 @@ int main(int argc, char **argv) {
 
   pass_manager.add(llvm::createLowerSwitchPass());
   // pass_manager.add(llvm::createCFGSimplificationPass());
-  pass_manager.add(llvm::createInternalizePass());
+  // Shaobo: sea-dsa is inconsistent with the pass below.
+  //pass_manager.add(llvm::createInternalizePass());
```
zvonimir: Do you maybe know why we added this pass? Does removing it mean that there'll be lots of external functions?

lib/utils/Devirt.cpp
```diff
@@ -495,7 +498,8 @@ Devirtualize::runOnModule (Module & M) {
   //
   // Get the targets of indirect function calls.
   //
-  CTF = &getAnalysis<dsa::CallTargetFinder<EQTDDataStructures> >();
+  //CTF = &getAnalysis<dsa::CallTargetFinder<EQTDDataStructures> >();
```
zvonimir: Let's completely erase this, no?

lib/utils/Devirt.cpp
```diff
@@ -374,16 +374,19 @@ Devirtualize::makeDirectCall (CallSite & CS) {
...
-    //   Targets.insert (Targets.begin(), CTF->begin(CS), CTF->end(CS));
-    // TODO presently we filter out unmatching targets:
-    for (auto F = CTF->begin(CS); F != CTF->end(CS); ++F)
+  //if (CTF->size(CS) && CTF->isComplete(CS)) {
+  //  // TODO should we allow non-matching targets?
+  //  // TODO non-matching targets leads to crashes in bounce creation
+  //  // TODO formerly, all call-target-finder tarets were included:
+  //  //   Targets.insert (Targets.begin(), CTF->begin(CS), CTF->end(CS));
+  //  // TODO presently we filter out unmatching targets:
+  //  for (auto F = CTF->begin(CS); F != CTF->end(CS); ++F)
```
zvonimir: Again, shouldn't we completely erase this?

lib/smack/Regions.cpp
```diff
@@ -332,14 +309,29 @@ void Regions::visitAtomicRMWInst(AtomicRMWInst &I) {
...
+
+  if (auto CI = dyn_cast<ConstantInt>(I.getLength()))
+    length = CI->getZExtValue();
+  else
+    length = std::numeric_limits<unsigned>::max();
+
+  idx(I.getDest(), length);
+}
+
+// Shaobo: we need to visit the source location otherwise
```
zvonimir: Put this comment next to the line where you are doing this. Also, maybe remove Shaobo from the comment.

lib/smack/Regions.cpp
```diff
@@ -143,15 +101,22 @@ void Region::init(const Value *V, unsigned length) {
...
-              !isFieldDisjoint(DSA, V, offset) ||
-              DSA->isMemcpyd(representative) || T->isIntegerTy(8));
-  incomplete = !representative || representative->isIncompleteNode();
+              !DSA->isTypeSafe(V) ||
+              //DSA->isMemcpyd(representative) ||
+             T->isIntegerTy(8));
+  incomplete = !representative || representative->isIncomplete();
   complicated = !representative || isComplicated(representative);
-  collapsed = !representative || representative->isCollapsedNode();
+  collapsed = !representative || representative->isOffsetCollapsed();
```
zvonimir: Do you understand the difference between offset-collapsed and type-collapsed?
Why are you using offset-collapsed here and not type-collapsed?

lib/smack/Regions.cpp
```diff
@@ -143,15 +101,22 @@ void Region::init(const Value *V, unsigned length) {
   }
 
   singleton =
-      DL && representative && isSingleton(representative, offset, length);
+      DL && representative && isSingleton(V, length);
   allocated = !representative || isAllocated(representative);
+  // Shaobo: I removed the dependency of the bytewise flag on isFieldDisjoint
```
zvonimir: It would be good to clean this up and figure it out.

lib/smack/Regions.cpp
```diff
@@ -20,109 +15,72 @@ const DataLayout *Region::DL = nullptr;
...
-      if (DL->getTypeAllocSize(T) != length)
-        break;
-      if (!T->isSingleValueType())
-        break;
-      return true;
-    }
-  }
-  return false;
+bool Region::isSingleton(const Value* v, unsigned length) {
+  auto node = DSA->getNode(v);
```
zvonimir: Doesn't seadsa have a function called `isSingleton`?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +23,96 @@ RegisterPass<DSAWrapper>
...
-  const llvm::EquivalenceClasses<const llvm::DSNode *> &eqs =
-      nodeEqs->getEquivalenceClasses();
-  const llvm::DSNode *nn = eqs.getLeaderValue(n);
-  for (unsigned i = 0; i < memcpys.size(); i++)
-    if (memcpys[i] == nn)
-      return true;
-  return false;
+void DSAWrapper::collectMemCpyds(llvm::Module &M) {
+  for (auto& f : M) {
+    for (auto& b : f) {
```
zvonimir: Why don't you use this iterator so that you have one less loop?
http://llvm.org/docs/ProgrammersManual.html#iterating-over-the-instruction-in-a-function

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +23,96 @@ RegisterPass<DSAWrapper>
...
-  for (unsigned i = 0; i < staticInits.size(); i++)
-    if (staticInits[i] == nn)
-      return true;
-  return false;
+void DSAWrapper::countGlobalRefs() {
+  for (auto& g : DG->globals()) {
+    auto& cellRef = g.second;
+    auto* node = cellRef->getNode();
+    if (node) {
+      if (!globalRefCount.count(node))
```
zvonimir: What do we need this for?

.gitmodules
shaobo-he: I replaced each tab with two spaces.

bin/build.sh
shaobo-he: This is something we have to figure out for building sea-dsa.

include/utils/Devirt.h
shaobo-he: As we discussed, sea-dsa doesn't have one.

shaobo-he:
> Currently this PR adds many sections of comments, including old code, general uncertainties, as well as legitimate documentation. Shouldn’t we get rid of the first two categories of comments?

Good idea, let me clean up them.

tools/llvm2bpl/llvm2bpl.cpp
shaobo-he: No, I don't remember why we needed this pass. I don't know if removing it creates a lot of external functions. Let me investigate it.

lib/utils/Devirt.cpp
shaobo-he: Done.

lib/utils/Devirt.cpp
shaobo-he: Done.

lib/smack/Regions.cpp
shaobo-he: This comment is not up-to-date. I removed it. I think I have figured out it.

lib/smack/Regions.cpp
shaobo-he: Done.

lib/smack/Regions.cpp
shaobo-he: No, I don't understand the difference. In our email communications with @caballa, he mentioned that `isOffsetCollapsed` is equivalent to LLVM DSA's `isCollapsedNode`. So I simply used it without thinking further more about the difference.

lib/smack/Regions.cpp
shaobo-he: I don't think I've ever seen it. Maybe you were referring to `isUnique`. If so, its meaning is different from `isSingleton`.

lib/smack/DSAWrapper.cpp
shaobo-he: I tried to use it but encountered some compilation errors. Let me try to figure it out.

lib/smack/DSAWrapper.cpp
shaobo-he: This is needed because we want to make sure that no more than one global variable refer to the same node otherwise we may end up having two non-aliasing pointers point to the same singleton region.

shaobo-he:
> > Currently this PR adds many sections of comments, including old code, general uncertainties, as well as legitimate documentation. Shouldn’t we get rid of the first two categories of comments?
> 
> Good idea, let me clean up them.

@michael-emmi The first two categories are cleaned up.

bin/build.sh
shaobo-he: It's no longer needed.

lib/smack/DSAWrapper.cpp
shaobo-he: Done.

lib/smack/Regions.cpp
zvonimir: I see. What is the meaning of `isUnique`?

CMakeLists.txt
```diff
@@ -230,8 +168,22 @@ add_executable(llvm2bpl
   tools/llvm2bpl/llvm2bpl.cpp
 )
 
+# We have to import LLVM's cmake definitions to build sea-dsa
+# Borrowed from sea-dsa's CMakeLists.txt
+find_package (LLVM 8.0 CONFIG NO_DEFAULT_PATH)
```
zvonimir: Does this have to be updated whenever we update LLVM? Should this be `8.0` or `8.0.1`?

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "sea-dsa"]
+  path = sea-dsa
+  url = https://github.com/seahorn/sea-dsa.git
+  branch = llvm-8.0
```
zvonimir: Can we get a particular commit?
We'll have problems once they start updating sea-dsa.

CMakeLists.txt
shaobo-he: I think so. It should be 8.0 or simply 8.

.gitmodules
shaobo-he: No, I think the way git submodules work is the version is fixed in the repo using them regardless of the updates in the submodule repos.

CMakeLists.txt
shaobo-he: The [version] argument requests a version with which the package found should be compatible (format is major[.minor[.patch[.tweak]]]). The EXACT option requests that the version be matched exactly.

lib/smack/Regions.cpp
shaobo-he: This is what I found in sea-dsa's comments: `// -- a unique scalar is a single-cell global variable.`

shaobo-he: I'm not sure but I guess if a node is unique then it only contains one object that has a scalar type. sea-dsa guys had a comment related to it: https://github.com/seahorn/sea-dsa/blob/55b29333e6661d15fa4d9f69b7e39b80c8f3b5b5/src/ShadowMem.cc#L87.

lib/smack/Regions.cpp
zvonimir: It would be good to understand the difference. So please ask them about this or learn it from their code.

lib/smack/Regions.cpp
zvonimir: So how is `isUnique` different from `isSingleton`?
To me they seems very similar.

shaobo-he: Maybe an example can show this better.

```C
struct Point {
  int x;
  int y;
};

struct Point p;

int main() {
  p.x = 0;
  p.y = 1;
  return 0;
}
```
In this example, the node `p` refers to is not unique whereas `p.x` and `p.y` are considered as singletons.

shaobo-he: So I think `isSingleton` is more or less a concept of cells instead of nodes.

michael-emmi:
Should I wait to review this until tests are passing?

shaobo-he:
> Should I wait to review this until tests are passing?

No, we're still waiting for some fixes on sea-dsa’s side. On smack’s side, it's all there.

tools/llvm2bpl/llvm2bpl.cpp
keram88: It appears this came from here:
https://github.com/smackers/smack/commit/3a3d81941cfc469502d745366b33f5f1a86c8fbc

docs/installation.md
```diff
@@ -153,9 +153,14 @@ installed via the Microsoft Store.
 
 ### Installing SMACK Itself
 
-SMACK is built using [CMake][] via the following sequence of shell commands
-from SMACK's root directory:
-````Shell
+The prerequisite step for building SMACK is to fetch its submodule. Then, it is
+built using using [CMake][]. Both steps can be done via the following sequence
```
zvonimir: You have `using` twice here.

tools/llvm2bpl/llvm2bpl.cpp
zvonimir: Did we get to the bottom of this?

test/c/memory-safety/global_and_heap_alloc_fail_no_free.c
```diff
@@ -9,5 +9,7 @@ int main(void) {
   int *a = &x;
   int *b = malloc(4);
   int c = a[0];
-  return c;
+  // We have to use `b` otherwise the call to malloc will be removed.
```
zvonimir: Do we have to make this change in other versions of this regression?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
-    for (llvm::Function::iterator block = func->begin(); block != func->end();
-         ++block) {
-
-      mcc->visit(*block);
+  SD = &getAnalysis<sea_dsa::DsaAnalysis>().getDsaAnalysis();
+  assert(SD->kind() == sea_dsa::GlobalAnalysisKind::CONTEXT_INSENSITIVE
+         && "Currently we only want the context-insensitive sea-dsa.");
+  for (auto& f : M.functions()) {
+    if(f.hasName() && SmackOptions::isEntryPoint(f.getName())) {
+      DG = &SD->getGraph(f);
```
zvonimir: I don't get this. Why do we need this loop if there is just one big graph?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
-    if (memcpys[i] == nn)
-      return true;
-  return false;
+void DSAWrapper::collectMemCpyds(llvm::Module &M) {
+  for (auto& f : M) {
+    for (inst_iterator I = inst_begin(&f), E = inst_end(&f); I != E; ++I) {
+      if (MemCpyInst* memcpyInst = dyn_cast<MemCpyInst>(&*I)) {
+        auto srcNode = getNode(memcpyInst->getSource());
+        auto destNode = getNode(memcpyInst->getDest());
+        if (srcNode)
```
zvonimir: Can this ever be NULL?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
+          memCpyds.insert(srcNode);
+        if (destNode)
```
zvonimir: Can this ever be NULL?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
+          memCpyds.insert(destNode);
+      }
+      if (MemSetInst* memsetInst = dyn_cast<MemSetInst>(&*I)) {
```
zvonimir: Why is this not `else if`?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
+        auto destNode = getNode(memsetInst->getDest());
+        if (destNode)
```
zvonimir: Can this ever be NULL?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
+void DSAWrapper::collectMemCpyds(llvm::Module &M) {
+  for (auto& f : M) {
+    for (inst_iterator I = inst_begin(&f), E = inst_end(&f); I != E; ++I) {
+      if (MemCpyInst* memcpyInst = dyn_cast<MemCpyInst>(&*I)) {
+        auto srcNode = getNode(memcpyInst->getSource());
+        auto destNode = getNode(memcpyInst->getDest());
+        if (srcNode)
+          memCpyds.insert(srcNode);
+        if (destNode)
+          memCpyds.insert(destNode);
```
zvonimir: We should probably rename `memCpyds` into something more general, since it is not just for memcpy any more, but also for memset too.

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
-  const llvm::DSNode *nn = eqs.getLeaderValue(n);
-  for (unsigned i = 0; i < staticInits.size(); i++)
-    if (staticInits[i] == nn)
-      return true;
-  return false;
+void DSAWrapper::countGlobalRefs() {
+  for (auto& g : DG->globals()) {
+    auto& cellRef = g.second;
+    auto* node = cellRef->getNode();
+    if (node) {
```
zvonimir: Can this ever be NULL?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
-bool DSAWrapper::isFieldDisjoint(const llvm::Value *V,
-                                 const llvm::Function *F) {
-  return TS->isFieldDisjoint(V, F);
+bool DSAWrapper::isStaticInitd(const sea_dsa::Node *n) {
+  return staticInits.count(n) > 0;
 }
 
-bool DSAWrapper::isFieldDisjoint(const GlobalValue *V, unsigned offset) {
-  return TS->isFieldDisjoint(V, offset);
+bool DSAWrapper::isMemCpyd(const sea_dsa::Node *n) {
```
zvonimir: This should be renamed since it is referring to both memcpy and memset.
Also, what about memmove? Also, are there any others?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
+  return memCpyds.count(n) > 0;
 }
 
 bool DSAWrapper::isRead(const Value *V) {
-  const DSNode *N = getNode(V);
-  return N && (N->isReadNode());
+  auto N = getNode(V);
```
zvonimir: Can this ever be NULL?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
+  return N && (N->isRead());
 }
 
 bool DSAWrapper::isAlloced(const Value *v) {
-  const DSNode *N = getNode(v);
-  return N && (N->isHeapNode() || N->isAllocaNode());
+  auto N = getNode(v);
+  return N && (N->isHeap() || N->isAlloca());
```
zvonimir: Can this ever be NULL?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +24,94 @@ RegisterPass<DSAWrapper>
...
-  EquivalenceClasses<const DSNode *>::member_iterator I = Cs.member_begin(C);
-  if (I == Cs.member_end())
-    return false;
-  ++I;
-  if (I != Cs.member_end())
-    return false;
-
-  return true;
+  auto N = getNode(v);
+  return N && N->isExternal();
```
zvonimir: Can this ever be NULL?

lib/smack/DSAWrapper.cpp
```diff
@@ -215,23 +126,118 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
 }
 
 int DSAWrapper::getOffset(const Value *v) {
-  return getOffset(new MemoryLocation(v));
+  if (DG->hasCell(*v)) {
+    auto cell = DG->getCell(*v);
+    auto node = cell.getNode();
+    // Be consistent with the old implementation.
+    if (node && node->isOffsetCollapsed())
```
zvonimir: Can this ever be NULL?

lib/smack/DSAWrapper.cpp
```diff
@@ -215,23 +126,118 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
...
-const DSNode *DSAWrapper::getNode(const Value *v) {
-  const llvm::EquivalenceClasses<const llvm::DSNode *> &eqs =
-      nodeEqs->getEquivalenceClasses();
-  auto *N = nodeEqs->getMemberForValue(v);
-  return N ? eqs.getLeaderValue(N) : nullptr;
+const sea_dsa::Node *DSAWrapper::getNode(const Value *v) {
+  // For sea-dsa, a node is obtained by getting the cell first.
+  if (DG->hasCell(*v))
+    return DG->getCell(*v).getNode();
+  return nullptr;
```
zvonimir: Can this be reached?

lib/smack/DSAWrapper.cpp
```diff
@@ -215,23 +126,118 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
...
-  TS->print(F, module);
+  // TODO: print the ds graph
+}
+
+bool DSAWrapper::isTypeSafe(const Value *v) {
+  typedef std::unordered_map<unsigned, bool> FieldMap;
+  typedef std::unordered_map<const sea_dsa::Node*, FieldMap> NodeMap;
+  static NodeMap nodeMap;
+
+  auto node = getNode(v);
```
zvonimir: Can node ever be NULL?

lib/smack/Regions.cpp
```diff
@@ -18,111 +13,38 @@ namespace smack {
...
-        break;
-      if (DL->getTypeAllocSize(T) != length)
-        break;
-      if (!T->isSingleValueType())
-        break;
-      return true;
-    }
-  }
-  return false;
+  if (DSA->getNumGlobals(node) != 1)
```
zvonimir: We should probably merge all of this into one if condition.

lib/smack/Regions.cpp
```diff
@@ -143,15 +65,15 @@ void Region::init(const Value *V, unsigned length) {
   }
 
   singleton =
-      DL && representative && isSingleton(representative, offset, length);
+      DL && representative && isSingleton(V, length);
```
zvonimir: Merge with the above line since the line is now shorter.

lib/smack/Regions.cpp
zvonimir: Any news on this one?

shaobo-he: > Any news on this one?

I think Arie and his student replied to us. The bottom line here is that we're using the correct flag here.

test/c/memory-safety/global_and_heap_alloc_fail_no_free.c
shaobo-he: No, I think only this benchmark was failing.

lib/smack/DSAWrapper.cpp
shaobo-he: I think sea-dsa's APIs only allow us to get the graph from a function. I may be able to get the graph by getting the first function in a module. Should I try it?

lib/smack/DSAWrapper.cpp
shaobo-he: I'm not sure about this.

tools/llvm2bpl/llvm2bpl.cpp
shaobo-he: No, but it seems that removing this line does not cause any performance issues. I propose that we get back to this question when it manifests in an example.

share/smack/top.py
```diff
@@ -339,6 +339,8 @@ def llvm_to_bpl(args):
 
   cmd = ['llvm2bpl', args.linked_bc_file, '-bpl', args.bpl_file]
   cmd += ['-warn-type', args.warn]
+  cmd += ['-sea-dsa=ci']
```
zvonimir: Why do we turn this on even when user selects `no-memory-splitting`?

test/c/memory-safety/global_and_heap_alloc_fail_no_free.c
zvonimir: Yes, but others passing just because code got optimized away?

lib/smack/DSAWrapper.cpp
zvonimir: Either that, or get it by looking for the entry point function (since we know that one is "normal" and it exists), and then when you find it, you get it from there. I would guess either way should work, and we should probably remove this loop.

lib/smack/DSAWrapper.cpp
```diff
@@ -22,186 +25,86 @@ RegisterPass<DSAWrapper>
...
+void DSAWrapper::collectMemCpyds(llvm::Module &M) {
+  for (auto &f : M) {
+    for (inst_iterator I = inst_begin(&f), E = inst_end(&f); I != E; ++I) {
+      if (MemCpyInst *memcpyInst = dyn_cast<MemCpyInst>(&*I)) {
+        auto srcNode = getNode(memcpyInst->getSource());
+        auto destNode = getNode(memcpyInst->getDest());
+        memCpyds.insert(srcNode);
+        memCpyds.insert(destNode);
+      }
+      if (MemSetInst *memsetInst = dyn_cast<MemSetInst>(&*I)) {
```
zvonimir: Shouldn't this be `else if` since the two ifs are mutually exclusing?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,186 +25,86 @@ RegisterPass<DSAWrapper>
...
-}
-
-bool DSAWrapper::isFieldDisjoint(const llvm::Value *V,
-                                 const llvm::Function *F) {
-  return TS->isFieldDisjoint(V, F);
+void DSAWrapper::countGlobalRefs() {
+  for (auto &g : DG->globals()) {
+    auto &cellRef = g.second;
+    auto *node = cellRef->getNode();
+    if (node) {
```
zvonimir: Replace this with `assert(node)` and remove the if statement.

lib/smack/DSAWrapper.cpp
zvonimir: This is still outstanding.

lib/smack/DSAWrapper.cpp
```diff
@@ -215,23 +118,115 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
 }
 
 int DSAWrapper::getOffset(const Value *v) {
-  return getOffset(new MemoryLocation(v));
+  if (DG->hasCell(*v)) {
```
zvonimir: This should always be true, and so remove this if condition.

lib/smack/DSAWrapper.cpp
```diff
@@ -215,23 +118,115 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
...
+    auto cell = DG->getCell(*v);
```
zvonimir: Add `assert(cell)` after this line as a sanity check.

lib/smack/DSAWrapper.cpp
```diff
@@ -215,23 +118,115 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
...
+    auto node = cell.getNode();
```
zvonimir: Add `assert(node)` after this line as a sanity check.

lib/smack/DSAWrapper.cpp
```diff
@@ -215,23 +118,115 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
...
+    assert(offset <= INT_MAX && "Cannot handle large offsets");
+    return offset;
+  }
+  // Return 0 if we can't find the cell that `v` associates with.
+  return 0;
+}
+
+const sea_dsa::Node *DSAWrapper::getNode(const Value *v) {
+  // For sea-dsa, a node is obtained by getting the cell first.
+  if (DG->hasCell(*v)) {
```
zvonimir: Again, this should always be true, and so remove this if condition.

lib/smack/DSAWrapper.cpp
```diff
@@ -215,23 +118,115 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
...
-  const llvm::EquivalenceClasses<const llvm::DSNode *> &eqs =
-      nodeEqs->getEquivalenceClasses();
-  auto *N = nodeEqs->getMemberForValue(v);
-  return N ? eqs.getLeaderValue(N) : nullptr;
+bool DSAWrapper::isTypeSafe(const Value *v) {
+  typedef std::unordered_map<unsigned, bool> FieldMap;
+  typedef std::unordered_map<const sea_dsa::Node *, FieldMap> NodeMap;
+  static NodeMap nodeMap;
+
+  auto node = getNode(v);
```
zvonimir: Add `assert(node)` after this line as a sanity check.

lib/smack/DSAWrapper.cpp
```diff
@@ -215,23 +118,115 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
...
+        // TODO: fix the const_cast
+        unsigned length =
+            dataLayout->getTypeStoreSize(const_cast<llvm::Type *>(t));
+        if (length > fieldLength)
+          fieldLength = length;
+      }
+
+      // Check if the current field overlaps with the next *fields*
+      for (auto oi = ++offsetIterator; oi != offsets.end(); ++oi) {
+        bool overlap = false;
```
zvonimir: Why do you need the `overlap` local variable?
To me it seems that breaking directly from the if condition would be semantically equivalent (by simpler) to what you have right now. No?

lib/smack/Regions.cpp
```diff
@@ -18,111 +13,27 @@ namespace smack {
...
-      if (I != GlobalECs->end()) {
-        count +=
-            std::distance(GlobalECs->member_begin(I), GlobalECs->member_end()) -
-            1;
-      }
-    }
-  }
+bool Region::isSingleton(const Value *v, unsigned length) {
+  // TODO can we do something for non-global nodes?
+  auto node = DSA->getNode(v);
```
zvonimir: Add `assert(node)` after this line, just as a sanity check.

test/c/memory-safety/global_and_heap_alloc_fail_no_free.c
shaobo-he: The corresponding benchmarks without errors seem to be well-defined so I don't think compiler optimizes anything.

share/smack/top.py
shaobo-he: Regions.cpp is not the only place where dsa is needed. CodifyStaticInits.cpp also relies on DSA. Furthermore, we have code that bypasses DSA when `no-memory-splitting` is enabled.

lib/smack/DSAWrapper.cpp
shaobo-he: I don't think we need a sanity check here since `getNode` already gets one upon returning.

lib/smack/Regions.cpp
shaobo-he: I don't think we need a sanity check here since getNode already gets one upon returning.

lib/smack/DSAWrapper.cpp
shaobo-he: Yes, you're right. It should break the loop from the else branch.

lib/smack/DSAWrapper.cpp
shaobo-he: Done.

shaobo-he: > Add `assert(cell)` after this line as a sanity check.

`cell` is a reference not a pointer.

test/c/memory-safety/global_and_heap_alloc_fail_no_free.c
zvonimir: Ah, got it, makes sense.

share/smack/top.py
zvonimir: I see. Ok, makes sense.

include/smack/DSAWrapper.h
```diff
@@ -7,52 +7,32 @@
...
+  std::unordered_map<const sea_dsa::Node *, unsigned> globalRefCount;
   const llvm::DataLayout *dataLayout;
 
-  std::vector<const llvm::DSNode *> collectMemcpys(llvm::Module &M,
-                                                   MemcpyCollector *mcc);
-  std::vector<const llvm::DSNode *> collectStaticInits(llvm::Module &M);
-  llvm::DSGraph *getGraphForValue(const llvm::Value *V);
-  int getOffset(const llvm::MemoryLocation *l);
+  void collectStaticInits(llvm::Module &M);
+  void collectMemCpyds(llvm::Module &M);
```
shaobo-he: Change it to MemOpeds.

include/smack/DSAWrapper.h
```diff
@@ -7,52 +7,32 @@
...
-  dsa::TypeSafety<llvm::TDDataStructures> *TS;
-  std::vector<const llvm::DSNode *> staticInits;
-  std::vector<const llvm::DSNode *> memcpys;
-  std::unordered_set<const llvm::DSNode *> intConversions;
+  sea_dsa::GlobalAnalysis *SD;
+  // The ds graph since we're using the context-insensitive version which
+  // results in one graph for the whole module.
+  sea_dsa::Graph *DG;
+  std::unordered_set<const sea_dsa::Node *> staticInits;
+  std::unordered_set<const sea_dsa::Node *> memCpyds;
```
zvonimir: Rename this into `memOpds`.

include/utils/Devirt.h
```diff
@@ -38,7 +41,9 @@ namespace llvm {
...
     private:
       // Access to analysis pass which finds targets of indirect function calls
-      dsa::CallTargetFinder<EQTDDataStructures> *CTF;
+      //dsa::CallTargetFinder<EQTDDataStructures> *CTF;
```
zvonimir: Remove this comment.

lib/smack/DSAWrapper.cpp
```diff
@@ -214,24 +102,104 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
...
-  auto *N = nodeEqs->getMemberForValue(v);
-  return N ? eqs.getLeaderValue(N) : nullptr;
+bool DSAWrapper::isTypeSafe(const Value *v) {
+  typedef std::unordered_map<unsigned, bool> FieldMap;
+  typedef std::unordered_map<const sea_dsa::Node *, FieldMap> NodeMap;
+  static NodeMap nodeMap;
+
+  auto node = getNode(v);
+
+  if (node->isOffsetCollapsed() || node->isExternal() || node->isIncomplete() ||
```
zvonimir: Variable `node` could be `nullptr` here, and so we need to check that first.

lib/smack/Regions.cpp
```diff
@@ -18,111 +13,27 @@ namespace smack {
...
-        break;
-      if (DL->getTypeAllocSize(T) != length)
-        break;
-      if (!T->isSingleValueType())
-        break;
-      return true;
-    }
-  }
-  return false;
+  return DSA->getNumGlobals(node) == 1 && !node->isArray() &&
```
zvonimir: I guess `node` can be nullptr here, and so we should check for that.

shaobo-he: Actually `node` here can't be nullptr because if it is, `isSingleton` will not be called.

lib/smack/DSAWrapper.cpp
shaobo-he: Good catch! I think we should check nullptr first and then call this function depending on the result. Please see my latest commit.

lib/smack/DSAWrapper.cpp
```diff
@@ -214,24 +102,104 @@ unsigned DSAWrapper::getPointedTypeSize(const Value *v) {
...
+  return node;
 }
 
-// TODO: Should this return the node or its leader?
-const DSNode *DSAWrapper::getNode(const Value *v) {
-  const llvm::EquivalenceClasses<const llvm::DSNode *> &eqs =
-      nodeEqs->getEquivalenceClasses();
-  auto *N = nodeEqs->getMemberForValue(v);
-  return N ? eqs.getLeaderValue(N) : nullptr;
+bool DSAWrapper::isTypeSafe(const Value *v) {
```
zvonimir: Why is the argument of this function of type `Value` and not of type `Node`?
Now that I look at this code, I don't understand why it should be `Value` because `Node` would seem more appropriate - since this function just uses `v` to get its node and for nothing else.

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +25,70 @@ RegisterPass<DSAWrapper>
...
-  return mcc->getMemcpys();
-}
-
-std::vector<const llvm::DSNode *>
-DSAWrapper::collectStaticInits(llvm::Module &M) {
-  std::vector<const llvm::DSNode *> sis;
+void DSAWrapper::collectStaticInits(llvm::Module &M) {
   for (GlobalVariable &GV : M.globals()) {
     if (GV.hasInitializer()) {
       if (auto *N = getNode(&GV)) {
```
zvonimir: Can this ever be null?

lib/smack/DSAWrapper.cpp
```diff
@@ -22,185 +25,70 @@ RegisterPass<DSAWrapper>
...
-    return TD->getGlobalsGraph();
-
-  // XXX I know this looks bad, but it works for now
-  for (auto U : V->users()) {
-    return getGraphForValue(U);
+void DSAWrapper::collectMemOpds(llvm::Module &M) {
+  for (auto &f : M) {
+    for (inst_iterator I = inst_begin(&f), E = inst_end(&f); I != E; ++I) {
+      if (MemCpyInst *memcpyInst = dyn_cast<MemCpyInst>(&*I)) {
+        memOpds.insert(getNode(memcpyInst->getSource()));
```
zvonimir: Can `getNode` here ever be null?
I would guess that undef could flow into memcpy for example. Not sure.
What happens if it is null and null gets added into `memOpds`?

lib/smack/Regions.cpp
zvonimir: Wait. I am not sure about this since `getNode` can return `nullptr`.

lib/smack/DSAWrapper.cpp
shaobo-he: This function actually further get the offset. Type safety is a property with respect to offsets, which means that just getting the node is not enough.

lib/smack/DSAWrapper.cpp
shaobo-he: I'm not sure. I would say it shouldn't be null here. Should I add an assertion?

lib/smack/DSAWrapper.cpp
shaobo-he: I think we don't have to worry about `getNode` being nullptr here because `getNumGlobals` will trigger assertion failure first before `isMemOpd` is called.

lib/smack/Regions.cpp
shaobo-he: `isSingleton` only gets called once the `representative` is not nullptr, which means that we don't need to check nullptr deference here.

lib/smack/DSAWrapper.cpp
zvonimir: I guess just add it. I cannot hurt.

michael-emmi:
What’s the deal with the failed memory-safety and pthread failures in the PR build? Wouldn’t we have to fix those before merging?

CMakeLists.txt
```diff
@@ -18,6 +18,14 @@ if (NOT WIN32 OR MSYS OR CYGWIN)
     OUTPUT_STRIP_TRAILING_WHITESPACE
   )
 
+  # sea-dsa requires this variable to be set
+  execute_process(
+    COMMAND ${LLVM_CONFIG_EXECUTABLE} --prefix
+    OUTPUT_VARIABLE LLVM_DIR
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+  string(APPEND LLVM_DIR "/share/llvm/cmake")
```
michael-emmi, line 27: Why is this set to the `share/llvm/cmake` subdirectory of `$(llvm-config --prefix)`?

With LLVM from Homebrew (on macOS) the hierarchy looks like this:
```console
% ls $(llvm-config --prefix)/share
clang      cmake      man        opt-viewer scan-build scan-view
```
Note that `cmake` exists instead of `llvm/cmake`.

This doesn’t seem to break anything, but I do wonder what the purpose is.

CMakeLists.txt
```diff
@@ -230,8 +168,22 @@ add_executable(llvm2bpl
   tools/llvm2bpl/llvm2bpl.cpp
 )
 
+# We have to import LLVM's cmake definitions to build sea-dsa
+# Borrowed from sea-dsa's CMakeLists.txt
+find_package (LLVM 8.0.1 EXACT CONFIG NO_DEFAULT_PATH)
```
michael-emmi, line 173: We are we specifying `NO_DEFAULT_PATH`? This seems to disallow pre-existing LLVM installations to be used. When running CMake with this parameter I see the following (on macOS with llvm@8 from Homebrew):
```console
% (rm -r build ; mkdir build && cd build && cmake .. -G Ninja)
rm: build: No such file or directory
-- The C compiler identification is AppleClang 11.0.0.11000033
-- The CXX compiler identification is AppleClang 11.0.0.11000033
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Could NOT find LLVM (missing: LLVM_DIR)
CMake Error at CMakeLists.txt:175 (include):
  include could not find load file:

    AddLLVM


CMake Error at CMakeLists.txt:176 (include):
  include could not find load file:

    HandleLLVMOptions


CMake Error at sea-dsa/src/CMakeLists.txt:1 (add_llvm_library):
  Unknown CMake command "add_llvm_library".


-- Configuring incomplete, errors occurred!
See also "/Users/mje/Code/smack/build/CMakeFiles/CMakeOutput.log".
```
Without this parameter CMake succeeds.

michael-emmi, line 173: Also, CMake ought to configure boost. Building on macOS with Boost from Homebrew fails like this:
```console
% (cd build && ninja)                                         
[2/57] Building CXX object CMakeFiles/smackTranslator.dir/lib/smack/Regions.cpp.o
FAILED: CMakeFiles/smackTranslator.dir/lib/smack/Regions.cpp.o 
/Library/Developer/CommandLineTools/usr/bin/c++  -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -I../include -I../sea-dsa/include -I/usr/local/Cellar/llvm@8/8.0.1_1/include -std=c++11   -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fno-exceptions -fno-rtti -Wno-undefined-var-template -fPIC -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -std=c++11 -w -fdiagnostics-color -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk -MD -MT CMakeFiles/smackTranslator.dir/lib/smack/Regions.cpp.o -MF CMakeFiles/smackTranslator.dir/lib/smack/Regions.cpp.o.d -o CMakeFiles/smackTranslator.dir/lib/smack/Regions.cpp.o -c ../lib/smack/Regions.cpp
In file included from ../lib/smack/Regions.cpp:4:
In file included from ../include/smack/Regions.h:7:
../sea-dsa/include/sea_dsa/Graph.hh:3:10: fatal error: 'boost/container/flat_map.hpp' file not found
#include "boost/container/flat_map.hpp"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
Adding the following fixes this:
```cmake
find_package (Boost 1.55)
if (Boost_FOUND)
  include_directories (${Boost_INCLUDE_DIRS})
endif ()
```
Shouldn’t this be present on all platforms?

I wonder how Linux builds are locating boost otherwise (?)

docs/installation.md
```diff
@@ -87,6 +87,7 @@ SMACK depends on the following projects:
 * [Mono][] version 5.0.0 or greater (except on Windows)
 * [Z3][] or compatible SMT-format theorem prover
 * [Boogie][] or [Corral][] or compatible Boogie-format verifier
+* [sea-dsa][]
```
michael-emmi, line 90: Shouldn’t we add Boost to the list of requirements, since Smack code now uses it directly?

shaobo-he:
> What’s the deal with the failed memory-safety and pthread failures in the PR build? Wouldn’t we have to fix those before merging?

These are known issues and the failures only manifest once we merge this PR. We will move these failing regressions into the failing folder once we merge the PR.

test/rust/functions/closure.rs
```diff
@@ -2,7 +2,6 @@
 mod smack;
 use smack::*;
 
-// @flag --no-memory-splitting
 // @expect verified
```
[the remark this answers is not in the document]
shaobo-he: > Why is it that memory splitting is no longer disabled on several (all Rust) regressions? Is this something that should be documented?

I don't think so, the old DSA is unsound on some Rust regressions so we have to disable memory splitting for them. sea-dsa doesn't have this issue. Therefore, this flag is removed for these regressions.

share/smack/top.py
```diff
@@ -339,6 +339,8 @@ def llvm_to_bpl(args):
...
+  #cmd += ['-sea-dsa-type-aware']
```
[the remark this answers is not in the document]
shaobo-he: This flag can lead to unsoundness. I'll add more comments to document it.

docs/installation.md
shaobo-he: Boogie is a dependency of sea-dsa. I think we should not list secondary dependencies.

CMakeLists.txt
shaobo-he: I think `LLVM_DIR` variable has to be set to a valid path in order to include LLVM cmake macros. These macros are used by sea-dsa, so I just copy-pasted their configuration.

CMakeLists.txt
shaobo-he: It seems c++ compilers on Linux are able to locate the header files.

docs/installation.md
michael-emmi: I would agree if you were consuming sea-dsa as an already-built library, but you are instead building sea-dsa directly in the build of Smack. This means that Smack’s build requires Boost.

If building sea-dsa were separate from building Smack, then I would agree that Boost could be considered not a direct dependency of Smack.

CMakeLists.txt
michael-emmi: Can you explain why? It seems a bit miraculous.

CMakeLists.txt
michael-emmi: Perhaps we should raise an issue with sea-dsa then? Can you verify whether that path even exists on your setup? (As mentioned, it does not in mine.)

docs/installation.md
shaobo-he: Sounds good. Let me put Boost in.

PR #552, opened 2020-03-26
Add experimental support for Yices2

keram88:
This adds support for SMACK to leverage Yices2 support in Corral and Boogie.

We should merge this after we decide on which new versions of Corral and Boogie we are going to use.


zvonimir:
Please refresh this with the current develop since there are merge conflicts. Then I'll review it. Thx!

zvonimir:
@keram88 Please update this to the latest develop and clean up any merge conflicts. Then I'll review it. Thx!

keram88:
I think I've addressed all concerns.
Yices doesn't support quantifiers, so memcpy, memset and memory safety don't work.
Yices also doesn't support bv2nat so would require `--bit-precise-pointers`.

shaobo-he:
> I think I've addressed all concerns.
> Yices doesn't support quantifiers, so memcpy, memset and memory safety don't work.
> Yices also doesn't support bv2nat so would require `--bit-precise-pointers` for bit-vector code with pointers.

I think maybe Yices can handle memcpy and memset if we encode them using lambda terms. This is what I found in Mathias Preiner's thesis,

```
Other SMT solvers that employ a lazy SMT solving approach and natively support lambda terms 
such as CVC4 [3] or Yices [25] also treat them eagerly, similarly to UCLID, and eliminate all
occurrences of lambda terms by substituting them with their instantiated function body
(cf. C-style macros). 
```

I'm sure boolector supports it. But producing lambda terms is a task of Boogie. I don't know if Boogie already does it.

zvonimir:
I just merged this. Let's continue the above discussion somewhere else. Thx!

PR #563, opened 2020-04-10
Update installation procedures, vagrant, travis

zvonimir:
This pull request tries to clean up the installation procedures, vagrant, and travis.


bin/build.sh
```diff
@@ -418,52 +360,34 @@ if [ ${INSTALL_CVC4} -eq 1 ] ; then
...
-    ${WGET} https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
-    mono ./nuget.exe restore Boogie.sln
-    rm -rf /tmp/nuget/
-    msbuild Boogie.sln /p:Configuration=Release
-    ln -sf ${Z3_DIR}/bin/z3 ${BOOGIE_DIR}/Binaries/z3.exe
-    ln -sf ${CVC4_DIR}/cvc4 ${BOOGIE_DIR}/Binaries/cvc4.exe
-    puts "Built Boogie"
+
+if [ ${INSTALL_BOOGIE} -eq 1 ] ; then
+  if [ ! -d "$BOOGIE_DIR" ] ; then
```
michael-emmi, line 368: You could simplify a bit here:
* Dotnet tools are content with sharing tool path, so you could replace `BOOGIE_DIR` and `CORRAL_DIR` with `DOTNET_TOOL_PATH` — or even just use the default global path.
* Detecting and reporting on whether Boogie/Corral is “already installed” seems redundant with `dotnet tool install`, which will do that anyways.

zvonimir: Good catch! Done.

PR #569, opened 2020-04-21
Integer encoding fix 2020

shaobo-he:


share/smack/top.py
```diff
@@ -304,6 +304,13 @@ def arguments():
                 (note that memory-safety is the union of valid-deref,
                 valid-free, memleak)''')
 
+    translate_group.add_argument(
```
zvonimir, line 307: I am wondering if we should merge various encoding options we currently have, into one option that contains several choices. We can all the option `--encoding-of-ints` or something like that, and the options would be `integer, bit-precise, bit-precise-pointers, and wrapped-integer`. The default would be integer. That would make it cleaner since currently you can select multiple of these options which does not make sense.

lib/smack/Naming.cpp
```diff
@@ -85,6 +85,8 @@ const std::string Naming::CONTRACT_EXPR = "$expr";
 const std::string Naming::MEMORY_SAFETY_FUNCTION =
     "__SMACK_check_memory_safety";
 const std::string Naming::MEMORY_LEAK_FUNCTION = "__SMACK_check_memory_leak";
+const std::pair<std::string, std::string> Naming::INT_WRAP_FUNCTIONS = {"$tos",
```
zvonimir, line 88: Maybe this should be a map from bool to string? Then you don't have to introduce the getter function.

lib/smack/Prelude.cpp
```diff
@@ -168,6 +168,44 @@ FuncDecl *builtinOp(std::string baseName, const Attr *attr,
 const std::vector<unsigned> IntOpGen::INTEGER_SIZES{
     1, 5, 6, 8, 16, 24, 32, 40, 48, 56, 64, 80, 88, 96, 128, 160, 256};
 
+const std::map<unsigned, std::string> IntOpGen::INT_LIMITS{
+    {0, "1"},
+    {1, "2"},
+    {4, "16"},
+    {5, "32"},
```
zvonimir, line 175: You have 5 twice. Also, where are the first numbers of these pairs coming from? I see that they do not match `INTEGER_SIZES` above, and so I am curious how you picked them.
Also, why do you need 0?

share/smack/top.py
shaobo-he: Good idea. Let me implement it.

lib/smack/Prelude.cpp
shaobo-he: This table computes both `2^(i-1)` and `2^i` for `i` in `INTEGER_SIZES`, that's probably I got 2 `5`s in the keys.

lib/smack/Naming.cpp
shaobo-he: I think I prefer the existing implementation since having a getter function increases readability. If it were implemented using a map, getting a function will be something like `Naming::INT_WRAP_FUNCTIONS[true]` which is less meaningful than `Naming::getIntWrapFunc(true)`.

share/smack/top.py
zvonimir: Good suggestion! But can all of these be mix and matched!?
We should try this (and maybe have a regression that tries all possible combinations).

lib/smack/Naming.cpp
zvonimir: Ok, sounds good.

lib/smack/Prelude.cpp
zvonimir: I don't get it. Integer sizes are `1, 5, 6, 8, 16, 24, 32, 40, 48, 56, 64, 80, 88, 96, 128, 160, 256`.
In this map though you are mapping things such as `0`, `4`, `7`, etc., which are not in integer sizes.
So what you wrote above "This table computes both 2^(i-1) and 2^i for i in INTEGER_SIZES" seems wrong to me, in particular the part "for i in INTEGER_SIZES".

Could you please clarify?

michael-emmi: This manual coding of strings is not great, because we cannot identify potential mistakes visually. Why don’t we just create this map programmatically? Looks like the formula is: compute the power of two whose exponent is given by the key, and convert to string.

include/smack/Naming.h
```diff
@@ -96,6 +96,7 @@ class Naming {
   static const std::vector<std::string> RUST_PANICS;
   static const std::string RUST_PANIC_ANNOTATION;
 
+  static const std::pair<std::string, std::string> INT_WRAP_FUNCTIONS;
```
michael-emmi, line 99: Would be less error prone to define two separate constants, e.g., `INT_WRAP_SIGNED_FUNCTION` and `INT_WRAP_UNSIGNED_FUNCTION`.

lib/smack/Prelude.cpp
```diff
@@ -280,6 +319,9 @@ struct IntOpGen::IntArithOp : public IntOp {
       : IntOp(opName, arity, intOp, bvOp, alsoUsedByPtr) {}
 
   FuncDecl *getIntFunc(unsigned size) const {
+    if (!intOp)
+      return nullptr;
```
michael-emmi, line 323: Are callers of getIntFunc ready to deal with the possibility of a null value here? Might be better to either use [the optional type](https://en.cppreference.com/w/cpp/utility/optional) or raise an error.

lib/smack/Prelude.cpp
shaobo-he: > This manual coding of strings is not great, because we cannot identify potential mistakes visually. Why don’t we just create this map programmatically? Looks like the formula is: compute the power of two whose exponent is given by the key, and convert to string.

There are some bit-widths that are too large for 64 bit. So we probably need arbitrary precison integer library unless we want to write a print function for large bit-widths ourselves. The former solution would require a new dependency.

lib/smack/Prelude.cpp
shaobo-he: Yes, it can deal with null values. I prefer the current pointer version otherwise I need to change a lot of code.

include/smack/Naming.h
shaobo-he: Can you clarify it? Do we still need the getter function `getIntWrapFunc`?

share/smack/top.py
shaobo-he: It seems the current implementation forces `--bit-precise-pointers` to imply `--bit-precise`. We should lift the restriction.

lib/smack/Prelude.cpp
keram88: We can use this: https://llvm.org/doxygen/classllvm_1_1APInt.html. We use it in the integer overflow checker for this purpose.

share/smack/top.py
shaobo-he: Please see #600

lib/smack/Prelude.cpp
```diff
@@ -556,30 +677,75 @@ struct IntOpGen::IntPred : public IntOp {
     }
     return funcs;
   }
+
+  template <BinExpr::Binary OP, bool SIGN>
+  static const Expr *intPredExpr(unsigned size) {
+    // SHAOBO: we apply modulo operations to the operands.
+    // Here is the reasoning: let's assume the cmp operation is unsigned,
+    // and there's a sequence of arithmetic operations which only contain
+    // addition, substraction, multiplication. The inputs to such a computation
```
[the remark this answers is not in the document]
shaobo-he: Good catch. Fixed.

PR #578, opened 2020-05-20
Remove smack.rs

keram88:
This removes the dependency on `smack.rs` in the local directory for verification.


zvonimir:
Ok, I am worried that weird things will start happening once we run regressions in parallel, and there are race conditions on `smack.rs`. No?

Could we not compile `smack.rs` as rlib and include it using `--extern` when rustc gets invoked?

keram88:
Good point. I wasn't able to get the rlib to work.

PR #580, opened 2020-06-10
Change Rust print*! macros to simply evaluate the arguments

### LINKED ISSUE #579: Rust print reduces performance
For example,

```Rust
#[macro_use]
extern crate smack;
use smack::*;

// @expect verified

struct Point {
  x: u64,
  y: u64,
  m: u64,
  n: u64,
  p: u64,
  q: u64
}

fn take_and_gives_back(p: Point) -> Point {
  print!("{}", p.x);
  print!("{}", p.y);
  print!("{}", p.m);
  print!("{}", p.n);
  print!("{}", p.p);
  print!("{}", p.q);
  p
}

fn main() {
  let v1 = 1u64.verifier_nondet();
  let v2 = 2u64.verifier_nondet();

  let a = Point {x :v1, y: v2, m: v1, n: v2, p:v1, q: v2};
  smack::assert!(take_and_gives_back(a).x == v1);
}
```

This program takes around 8s on my machine to verify. This is because the `print` macro expands to a lot of code. We should at least redefine this macro such that it only evaluates the arguments.

shaobo-he:
Fixes #579

### DIFF
--- a/test/rust/basic/print.rs
+++ b/test/rust/basic/print.rs
@@ -1,0 +1,34 @@
+#[macro_use]
+extern crate smack;
+use smack::*;
+
+// @expect verified
+
+static mut NUM: i32 = 0;
+
+fn incr(x: i32) -> i32 {
+  unsafe {
+    NUM += x;
+    NUM
+  }
+}
+
+fn test_print() {
+  print!("hola");
+  println!("hola");
+  print!("hola, senor {}", incr(1));
+  println!("hola, senor {}", incr(2));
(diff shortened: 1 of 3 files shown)


test/rust/basic/print.rs
```diff
@@ -0,0 +1,34 @@
...
+  unsafe {
+    NUM += x;
+    NUM
+  }
+}
+
+fn test_print() {
+  print!("hola");
+  println!("hola");
+  print!("hola, senor {}", incr(1));
```
keram88, line 19: Can we make this `"hola señor"`?

PR #582, opened 2020-06-11
Add Rust regressions

shaobo-he:


zvonimir:
Or maybe we should create a folder called `arrays`? Maybe that would be nicer.

shaobo-he:
> Or maybe we should create a folder called `arrays`? Maybe that would be nicer.

I'd prefer the `array` folder. Let me do it.

PR #584, opened 2020-06-13
Updated to LLVM 9

zvonimir:
Updated SMACK to LLVM 9.


zvonimir:
@keram88 Could you please update Rust in this branch to LLVM 9 as well? Thanks!

.travis.yml
```diff
@@ -16,6 +16,7 @@ addons:
 env:
   global:
     - COMPILER_NAME=clang COMPILER=clang++ CXX=clang++ CC=clang
+    - LLVM_SHORT_VERSION=9
```
michael-emmi, line 19: Perhaps we could read this from `bin/versions` to eliminate redundancy here?

https://stackoverflow.com/questions/59909029/travis-yml-file-should-read-from-another-file-a-value-and-use-it

CMakeLists.txt
```diff
@@ -167,7 +168,7 @@ if (Boost_FOUND)
 endif ()
 # We have to import LLVM's cmake definitions to build sea-dsa
 # Borrowed from sea-dsa's CMakeLists.txt
-find_package (LLVM 8.0.1 EXACT CONFIG)
+find_package (LLVM 9 CONFIG)
```
michael-emmi, line 171: Perhaps we could read this from `bin/versions` to eliminate redundancy here?

https://cmake.org/cmake/help/v3.0/command/file.html

zvonimir: Done.

.travis.yml
zvonimir: Done.

PR #587, opened 2020-07-10
Demangle C++/Rust names in the error traces

shaobo-he:
This commit demangles function/variable names shown in the error traces of
failing C++/Rust programs. It invokes cxxfilt and rustfilt if they are in the
PATH. It also installs rustfilt when Rust is installed.


zvonimir:
I have a more general question about this pull request. Why are we demangling error traces as opposed to, for example, bpl files?

shaobo-he:
> I have a more general question about this pull request. Why are we demangling error traces as opposed to, for example, bpl files?
> It is because multiple mangled functions are going to demangle into the same function name, hence causing name clashes?

No, I think it's because that the demangled names are not valid Boogie identifiers (e.g., `smack::Unique<T>::as_ptr`).

PR #588, opened 2020-07-11
Update SMACK to latest SV-COMP Benchmarks

shaobo-he:


share/smack/svcomp/utils.py
```diff
@@ -169,10 +169,20 @@ def is_stack_benchmark(args, csource):
       print("Stumbled upon a stack-based memory safety benchmark\n")
     sys.exit(smack.top.results(args)['unknown'])
 
+def inject_assert_false(args):
+    with open(args.bpl_file, 'r') as bf:
```
zvonimir, line 173: I think that indentation is messed up here. We are using two space.

shaobo-he: Maybe it's a known issue but four space indentation is recommended for Python programs. Moreover, this is not closely related to this PR but I'm thinking of reformating our Python scripts into PEP 8-compliant. I tried automatic formating tools like autopep8 and it turns out lightweight. So I was wondering if we can keep these and I'll submit a new PR that turns our Python scripts PEP 8-compliant.

zvonimir: Sounds good.

PR #590, opened 2020-07-14
Mark MemOpd nodes type-unsafe

### LINKED ISSUE #589: Invalid Boogie programs generated for programs with memset/memcpy when --bit-precise is used
Consider the following program,

```C
#include <string.h>
#include "smack.h"

int main(void) {
  unsigned long long x = 0;
  memset(&x, 1, sizeof(x));
  assert(x == 1);
  return 0;
}
```

SMACK generates a syntactically incorrect Boogie program.
This is due to sea-dsa only records one accessed type so `isTypeSafe` returns true for that node (https://github.com/seahorn/sea-dsa/issues/88).

shaobo-he:
Fixed #589

### DIFF
--- a/lib/smack/DSAWrapper.cpp
+++ b/lib/smack/DSAWrapper.cpp
@@ -126,7 +126,8 @@ bool DSAWrapper::isTypeSafe(const Value *v) {
   auto node = getNode(v);
 
   if (node->isOffsetCollapsed() || node->isExternal() || node->isIncomplete() ||
-      node->isUnknown() || node->isIntToPtr() || node->isPtrToInt())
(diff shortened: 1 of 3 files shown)


zvonimir:
How come this works when bit-vectors are not used?
Also, is this going to cause slow-downs or do things remain more or less the same?

shaobo-he:
> How come this works when bit-vectors are not used?
> Also, is this going to cause slow-downs or do things remain more or less the same?

It works since all the `i*` types are essentially type aliases but with bit-precise, the store function types are not consistent with the map types. I doubt it will cause a lot of slow-downs since memOpd memory regions normally should have `bv8` types.

PR #591, opened 2020-07-14
Format Python scripts to conform to the PEP 8 style guide

shaobo-he:
This PR excludes SV-COMP scripts because they are going to be changed soon.


zvonimir:
How are we going to make sure we maintain this? Could we run the check automatically like clang-format?

shaobo-he:
> How are we going to make sure we maintain this? Could we run the check automatically like clang-format?

Good question. I was thinking about it. Maybe we can add https://github.com/PyCQA/pycodestyle to our CI.

zvonimir:
I think I like this one better: https://github.com/google/yapf
It is supported by Google. It should be easy to set up since it resemble clang-format.

shaobo-he:
> I think I like this one better: https://github.com/google/yapf
> It is supported by Google. It should be easy to set up since it resemble clang-format.

It think when yapf shows nothing to do, the code still may not be PEP8-compliant. https://github.com/google/yapf#i-still-get-non-pep8-compliant-code-why. So I think if our goal is to be PEP8-compliant, we should use pycodestyle.

zvonimir:
pycodestyle is also not PEP8-compliant: https://pycodestyle.pycqa.org/en/latest/intro.html#disclaimer

I guess I don't care which tool we use. yapf seems to be trivial to add to Travis CI as checker and hence I suggested it. It looks like you just have to add a line in there and you are done - super easy. But if you are set on pycodestyle, sure, add it to Travis CI - my quick search did not reveal how to do that as it did with yapf.

PR #592, opened 2020-07-17
Apply rustfmt to reformat SMACK's Rust programs

shaobo-he:


zvonimir:
Can we enable checking in travis like we do with everything else?

shaobo-he:
> Can we enable checking in travis like we do with everything else?

It's possible but not ideal  if we keep the current Rust installation method. The reason is that rustfmt is a component of  the Rust tool chain but not a package. It's recommended to be installed using rustup which we do not have. What's worse, the particular nightly version SMACK uses does not have rustfmt in its tool chain.

zvonimir:
Got it. So I'll merge this one, and you should create an issue to add rustfmt later on to travis, once we figure out rustup.

PR #593, opened 2020-07-17
Improve Rust panics

keram88:
This improves detection and handling of panics arising in Rust programs.

### DIFF
--- a/lib/smack/Naming.cpp
+++ b/lib/smack/Naming.cpp
@@ -62,8 +62,10 @@ const std::string Naming::REC_MEM_OP = "boogie_si_record_mop";
 const std::string Naming::MEM_OP_VAL = "$MOP";
 
 const std::string Naming::RUST_ENTRY = "_ZN3std2rt10lang_start";
-const std::string Naming::RUST_PANIC1 = "_ZN4core9panicking5panic";
-const std::string Naming::RUST_PANIC2 = "_ZN3std9panicking11begin_panic";
+const std::vector<std::string> Naming::RUST_PANICS = {
+    "_ZN3std9panicking15begin_panic_fmt17h", "_ZN4core9panicking5panic17h",
+    "_ZN3std9panicking11begin_panic17h", "_ZN4core9panicking9panic_fmt17h"};
+
 const std::string Naming::RUST_PANIC_ANNOTATION = "rust_panic";
 
 const std::string Naming::BLOCK_LBL = "$bb";
--- a/lib/smack/SmackInstGenerator.cpp
+++ b/lib/smack/SmackInstGenerator.cpp
@@ -137,6 +137,17 @@ void SmackInstGenerator::annotate(llvm::Instruction &I, Block *B) {
   }
 }
 
+bool isRustPanic(const std::string &name) {
+  for (const auto &panic : Naming::RUST_PANICS) {
+    // We are interested in exact functional matches.
+    // Rust mangled names include a 17 byte hash at the end.
+    if (name.find(panic) == 0 && name.size() == panic.size() + 17) {
+      return true;
+    }
+  }
+  return false;
+}
+
 void SmackInstGenerator::processInstruction(llvm::Instruction &inst) {
   SDEBUG(errs() << "Inst: " << inst << "\n");
   annotate(inst, currBlock);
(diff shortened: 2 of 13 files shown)


shaobo-he:
Quick question: panic checking is always enabled, right?

keram88:
> Quick question: panic checking is always enabled, right?

That is the intended behavior. Do you have something in mind?

zvonimir:
@keram88 and @shaobo-he : my hunch is that this should not always be enabled. There should be a command line option to either enable it or maybe to disable it (if it is enabled by default). No?
Otherwise is some scenarios users might get tons of false bugs. Yeah, if I think about it more, I would say that it should not be enabled by default. Thoughts?

lib/smack/Naming.cpp
```diff
@@ -62,8 +62,11 @@ const std::string Naming::REC_MEM_OP = "boogie_si_record_mop";
 const std::string Naming::MEM_OP_VAL = "$MOP";
 
 const std::string Naming::RUST_ENTRY = "_ZN3std2rt10lang_start";
-const std::string Naming::RUST_PANIC1 = "_ZN4core9panicking5panic";
-const std::string Naming::RUST_PANIC2 = "_ZN3std9panicking11begin_panic";
+const std::vector<std::string> Naming::RUST_PANICS = {
+    "_ZN3std9panicking11begin_panic17h",
+    "_ZN3std9panicking15begin_panic_fmt17h", "_ZN4core9panicking5panic17h",
```
zvonimir, line 67: Any particular reason why these two share one line? Maybe just put them one per line.

lib/smack/SmackInstGenerator.cpp
```diff
@@ -137,6 +137,17 @@ void SmackInstGenerator::annotate(llvm::Instruction &I, Block *B) {
   }
 }
 
+bool is_rust_panic(const std::string &name) {
```
zvonimir, line 140: We are using camel case for function names, no?

lib/smack/SmackInstGenerator.cpp
```diff
@@ -640,11 +651,12 @@ void SmackInstGenerator::visitCallInst(llvm::CallInst &ci) {
...
     emit(Stmt::call(mainFunction->getName(), {}, {}));
 
-  } else if (name.find(Naming::RUST_PANIC1) != std::string::npos ||
-             name.find(Naming::RUST_PANIC2) != std::string::npos) {
+  } else if (is_rust_panic(name)) {
     // Convert Rust's panic functions into assertion violations
     emit(Stmt::assert_(Expr::lit(false),
                        {Attr::attr(Naming::RUST_PANIC_ANNOTATION)}));
+    // Block further assertions on this path
+    emit(Stmt::assume(Expr::lit(false)));
```
zvonimir, line 659: Why do we need this assumption? Please clarify. Thx!

test/rust/panic/core_panic_fail.rs
```diff
@@ -0,0 +1,20 @@
...
+#[start]
+fn main(_x: isize, _y: *const *const u8) -> isize {
+    panic!();
+}
+
+#[panic_handler]
+fn panic(_expr: &PanicInfo) -> !{
+    loop {}
+}
+#[lang="eh_personality"]
```
zvonimir, line 17: Empty line missing it seems.

test/rust/panic/core_panic_fmt_fail.rs
```diff
@@ -0,0 +1,20 @@
...
+#[start]
+fn main(_x: isize, _y: *const *const u8) -> isize {
+    panic!("Something {}", 7);
+}
+
+#[panic_handler]
+fn panic(_expr: &PanicInfo) -> !{
+    loop {}
+}
+#[lang="eh_personality"]
```
zvonimir, line 17: Empty line missing it seems.

shaobo-he:
> @keram88 and @shaobo-he : my hunch is that this should not always be enabled. There should be a command line option to either enable it or maybe to disable it (if it is enabled by default). No?
> Otherwise is some scenarios users might get tons of false bugs. Yeah, if I think about it more, I would say that it should not be enabled by default. Thoughts?

That's my thought too. Another thing to notice is that the verifier may spend a lot of time trying to prove the panics are unreachable instead of focusing on the assertions.

lib/smack/Naming.cpp
keram88: This is what clang format wanted.

keram88:
I'll add functionality to make Rust panic opt in.

lib/smack/SmackInstGenerator.cpp
keram88: I realized that this is what will happen anyway. Panics are always terminal.

keram88:
Do we want to disable assertion checking when checking for panics?

PR #594, opened 2020-07-20
Flake8

shaobo-he:
`flake8` (https://flake8.pycqa.org/en/latest/) is Python linter that incorporates `pycodestyle` which we previously used for checking conformality to PEP8. This PR fixes the issues reported by flake8 and also adds it the CI.

### DIFF
--- a/share/smack/doctor.py
+++ b/share/smack/doctor.py
@@ -128,7 +128,7 @@ def main():
     check_command("llvm2bpl")
     check_command("smack")
 
-    if args.prefix != '':
+    if not args.prefix:
         check_headers(args.prefix)
 
     exit(count)
--- a/share/smack/doctor.py
+++ b/share/smack/doctor.py
@@ -8,7 +8,6 @@ from subprocess import Popen, PIPE
 import sys
 import re
 import argparse
-import platform
 
 
 def red(text):
@@ -129,7 +128,7 @@ def main():
     check_command("llvm2bpl")
     check_command("smack")
 
-    if args.prefix is not '':
+    if args.prefix != '':
         check_headers(args.prefix)
 
     exit(count)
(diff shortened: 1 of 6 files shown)


share/smack/frontend.py
```diff
@@ -249,16 +251,17 @@ def json_compilation_database_frontend(input_file, args):
                 bit_codes = [re.sub('[.]o$', '.bc', f) for f in cc['objects']]
                 try_command(['llvm-link', '-o', args.bc_file] + bit_codes)
                 try_command(['llvm-link', '-o', args.linked_bc_file,
-                             args.bc_file] + build_libs(args))
+                             args.bc_file] + default_build_libs(args))
 
             else:
-                out_file = output_flags.findall(cc['command'])[0] + '.bc'
+                # out_file = output_flags.findall(cc['command'])[0] + '.bc'
```
zvonimir, line 257: Why did you comment out this one?

shaobo-he: `out_file` was never used and it led to a warning.

zvonimir: So why not delete it completely?

shaobo-he: > So why not delete it completely?

I thought maybe this line contains some information may be useful to revive this feature.

zvonimir: What does it do? What is its purpose?

share/smack/doctor.py
```diff
@@ -129,7 +128,7 @@ def main():
     check_command("llvm2bpl")
     check_command("smack")
 
-    if args.prefix is not '':
+    if args.prefix != '':
```
zvonimir, line 131: Let's rewrite this as `if not args.prefix` since that is it seems that pep8 prescribes:
https://stackoverflow.com/questions/9573244/how-to-check-if-the-string-is-empty

share/smack/frontend.py
zvonimir: Ok, we agreed to remove this line.

PR #595, opened 2020-07-22
Implemented check option for selecting properties to check

zvonimir:
This cleans up the interaction between checking of various
properties. A user can select all properties they would like
for SMACK to check on their code.

### DIFF
--- a/share/smack/frontend.py
+++ b/share/smack/frontend.py
@@ -85,9 +85,10 @@ def default_clang_compile_command(args, lib=False):
     cmd += ['-I' + path for path in smack_headers(args)]
     cmd += args.clang_options.split()
     cmd += ['-DMEMORY_MODEL_' + args.mem_mod.upper().replace('-', '_')]
-    if args.memory_safety:
+    if ('memory-safety' in args.check or 'valid-deref' in args.check or
+            'valid-free' in args.check or 'memleak' in args.check):
         cmd += ['-DMEMORY_SAFETY']
-    if args.integer_overflow:
+    if 'integer-overflow' in args.check:
         cmd += (['-fsanitize=signed-integer-overflow,shift']
                 if not lib else ['-DSIGNED_INTEGER_OVERFLOW_CHECK'])
     if args.float:
@@ -313,7 +314,7 @@ def default_build_libs(args):
     if args.pthread:
         libs += ['pthread.c']
 
-    if args.strings or args.memory_safety or args.integer_overflow:
+    if args.strings:
         libs += ['string.c']
 
     if args.float:
--- a/share/smack/top.py
+++ b/share/smack/top.py
@@ -285,14 +285,14 @@ def arguments():
     translate_group.add_argument(
         '--entry-points',
         metavar='PROC',
-        nargs='*',
+        nargs='+',
         default=['main'],
         help='specify top-level procedures [default: %(default)s]')
 
     translate_group.add_argument(
         '--check',
         metavar='PROPERTY',
-        nargs='*',
+        nargs='+',
         choices=['assertions', 'memory-safety', 'valid-deref', 'valid-free',
                  'memleak', 'integer-overflow'],
         default=['assertions'],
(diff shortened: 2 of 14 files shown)


zvonimir:
SVCOMP script is a major problem here since it is extremely messy when it comes to memory safety checking.
@shaobo-he how should we go about this? I haven't touched that code, but it does have to be updated.

share/smack/frontend.py
```diff
@@ -313,7 +314,7 @@ def default_build_libs(args):
     if args.pthread:
         libs += ['pthread.c']
 
-    if args.strings or args.memory_safety or args.integer_overflow:
```
keram88, line 316: Why were we enabling this for memory safety and overflow checking?

share/smack/frontend.py
```diff
@@ -313,7 +314,7 @@ def default_build_libs(args):
...
+    if args.strings:
         libs += ['string.c']
```
keram88, line 318: Why were we enabling this for memory safety and overflow checking?

zvonimir: Yeah, I was using git blame to try to figure it out, but no luck. I think it must be related to SVCOMP benchmarks.
Since I would like to completely rewrite our SVCOMP script, I figured I might as well clean this up too.

shaobo-he:
> SVCOMP script is a major problem here since it is extremely messy when it comes to memory safety checking.
> @shaobo-he how should we go about this? I haven't touched that code, but it does have to be updated.

I propose we remove that part of logic and replace it with enabling `--memory-safety` only.

share/smack/top.py
```diff
@@ -610,13 +589,15 @@ def annotate_bpl(args):
...
 def property_selection(args):
-    selected_props = []
-    if args.only_check_valid_deref:
-        selected_props.append('valid_deref')
-    elif args.only_check_valid_free:
-        selected_props.append('valid_free')
-    elif args.only_check_memleak:
-        selected_props.append('valid_memtrack')
+    selected_props = {}
+    if 'memory-safety' in args.check:
```
shaobo-he, line 593: I think that this function can simply return here.

share/smack/top.py
```diff
@@ -610,13 +589,15 @@ def annotate_bpl(args):
 
 
 def property_selection(args):
```
shaobo-he, line 591: Maybe this function is better named as `memory_safety_property_selection`

zvonimir:
Sorry, which documentation?

shaobo-he:
This one: https://github.com/smackers/smack/blob/master/docs/running.md#running-smack

share/smack/top.py
zvonimir: Done. Good point.

share/smack/top.py
zvonimir: Good point. Done.

zvonimir:
Yeah, wrt documentation, we are missing documentation on all of the command line options where this would be described. So I think I won't change it now. I hope we are ready to get this merged. Then @keram88 can build on top of this and add `rust-panic` into properties we can check.

share/smack/top.py
```diff
@@ -609,14 +588,16 @@ def annotate_bpl(args):
...
-    selected_props = []
-    if args.only_check_valid_deref:
-        selected_props.append('valid_deref')
-    elif args.only_check_valid_free:
-        selected_props.append('valid_free')
-    elif args.only_check_memleak:
-        selected_props.append('valid_memtrack')
+def memsafety_subproperty_selection(args):
+    selected_props = {}
+    if 'memory-safety' in args.check:
```
shaobo-he, line 593: Actually, we don't even need this statement 😂

zvonimir: I think we do since you can have both `memory-safety` and `valid-deref` selected. That will lead to a wrong thing being checked if we remove this statement. No?

shaobo-he: Ah, yes.

PR #600, opened 2020-08-11
Decouple bit-precise-pointers from bit-precise

shaobo-he:
In the previous implementation, the former flag implies the latter flag, which
seems unnecessary.

### DIFF
--- a/lib/smack/Prelude.cpp
+++ b/lib/smack/Prelude.cpp
@@ -328,7 +328,7 @@ struct IntOpGen::IntArithOp : public IntOp {
         (!SmackOptions::BitPrecisePointers && alsoUsedByPtr))
       funcs.push_back(getIntFunc(size));
     if (SmackOptions::BitPrecise ||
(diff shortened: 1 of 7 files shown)


lib/smack/Prelude.cpp
```diff
@@ -327,7 +327,8 @@ struct IntOpGen::IntArithOp : public IntOp {
     if (!SmackOptions::BitPrecise ||
         (!SmackOptions::BitPrecisePointers && alsoUsedByPtr))
       funcs.push_back(getIntFunc(size));
-    if (SmackOptions::BitPrecise)
+    if (SmackOptions::BitPrecise ||
+        SmackOptions::BitPrecisePointers && alsoUsedByPtr)
```
zvonimir, line 331: Please add parenthesis so that we do not have to rely on operator precedence (see above).

shaobo-he: > Please add parenthesis so that we do not have to rely on operator precedence (see above).

Done.

Issue #434, opened 2019-04-17
Report inability of verification


bitcalc:
I want to use SMACK to achieve high confidence by formally verifying certain properties. This requires SMACK to report assertions that it fails to verify after runs. For example, I have code

```
0: int buf[10];
1: int i = 0, x;
2: while(complicated_confition) {
3: …
4: i++;
5: }
6: assert(i + 12 < 10);
7: x = buf[i + 12];
```

When I run SMACK with --unroll=2, it silently completes without any warning. SMACK cannot determine that the while loop exits in this run and finishes as if the verification succeeded. It makes us lose confidence in verified code when it completes.

A much-needed feature is to give a user feedback on which assertions that SMACK decides to give up, which could come in different flavors.

1. Ideally, SMACK should report which assertions it doesn't verify after a run.
2. If the above is too difficult, then SMACK should report the loops that it cannot determinate their exits so that we know assertions inside and/or after the loops are not verified.

This information warns a user that SMACK doesn't finish all verification tasks, although it doesn't help solve them. Without this information, a user doesn't know if a normal completion means a successful verification or trivial non-verification.

PR #601, opened 2020-08-13
Assert false on loop exit

jjgarzella:
This partly addresses #434 

This PR adds a pass that finds every loop exit and adds an `assert false` using `__SMACK_code` at each exit. The idea is that if a user is not sure if their loop is unrolling enough, then they can use the flag `--fail-on-loop-exit`, and if the program does not fail, that means the unroll bound is not enough.

Some technical notes:
* this pass requires the LoopSimplifyPass from LLVM to run before it does. Among other things, LoopSimplifyPass guarentees that no BasicBlock which is an "exit block" of the loop (that is, it has a predecessor block inside the loop) has a predecessor that is not inside the loop. If such a block existed, this would make a subtle bug where the assertion could fail even though the loop hadn't been unrolled enough. LoopSimplifyPass should keep this from happening. I haven't been able to come up with C code that compile (at level -O0) into this edge case, so I can't demonstrate the necessity of this pass yet.
* as this pass adds an `assert false` to every single loop exit it finds, if SMACK fails, what it really tells us is that ONE exit point of the FIRST loop is reachable. There could a few possible enhancements:
--> Right now, nested loops inside an outer loop are not considered. If SMACK runs and does not fail in the situation of a nested loop, there's no way to tell which loop wasn't unrolled enough. Modifying the pass so that the user can detect precisely which nested loop is the issue should be easy, however @shaobo-he and I aren't sure if it would be that useful, so it has not been implemented.
--> There could be multiple exit points to the loop, i.e. there could be a break statement (or goto, I guess). Because this pass can only tell if ONE of the exit points of the loop can be reached, there could be a loop which isn't unrolled enough for all possible inputs, but only for some inputs (see `test/c/unroll/two_exits.c`). Enhancing this would require some type of annotating the break statements or something, and I imagine it could be hard.
--> There could be multiple outer loops in the program (`two_loops.c` and `two_loops2.c`). In this case, because this pass adds assertions to all loops, as soon as it finds the exit to the first loop, it will terminate. If there is a second loop which requires greater unroll to exit than the first, SMACK won't be able to tell. There are a couple of ways one could think about trying to fix this. 

I'm interested in thoughts from @shaobo-he @keram88 @zvonimir and @michael-emmi on how we might enhance this functionality. In the meantime maybe we can merge this so we have something. This is also the first LLVM pass I've written, so I'm eager for feedback on coding style, etc.

### DIFF
--- a/include/smack/AnnotateLoopExits.h
+++ b/include/smack/AnnotateLoopExits.h
@@ -1,8 +1,8 @@
 //
 // This file is distributed under the MIT License. See LICENSE for details.
 //
-#ifndef ANNOTATELOOPENDS_H
-#define ANNOTATELOOPENDS_H
+#ifndef ANNOTATELOOPEXITS_H
+#define ANNOTATELOOPEXITS_H
 
 #include "llvm/IR/Instructions.h"
 #include "llvm/IR/Module.h"
@@ -11,14 +11,14 @@
 
 namespace smack {
 
-class AnnotateLoopEnds : public llvm::ModulePass {
+class AnnotateLoopExits : public llvm::ModulePass {
 public:
   static char ID; // Pass identification, replacement for typeid
-  AnnotateLoopEnds() : llvm::ModulePass(ID) {}
+  AnnotateLoopExits() : llvm::ModulePass(ID) {}
   virtual llvm::StringRef getPassName() const override;
   virtual bool runOnModule(llvm::Module &m) override;
   virtual void getAnalysisUsage(llvm::AnalysisUsage &) const override;
 };
 } // namespace smack
 
-#endif // ANNOTATELOOPENDS_H
+#endif // ANNOTATELOOPEXITS_H
--- a/lib/smack/AnnotateLoopExits.cpp
+++ b/lib/smack/AnnotateLoopExits.cpp
@@ -46,23 +46,22 @@ void processExitBlock(BasicBlock *block, Function *le) {
 
   SDEBUG(errs() << "Processing an Exit Block\n");
 
-  Instruction& front = block->front();
-  insertLoopEndAssertion(le,&front);
+  Instruction &front = block->front();
+  insertLoopEndAssertion(le, &front);
 }
 
 void annotateLoopEnd(Loop *loop, Function *le) {
-  //BasicBlock *headerBlock = loop->getHeader();
 
   // I imagine that it is very uncommon for a loop to have
   //   more than 5 exit points. This is a complete guess
   //   and we could probably have a better heuristic
-  SmallVector<BasicBlock *, 5> exitBlocks; 
+  SmallVector<BasicBlock *, 5> exitBlocks;
 
   loop->getExitBlocks(exitBlocks);
-  
+
   for (BasicBlock *b : exitBlocks) {
-    processExitBlock(b,le);
-  } 
+    processExitBlock(b, le);
+  }
 }
 
 bool AnnotateLoopExits::runOnModule(Module &m) {
@@ -90,5 +89,8 @@ bool AnnotateLoopExits::runOnModule(Module &m) {
 // Pass ID variable
 char AnnotateLoopExits::ID = 0;
 
-StringRef AnnotateLoopExits::getPassName() const { return "Annotate Loop Ends with assert(false)"; }
+StringRef AnnotateLoopExits::getPassName() const {
+  return "Annotate Loop Ends with assert(false)";
+}
+
 } // namespace smack
--- a/lib/smack/AnnotateLoopExits.cpp
+++ b/lib/smack/AnnotateLoopExits.cpp
@@ -8,8 +8,10 @@
 // is added between a branching block and the loop header.
 //
 
-#include "smack/AnnotateLoopEnds.h"
+#define DEBUG_TYPE "smack-loop-unroll"
+#include "smack/AnnotateLoopExits.h"
 #include "smack/Naming.h"
+#include "smack/Debug.h"
 #include "llvm/Analysis/LoopInfo.h"
 #include "llvm/Transforms/Utils.h"
 #include "llvm/IR/Constants.h"
@@ -29,11 +31,12 @@ namespace smack {
 using namespace llvm;
 
 // Register LoopInfo
-void AnnotateLoopEnds::getAnalysisUsage(AnalysisUsage &AU) const {
+void AnnotateLoopExits::getAnalysisUsage(AnalysisUsage &AU) const {
   AU.addRequiredID(LoopSimplifyID);
   AU.addRequired<LoopInfoWrapperPass>();
 }
 
+// This method is for clarity and self-documentingness
 void insertLoopEndAssertion(Function *le, Instruction *insertBefore) {
   CallInst::Create(le, "", insertBefore);
 }
@@ -41,7 +44,7 @@ void insertLoopEndAssertion(Function *le, Instruction *insertBefore) {
 void processExitBlock(BasicBlock *block, Function *le) {
   // print exit block found!!
 
-  errs() << "    Exit block found! \n";
+  SDEBUG(errs() << "Processing an Exit Block\n");
 
   Instruction& front = block->front();
   insertLoopEndAssertion(le,&front);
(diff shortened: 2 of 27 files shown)


include/smack/AnnotateLoopExits.h
```diff
@@ -0,0 +1,24 @@
...
+#define ANNOTATELOOPEXITS_H
+
+#include "llvm/IR/Instructions.h"
+#include "llvm/IR/Module.h"
+#include "llvm/Pass.h"
+#include <map>
+
+namespace smack {
+
+class AnnotateLoopExits : public llvm::ModulePass {
```
michael-emmi, line 14: Why should this be a module pass, as opposed to a more fine-grained pass, e.g., function or loop pass: https://llvm.org/docs/WritingAnLLVMPass.html#the-looppass-class

Generally speaking, you want to reduce the scope as much as possible — see docs above.

michael-emmi:
This is a great start. Regarding possible refinements, I wonder if it would be possible to annotate each loop exit with a distinct identifier, and then have smack report all assertion violations (instead of just one). Then you can identify loops which need more unrolling as the set of all loops minus the ones which hit assertion violations.

lib/smack/AnnotateLoopExits.cpp
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  return true;
+}
+
+// Pass ID variable
+char AnnotateLoopExits::ID = 0;
+
+StringRef AnnotateLoopExits::getPassName() const {
+  return "Annotate Loop Ends with assert(false)";
```
zvonimir, line 93: We are interchangeably using "loop end" and "loop exit". Let's stick with "loop exit" everywhere since I think that is the right terminology. Change this to "Annotate all loop exits with assert(false)".

lib/smack/AnnotateLoopExits.cpp
```diff
@@ -0,0 +1,96 @@
+//
+// This file is distributed under the MIT License. See LICENSE for details.
+//
+
+//
+// This pass normalizes loop structures to allow easier generation of Boogie
```
zvonimir, line 6: This comment seems to be copy-pasted from a previous pass. Update it or remove it.

lib/smack/AnnotateLoopExits.cpp
```diff
@@ -0,0 +1,96 @@
...
+using namespace llvm;
+
+// Register LoopInfo
+void AnnotateLoopExits::getAnalysisUsage(AnalysisUsage &AU) const {
+  AU.addRequiredID(LoopSimplifyID);
+  AU.addRequired<LoopInfoWrapperPass>();
+}
+
+// This method is for clarity and self-documentingness
+void insertLoopEndAssertion(Function *le, Instruction *insertBefore) {
```
zvonimir, line 40: Rename into insertLoopExitAssertion.

lib/smack/AnnotateLoopExits.cpp
```diff
@@ -0,0 +1,96 @@
...
+  CallInst::Create(le, "", insertBefore);
+}
+
+void processExitBlock(BasicBlock *block, Function *le) {
+  // print exit block found!!
```
zvonimir, line 45: Remove this comment.

lib/smack/AnnotateLoopExits.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  SDEBUG(errs() << "Processing an Exit Block\n");
+
+  Instruction &front = block->front();
+  insertLoopEndAssertion(le, &front);
+}
+
+void annotateLoopEnd(Loop *loop, Function *le) {
```
zvonimir, line 53: Rename into annotateLoopExit

lib/smack/AnnotateLoopExits.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  // I imagine that it is very uncommon for a loop to have
+  //   more than 5 exit points. This is a complete guess
+  //   and we could probably have a better heuristic
+  SmallVector<BasicBlock *, 5> exitBlocks;
```
zvonimir, line 58: Documentation seems to indicate that if you don't know the size ahead of time, you should set it to 0. Also, then you can remove the comment above.

lib/smack/AnnotateLoopExits.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  loop->getExitBlocks(exitBlocks);
+
+  for (BasicBlock *b : exitBlocks) {
+    processExitBlock(b, le);
+  }
+}
+
+bool AnnotateLoopExits::runOnModule(Module &m) {
+
```
zvonimir, line 68: Remove this empty line.

lib/smack/AnnotateLoopExits.cpp
```diff
@@ -0,0 +1,96 @@
...
+  Function *le = m.getFunction("__SMACK_loop_end");
+  assert(le != NULL && "Function __SMACK_loop_end shoudl be present.");
```
zvonimir, line 70: Typo - "shoudl"

lib/smack/AnnotateLoopExits.cpp
```diff
@@ -0,0 +1,96 @@
...
+  loop->getExitBlocks(exitBlocks);
+
+  for (BasicBlock *b : exitBlocks) {
+    processExitBlock(b, le);
+  }
+}
+
+bool AnnotateLoopExits::runOnModule(Module &m) {
+
+  Function *le = m.getFunction("__SMACK_loop_end");
```
zvonimir, line 69: Let's add this string constant into Naming.cpp, which is the convention we follow.

shaobo-he:
@jjgarzella Is it possible that you could address the comments soon? If not, I can take the PR over. The feature implemented in it is much needed.

Btw, I found a good reference by CBMC developers: http://www.cprover.org/cprover-manual/cbmc/unwinding/

share/smack/top.py
```diff
@@ -518,6 +518,12 @@ def arguments():
         default=False,
         help='enable support for string')
 
+    translate_group.add_argument(
+        '--fail-on-loop-exit',
```
shaobo-he, line 523: I think we may want a better name here. For example, CBMC's `--unwinding-assertions` makes more sense. Maybe we should call it `--enable-loop-unrolling-checks` ?

shaobo-he:
> This is a great start. Regarding possible refinements, I wonder if it would be possible to annotate each loop exit with a distinct identifier, and then have smack report all assertion violations (instead of just one). Then you can identify loops which need more unrolling as the set of all loops minus the ones which hit assertion violations.

This is a great suggestion. We need to think about how we can map assertions to loop locations. Another tricky part is Corral's multi-assertion checking scheme turns assertions into assumes, which block the execution. Let's say we have two loops in parallel and the loop bound is deep enough. If Corral finds the first assert false and turns it into assume false, the second assert false would be unreachable, causing a false result.

zvonimir:
Hi all. @jjgarzella and @shaobo-he what is happening with this pull request?
Please let me know. I'll try to get it merged ASAP based on what you report back.
Thanks both!

jjgarzella:
> 
> 
> Hi all. @jjgarzella and @shaobo-he what is happening with this pull request? Please let me know. I'll try to get it merged ASAP based on what you report back. Thanks both!

At this point, the pass now passes all of the tests in the `c/unroll` folder. CI keeps failing because of failing benchmarks in `c/contracts`, but I am unable to run them because somehow installing dependencies seems to be broken. I have no idea why these benchmarks would fail since they never use the new flag, and I don't know how to fix `apt` at the moment.

shaobo-he:
Let me take a look at them and get back to you.

shaobo-he:
@jjgarzella it appears that you may need add `__SMACK_loop_exit` to function `inlined_procedures` in `share/smack/top.py`.

lib/smack/AnnotateLoopExits.cpp
```diff
@@ -0,0 +1,94 @@
...
+  if (F.isIntrinsic() || F.empty()) {
+    return false;
+  }
+
+  LoopInfo &loopInfo = getAnalysis<LoopInfoWrapperPass>().getLoopInfo();
+  for (LoopInfo::iterator LI = loopInfo.begin(), LIEnd = loopInfo.end();
+       LI != LIEnd; ++LI) {
+
+    SDEBUG(errs() << "Processing Loop in " << F.getName() << "\n");
+    annotateLoopExit(*LI, LoopExitFunction);
```
zvonimir, line 81: Why is `LoopExitFunction` an argument here?
It is a field of this class, and so we can use it directly in the called functions instead of passing it as an argument.

PR #604, opened 2020-08-14
Add integer encoding and pointer encoding options

shaobo-he:
This PR replaces old `--bit-precise`, `--bit-precise-pointers`, and `--wrapped-integer-encoding` flags with new `--integer-encoding` and `--pointer-encoding` flags.

### DIFF
--- a/share/smack/frontend.py
+++ b/share/smack/frontend.py
@@ -95,7 +95,7 @@ def default_clang_compile_command(args, lib=False):
         cmd += ['-DFLOAT_ENABLED']
     if args.pthread:
         cmd += ['-DSMACK_MAX_THREADS=' + str(args.max_threads)]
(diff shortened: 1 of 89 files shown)


share/smack/frontend.py
```diff
@@ -95,8 +95,6 @@ def default_clang_compile_command(args, lib=False):
         cmd += ['-DFLOAT_ENABLED']
     if args.pthread:
         cmd += ['-DSMACK_MAX_THREADS=' + str(args.max_threads)]
-    if args.bit_precise:
```
zvonimir, line 98: How come we don't need this any more?

shaobo-he: Good catch. I grepped the wrong pattern. We did need it.

PR #613, opened 2020-09-03
Fixing discrepancy between bit-vector flag in llvm2bpl and Python script

### LINKED ISSUE #611: [Feedback] bit precision
It seems the old --bit-precise options has been replaced by `--integer-encoding bit-vector` in version 2.5.0.
If the options is not given, the integer encoding is used by default.
If the program has bitwise operations, the warning still suggests to use the old --bit-precision flag
```
warning: over-approximating bitwise operation xor; try adding flag(s): --bit-precise 
```
Probably you forgot to update the warning msg.

zvonimir:
Fixes #611


shaobo-he (approved):

shaobo-he:
LGTM.

PR #618, opened 2020-10-01
Update Rust compiler installation to use rust-up

keram88:
- The Rust compiler version for smack is now installed as a toolchain
- bin/versions is now installed as a python module for access from frontend.py
- Rust is no longer installed globally, and now installs in the preferred manner


shaobo-he (approved):
LGTM.

zvonimir (approved):
Looks good to me.

PR #619, opened 2020-10-02
Update handling of Rust entry

keram88:
- Removes the call to lang_start in the main function at the IR level
  * This enables certain optimizations that are not available when rewriting
    at the bpl level
- Remove lang_start functions from the IR
  * This allows dead code elimination to remove some globals that slow
    verification

### DIFF
--- a/lib/smack/Naming.cpp
+++ b/lib/smack/Naming.cpp
@@ -62,6 +62,8 @@ const std::string Naming::REC_MEM_OP = "boogie_si_record_mop";
 const std::string Naming::MEM_OP_VAL = "$MOP";
 
 const std::string Naming::RUST_ENTRY = "_ZN3std2rt10lang_start";
+const std::string Naming::RUST_LANG_START_INTERNAL =
+    "_ZN3std2rt19lang_start_internal";
 const std::vector<std::string> Naming::RUST_PANICS = {
     "_ZN3std9panicking15begin_panic_fmt17h", "_ZN4core9panicking5panic17h",
     "_ZN3std9panicking11begin_panic17h", "_ZN4core9panicking9panic_fmt17h",
--- a/lib/smack/RustFixes.cpp
+++ b/lib/smack/RustFixes.cpp
@@ -1,0 +1,99 @@
+//
+// This file is distributed under the MIT License. See LICENSE for details.
+//
+// This patches Rust programs by removing certain language specific functions,
+// enabling later optimizations.
+//
+
+#include "smack/RustFixes.h"
+#include "smack/Naming.h"
+#include "llvm/Analysis/CallGraph.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/InstIterator.h"
+#include "llvm/IR/Module.h"
+#include "llvm/Support/raw_ostream.h"
+#include <set>
+#include <string>
+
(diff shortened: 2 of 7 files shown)


lib/smack/RustFixes.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+  for (auto I : instToErase) {
+    I->eraseFromParent();
+  }
+}
+
+bool RustFixes::runOnModule(Module &m) {
+  std::set<Function *> funcToErase;
+
+  for (auto &F : m) {
```
shaobo-he, line 72: Maybe we can make it a function pass?

shaobo-he (approved):
LGTM.

lib/smack/RustFixes.cpp
keram88: Apparently not. From https://llvm.org/doxygen/classllvm_1_1FunctionPass.html#details:

```Optimizing a function does not cause the addition or removal of any functions in the module```

Comment (shaobo-he) [reply lib/smack/RustFixes.cpp:72]: > Apparently not. From https://llvm.org/doxygen/classllvm_1_1FunctionPass.html#details:
> 
> `Optimizing a function does not cause the addition or removal of any functions in the module`

I see. Is removing functions necessary for this pass to do? Maybe removing deadcode pass will do it for us.

File: lib/smack/Naming.cpp [hunk 2]
```diff
@@ -62,6 +62,8 @@ const std::string Naming::REC_MEM_OP = "boogie_si_record_mop";
 const std::string Naming::MEM_OP_VAL = "$MOP";
 
 const std::string Naming::RUST_ENTRY = "_ZN3std2rt10lang_start";
+const std::string Naming::RUST_LANG_START_INTERNAL =
```
zvonimir, line 65: Why do you need both `RUST_ENTRY` and `RUST_LANG_START_INTERNAL`?

lib/smack/RustFixes.cpp
```diff
@@ -0,0 +1,99 @@
...
+...
+*/
+void fixEntry(Function &main) {
+  std::vector<Instruction *> instToErase;
+
+  for (inst_iterator I = inst_begin(main), E = inst_end(main); I != E; ++I) {
+    if (auto ci = dyn_cast<CallInst>(&*I)) {
+      if (Function *f = ci->getCalledFunction()) {
+        std::string name = f->hasName() ? f->getName() : "";
+        if (name.find(Naming::RUST_ENTRY) != std::string::npos) {
```
zvonimir, line 41: How come `RUST_LANG_START_INTERNAL` is not mentioned in this if condition?

zvonimir (changes requested):

lib/smack/Naming.cpp
keram88: `RUST_ENTRY` is used for the function that appears in `@main` that calls the real main function. `RUST_LANG_START_INTERNAL` is a function which appears in the set up of closure functions.

lib/smack/RustFixes.cpp
keram88: This is aimed at fixing the entry point in the main function.

zvonimir (approved):
Looks good to me.

PR #625, opened 2020-10-14
Enumerize SMACK's Python Scripts

shaobo-he:
This PR intends to replace string literals such as SMACK's results and checks with Python's enum classes.

- [x] Result handling
- [x] `--check` option
- [x] Fix sv-comp scripts


zvonimir (approved):
I fine fine with the previous implementation too, but this also looks good to me.

keram88 (approved):
This looks good. I think it will be helpful in the long run to avoid our current, ad hoc approach.

zvonimir:
Are we ok to merge this? @shaobo-he ?
SVCOMP scripts changed a lot, and so we should probably take care of them in a separate pull request.

shaobo-he:
> Are we ok to merge this? @shaobo-he ?
> SVCOMP scripts changed a lot, and so we should probably take care of them in a separate pull request.

SVCOMP scripts won't work with the current implementation of the PR. I propose we wait until these scripts are stable.

zvonimir:
Ok, I merged in the SCVOMP cleanup pull request. Could you now wrap this up @shaobo-he ? Thx!

share/smack/svcomp/utils.py
```diff
@@ -48,15 +48,18 @@ def svcomp_check_property(args):
   if args.svcomp_property:
     with open(args.svcomp_property, "r") as f:
       prop = f.read()
+    from smack.top import VProperty
+    from smack.top import VResult
     if "valid-deref" in prop:
-      args.check = ['memory-safety']
+      args.check = VProperty.MEMORY_SAFETY
```
zvonimir, line 54: Should this be a list/set?

share/smack/svcomp/utils.py
```diff
@@ -73,7 +76,12 @@ def verify_bpl_svcomp(args):
   """Verify the Boogie source file using SVCOMP-tuned heuristics."""
   heurTrace = "\n\nHeuristics Info:\n"
 
-  if not 'memory-safety' in args.check and not 'memleak' in args.check and not 'integer-overflow' in args.check:
+  from smack.top import VProperty
+  from smack.top import VResult
+
+  if (VProperty.MEMORY_SAFETY not in args.check
```
zvonimir, line 82: We could simplify this by pulling `not` out.

zvonimir: You have this pattern I suggested just below.

share/smack/top.py
```diff
@@ -446,10 +611,8 @@ def arguments():
         args.bpl_file = 'a.bpl' if args.no_verify else temporary_file(
             'a', '.bpl', args)
 
-    if not args.check:
-        args.check = {'assertions'}
-    else:
-        args.check = set([val for sublist in args.check for val in sublist])
+    if args.check == VProperty.NONE:
+        args.check = VProperty.ASSERTIONS
```
zvonimir, line 615: Shouldn't this be a list/set?

zvonimir (changes requested):

share/smack/svcomp/utils.py
shaobo-he: No, `args.check` is an element of class `VProperty`.

zvonimir (approved):
Looks good to me.

share/smack/svcomp/utils.py
shaobo-he: If I understand it correctly, an enum like this in Python is basically a bitmask. So a list of properties like what we had amounts to a mask having certain bits set.

PR #633, opened 2020-10-31
Fix smack crate installation

keram88:
This fixes an issue where some files and dependencies weren't installed correctly, because of an erroneous merge.


shaobo-he:
Do we need to install `Cargo.toml` and `build.rs` as well?

shaobo-he:
Closed because #634 contains all the fixes in this PR.

PR #636, opened 2020-11-05
Add a pass to transform certain bit-wise operations into integer operations

keram88:
This pass approximates bit-wise operations by encoding them as integer operations. Specifically,
Masking by a constant value equal to 2^x-1, e.g., 0000111, is translated to a remainder operations.
A left shift by a constant is translated into a multiplication.
A right shift by a constant is translated into a division.


lib/smack/ConstantBVOps.cpp
```diff
@@ -0,0 +1,107 @@
...
+      BinaryOperator *bi = dyn_cast<BinaryOperator>(&*I);
+      Value *amt = bi->getOperand(1);
+      if (ConstantInt *ci = dyn_cast<ConstantInt>(amt)) {
+        const APInt &value = ci->getValue();
+        unsigned opcode = bi->getOpcode();
+        Instruction::BinaryOps op;
+        if (opcode == Instruction::AShr || opcode == Instruction::LShr) {
+          // Shifting right by a constant amount is equivalent to dividing by
+          // 2^amt
+          op = Instruction::SDiv;
```
keram88, line 52: I chose SDiv because it seems like these numbers might more often be signed. There's not a good reason I can think of to choose one or the other.

shaobo-he: The signed or unsigned division is the same when the rhs is a positive number, right?

shaobo-he:
A couple of feedback,

* Can we add a couple of regressions?
* Is this pass necessary when bv is enabled?

lib/smack/ConstantBVOps.cpp
keram88: I think so.

lib/smack/ConstantBVOps.cpp
```diff
@@ -31,6 +31,9 @@ unsigned getOpWidth(const Value *t) {
 }
 
 bool ConstantBVOps::runOnFunction(Function &f) {
+  if (!(SmackOptions::BitPrecise || SmackOptions::BitPrecisePointers))
```
shaobo-he, line 34: Sorry I missed this. Wouldn't be better if we move this guard to llvm2bpl.cpp so that this pass is independent of what SMACK is configured?

keram88: That might be a better place. On the other hand, it documents something about this pass.

lib/smack/ConstantBVOps.cpp
```diff
@@ -31,7 +31,7 @@ unsigned getOpWidth(const Value *t) {
 }
 
 bool ConstantBVOps::runOnFunction(Function &f) {
-  if (!(SmackOptions::BitPrecise || SmackOptions::BitPrecisePointers))
+  if (SmackOptions::BitPrecise || SmackOptions::BitPrecisePointers)
```
shaobo-he, line 34: I think we don't need to guard against `BitPrecisePointers` since these are integer operations.

lib/smack/ConstantBVOps.cpp
keram88: I think they are the same in integer modeling. One thing about it is if this were done in bv mode, we would need to differentiate between AShr and LShr, but I don't think this makes sense for bv to begin with.

keram88: I forgot we have the new models for integers. How are divisions handled?

lib/smack/ConstantBVOps.cpp
```diff
@@ -0,0 +1,107 @@
...
+      }
+    }
+    if (I->isBitwiseLogicOp()) {
+      // If the operation is a bit-wise `and' and the mask variable is constant,
+      // it may be possible to replace this operation with a remainder
+      // operation. If the rhs has only ones, and they're only in the least
+      // significant bit, then the mask is 2^(number of ones) - 1. This is
+      // equivalent to the remainder when dividing by 2^(number of ones).
+      BinaryOperator *bi = dyn_cast<BinaryOperator>(&*I);
+      Value *mask = bi->getOperand(1);
```
zvonimir, line 76: With bit-and, it would be good to handle a case where the other operand is constant, if that makes sense.

zvonimir (changes requested):

lib/smack/ConstantBVOps.cpp
```diff
@@ -0,0 +1,107 @@
...
+  return iType->getBitWidth();
+}
+
+bool ConstantBVOps::runOnFunction(Function &f) {
+  if (!(SmackOptions::BitPrecise || SmackOptions::BitPrecisePointers))
+    // Do not run this pass in bitvector mode.
+    return false;
+  std::vector<Instruction *> instsFrom;
+  std::vector<Instruction *> instsTo;
+  if (Naming::isSmackName(f.getName()))
```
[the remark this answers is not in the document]
keram88: We do this in the integer overflow pass.

lib/smack/ConstantBVOps.cpp
keram88: Good catch.

lib/smack/ConstantBVOps.cpp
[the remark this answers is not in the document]
zvonimir: Yep, this is fine. It returns empty string if there is no name.

lib/smack/ConstantBVOps.cpp
shaobo-he: > I forgot we have the new models for integers. How are divisions handled?

They have a wrapper around their arguments to convert them into proper unsigned/signed values.

zvonimir (approved):
Looks good to me.

shaobo-he (approved):

shaobo-he:
LGTM.

PR #637, opened 2020-11-07
Fixed invalid translation/assertion failure related to inline assembly

shaobo-he:
Inline assembly with side effects do not have functions being called values.
This triggers invalid translation by the `Devirt` pass and assertion failure in
`SmackInstGenerator`.


zvonimir (approved):
Looks good to me.

PR #638, opened 2020-11-07
Fixed incorrect reliance on instruction order for overflow checking

shaobo-he:
We used to assume two extractvalue instructions that use the overflow checking
intrinsics are consecutive, which may not be the case. This commit remove the
reliance on this assumption.


zvonimir (approved):
Looks good to me.

keram88 (approved):
Looks good to me. This should be right.

PR #642, opened 2020-11-13
Temporarily fixed LLVM intrinsics modeling

shaobo-he:
This commit always enables precise modeling of the overflow flag and does not
inject blocking assumes.


keram88 (approved):
Looks good to me.

zvonimir (approved):
Looks good to me.

PR #644, opened 2020-11-14
Do not emit warnings about bitwise operations of bit-width 1

shaobo-he:
We have precisely modeled them using axioms.


zvonimir (approved):
Looks good to me.

PR #654, opened 2020-11-22
GitHub actions

shaobo-he:


.github/workflows/smack-ci.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - run: sudo apt-get install -y apt-transport-https
+      - run: sudo apt-get update
+      - run: sudo apt-get install -y clang-${LLVM_SHORT_VERSION} clang-format-${LLVM_SHORT_VERSION} llvm-${LLVM_SHORT_VERSION}-dev dotnet-sdk-3.1
+      - run: sudo pip3 install -U flake8
+      - run: sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${LLVM_SHORT_VERSION} 20
+      - run: sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-${LLVM_SHORT_VERSION} 20
+      - run: sudo update-alternatives --install /usr/bin/llvm-config llvm-config /usr/bin/llvm-config-${LLVM_SHORT_VERSION} 20
+      - run: sudo update-alternatives --install /usr/bin/llvm-link llvm-link /usr/bin/llvm-link-${LLVM_SHORT_VERSION} 20
+      - run: sudo update-alternatives --install /usr/bin/llvm-dis llvm-dis /usr/bin/llvm-dis-${LLVM_SHORT_VERSION} 20
+      - run: sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-${LLVM_SHORT_VERSION} 20
```
michael-emmi, line 64: Maybe this list of 17 steps be better put into its own script(s)? Especially if this will continue to overlap with .travis.yml.

.travis.yml
```diff
@@ -2,6 +2,10 @@ language: generic
 os: linux
 dist: focal
 
+branches:
+  except:
+    - github-actions
+
```
michael-emmi, line 8: I’m guessing that you’ll remove this before merging?

CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ project(smack)
 
 if (NOT WIN32 OR MSYS OR CYGWIN)
 
-  file(STRINGS "bin/versions" LLVM_VERSION_STR REGEX "LLVM_SHORT_VERSION=[0-9]+")
+  file(STRINGS "bin/versions" LLVM_VERSION_STR REGEX "LLVM_SHORT_VERSION=\"[0-9]+\"")
```
michael-emmi, line 10: How is this change related? is it because workflows run with a different shell or version of CMake?

.github/workflows/smack-ci.yaml
zvonimir: Hmmm, build script installs these as well. I wonder why I pulled it out. Maybe to enable caching and speed up the installation process? Honestly, I can't remember any more, but maybe we don't need all of this at all, and instead we can just invoke the build script.

shaobo-he: I'm not an expert on shell scripts. But a benefit of running them individually as a `run` action is we easily know which command fails.

.travis.yml
shaobo-he: > I’m guessing that you’ll remove this before merging?

Yes, this line was added to avoid triggering Travis.

CMakeLists.txt
shaobo-he: No, we used to have assignments like `A=a` in `bin/versions` but now they are changed to `A="a"` so the old regex didn't work. It seems the variables in the CMakefil were still set correctly on Travis CI VMs but not on GitHub actions.

.github/workflows/smack-ci.yaml
zvonimir: Sure, but I think it is nice to avoid code duplication. I saw we remove this and just run the build script.

.github/workflows/smack-ci.yaml
```diff
@@ -0,0 +1,85 @@
...
+            "--exhaustive --folder=rust/structures --languages=rust",
+            "--exhaustive --folder=rust/vector --languages=rust"
+          ]
+    steps:
+      - uses: actions/checkout@v2
+
+      # add-ons
+      - run: sudo apt-get install -y cmake python3-yaml python3-psutil python3-pip unzip ninja-build libboost-all-dev 
+      # before_install:
+      - run: sudo rm -rf /usr/local/clang-7.0.0
```
zvonimir, line 48: I am not sure if we need this on github actions VMs. This was a hack to get it to work on Travis VMs that install clang by default. It might work without this.

bin/build.sh
```diff
@@ -459,6 +462,16 @@ if [ ${BUILD_LOCKPWN} -eq 1 ] ; then
 fi
 
 
+if [ ${INSTALL_DEV_DEPENDENCIES} -eq 1 ] ; then
+  sudo apt-get install -y python3-pip
+  sudo pip3 install -U flake8
+  sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-${LLVM_SHORT_VERSION} 30
```
zvonimir, line 468: Shouldn't clang format be installed first? It seems you are assuming it is already installed.

zvonimir (changes requested):

bin/build.sh
shaobo-he: Good catch. It seems clang-format is already installed by some step in our build script on the GitHub Actions machines.

zvonimir (approved):
Looks good to me.

PR #666, opened 2020-12-14
Make memory-safety instrumentations copy dbg metadata

shaobo-he:


zvonimir:
What does this do? I don't quite get it.

shaobo-he:
> What does this do? I don't quite get it.

This is to provide more precise error traces for memory-safety violations. If you run the regressions in this PR using the develop branch, you should notice that the error trace is weird.

zvonimir:
What happened with CI here?

test/c/memory-safety/test_sourceloc.c
```diff
@@ -0,0 +1,18 @@
+#include "smack.h"
+#include <stdlib.h>
+
+// @expect verified
+
+int main() {
```
zvonimir, line 6: This should be `int main(void)`

test/c/memory-safety/test_sourceloc_fail.c
```diff
@@ -0,0 +1,19 @@
+#include "smack.h"
+#include <stdlib.h>
+
+// @expect error
+// @checkout grep -v "test_sourceloc_fail.c(16,9)"
+
+int main() {
```
zvonimir, line 7: This should be `int main(void)`

zvonimir (changes requested):

shaobo-he:
> What happened with CI here?

It's related to this issue: https://github.com/smackers/smack/issues/668

zvonimir:
Hey @shaobo-he . I can't merge this until CI is fixed. What is happening there?

zvonimir:
How is it related to the issue you pointed out? I don't understand how come that regression works now?

zvonimir:
Ah, I see. You added a regression! Ok, issue #668 might be an easy fix. Can you take care of that one first then? Or how should we go about this?

shaobo-he:
> Ah, I see. You added a regression! Ok, issue #668 might be an easy fix. Can you take care of that one first then? Or how should we go about this?

Yes, I think I'll fix it first.

shaobo-he:
@zvonimir please see https://github.com/smackers/smack/pull/670.

zvonimir (approved):
Looks good to me.

PR #676, opened 2021-01-12
Refactored warning message APIs

### LINKED ISSUE #614: Weird naming convention in SmackWarnings
To me it seems that functions `warnIfUnsound` and `warnUnsound` are also reporting incompleteness warnings.
For example, this `warning: over-approximating bitwise operation and` is an incompleteness warning if we are indeed over-approximating this operation. So I am confused about the naming convention here.

To me it seems we should split this into `warnUnsound` and `warnIncomplete` functions. Thoughts?

shaobo-he:
This commit changes incorrectly named `warn.*Unsound` into `warnIfIncomplete` or
`warnUnModeled` to better reflect their meanings. In the meanwhile, the
incorretly named warning level `unsound` gets changed to `imprecise`.

Fixes #614

### DIFF
--- a/lib/smack/SmackInstGenerator.cpp
+++ b/lib/smack/SmackInstGenerator.cpp
@@ -319,9 +319,9 @@ void SmackInstGenerator::visitInvokeInst(llvm::InvokeInst &ii) {
   if (f) {
     emit(rep->call(f, ii));
   } else {
-    // llvm_unreachable("Unexpected invoke instruction.");
-    SmackWarnings::warnUnsound("invoke instruction", currBlock, &ii,
-                               ii.getType()->isVoidTy());
+    llvm_unreachable("Unexpected invoke instruction.");
+    // SmackWarnings::warnUnsound("invoke instruction", currBlock, &ii,
+    //                           ii.getType()->isVoidTy());
   }
   std::vector<std::pair<const Expr *, llvm::BasicBlock *>> targets;
   targets.push_back(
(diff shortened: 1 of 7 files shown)


lib/smack/SmackInstGenerator.cpp
```diff
@@ -319,9 +319,9 @@ void SmackInstGenerator::visitInvokeInst(llvm::InvokeInst &ii) {
   if (f) {
     emit(rep->call(f, ii));
   } else {
-    // llvm_unreachable("Unexpected invoke instruction.");
-    SmackWarnings::warnUnsound("invoke instruction", currBlock, &ii,
-                               ii.getType()->isVoidTy());
+    llvm_unreachable("Unexpected invoke instruction.");
```
zvonimir, line 322: What is the reason for this change?

zvonimir (changes requested):

lib/smack/SmackInstGenerator.cpp
shaobo-he: I think the original handling (i.e., `llvm_unreachable`) makes more sense since we want to be seriously warned (i.e., by a crash) when an invoke instruction invokes a non-function object.

zvonimir: Ok, then please remove the comments. No need to leave them behind. We should also probably test this on SVCOMP one it gets merged.

shaobo-he: Done.

zvonimir (approved):
Looks good to me.

PR #680, opened 2021-01-21
Update LLVM source locations to use github addresses

keram88:
LLVM has relocated the hosting for their source archives. This PR updates the locations in build.sh so LLVM can be built from source.


zvonimir (approved):
Looks good to me.

PR #681, opened 2021-01-22
Update sea-dsa version

keram88:
This updates the sea-dsa commit used by smack. This version of sea-dsa includes a performance fix for the `CompleteCallGraph` pass.


keram88:
Superseded by #686

PR #684, opened 2021-01-25
Enabled wrapped pointer arithmetic when integer encoding is wrapped integer

### LINKED ISSUE #682: False alarm when the integer encoding is wrapped integer
Consider the following program (a C version of `test/rust/array.rs`),
```C
#include "smack.h"
#include <stdlib.h>

int main() {
  int* arr = (int*)malloc(3*sizeof(int));
  unsigned long long idx = __VERIFIER_nondet_unsigned_long_long();
  assume(arr[0] < 4);
  assume(arr[1] < 5);
  assume(arr[2] < 6);
  assume(idx < 3);
  assert(arr[idx] <= 5);
}
```

SMACK verifies it when the default integer encoding is used while reports an error when wrapped integer issued. This is because `idx` can get a value `-2**64` which satisfies the assumptions ($ule.i64(idx, -1) && $uge.i64(idx, 0)) in `__VERIFIER_nondet_unsigned_long_long`. However, `arr[-2**64]` is clearly uninitialized and thus leads to an error.

One possible solution is to wrap the arguments of the pointer arithmetic using the `$tos` function.

shaobo-he:
Fixed #682

### DIFF
--- a/lib/smack/Prelude.cpp
+++ b/lib/smack/Prelude.cpp
@@ -1221,7 +1221,12 @@ void PtrOpGen::generatePreds(std::stringstream &s) const {
     const Expr *predExpr =
         SmackOptions::BitPrecisePointers
             ? condExpr
-            : new BinExpr(binPred, makePtrVarExpr(1), makePtrVarExpr(2));
+            : new BinExpr(
+                  binPred,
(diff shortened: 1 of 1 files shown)


zvonimir:
Is this ready? Could you create a regression for it?

shaobo-he:
> Is this ready? Could you create a regression for it?

No, it's actually far from being ready 😂

PR #687, opened 2021-02-04
Enable multiple assertion checking when Corral is used

shaobo-he:
Previously, when `--max-violations` is set to a value greater than
1, Corral still reports at most one error trace. This is
because Corral continues the verification by turning a failing
assertion into an assumption. Therefore, even though there may
exist multiple calls to assertion checking procedures such as
`__VERIFIER_assert`, Corral only reports one of them. This commit
deals with this issue by leveraging Boogie's `/printInstrumented`
option to inline procedures containing assertions.

### DIFF
--- a/share/smack/top.py
+++ b/share/smack/top.py
@@ -747,8 +747,12 @@ def llvm_to_bpl(args):
 def procedure_annotation(name, args):
     if name in args.entry_points:
         return "{:entrypoint}"
-    elif (args.modular and
-          re.match("|".join(inlined_procedures()).replace("$", r"\$"), name)):
+    elif ((args.verifier == 'corral' and args.max_violations > 1 and
+           re.match("__SMACK_check|__VERIFIER_assert", name)) or
+          (args.modular and
+           re.match("|".join(
+                             inlined_procedures()).replace("$", r"\$"),
+                    name))):
         return "{:inline 1}"
     elif (not args.modular) and args.verifier == 'boogie':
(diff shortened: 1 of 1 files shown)


zvonimir:
Question. Is this adding a dependency on Boogie even if a user wants to just use Corral?
Meaning, Boogie is maybe not available/installed in this case.

zvonimir:
Wouldn't a better solution be to inline using an LLVM pass? Or to force inlining somehow during compilation?

shaobo-he:
> Question. Is this adding a dependency on Boogie even if a user wants to just use Corral?
> Meaning, Boogie is maybe not available/installed in this case.

I think I saw a similar flag in Corral. Let me see if I can get it working.

shaobo-he:
> Wouldn't a better solution be to inline using an LLVM pass? Or to force inlining somehow during compilation?

Currently, using LLVM's inline pass is not an option because `__SMACK_check_memory_safety` is implemented as Boogie code.

shaobo-he:
> > Question. Is this adding a dependency on Boogie even if a user wants to just use Corral?
> > Meaning, Boogie is maybe not available/installed in this case.
> 
> I think I saw a similar flag in Corral. Let me see if I can get it working.

I can't find relevant options in Corral that archive similar goals.

zvonimir:
See: https://github.com/boogie-org/corral/issues/138

shaobo-he:
Closed this PR as the implementation is outdated.

PR #686, opened 2021-02-01
Only display sea-dsa's warning messages at SMACK's warning level info

shaobo-he:

PR #688, opened 2021-02-11
Rewrite undef pointers

keram88:
This adds a pass which converts certain pointers to `undef` values into pointers to `null`. This is meant to move the program into a more defined state for alias analysis. Because this pass changes semantics, it is locked behind the option `--rewrite-undef-pointers` in order to opt in to this transformation.

This pass builds off of #686 because it includes optimizations relevant to verification of linked Rust programs.

### DIFF
--- a/lib/smack/RewriteUndefPtrs.cpp
+++ b/lib/smack/RewriteUndefPtrs.cpp
@@ -8,7 +8,7 @@
 //===----------------------------------------------------------------------===//
 
 //===----------------------------------------------------------------------===//
-#include "smack/RemoveUndefPtrs.h"
+#include "smack/RewriteUndefPtrs.h"
 #include "llvm/IR/InstIterator.h"
 #include "llvm/Support/raw_ostream.h"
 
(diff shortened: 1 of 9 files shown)


lib/smack/RewriteUndefPtrs.cpp
```diff
@@ -0,0 +1,41 @@
...
+namespace smack {
+
+char RewriteUndefPtrs::ID = 0;
+static RegisterPass<RewriteUndefPtrs>
+    X("rewrite-undef-ptrs", "Transform undef pointers into null pointers");
+
+bool RewriteUndefPtrs::runOnFunction(Function &F) {
+  bool changed = false;
+  for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
+    for (auto OI = I->op_begin(), OE = I->op_end(); OI != OE; ++OI) {
```
shaobo-he, line 26: Is it possible that undefined pointers are not direct operands? Namely, they can show up in some expressions which are the operands of instructions. Maybe we need to do a BFS/DFS here.

keram88: This is possible. For example if we have `%1 = { i8*, i8* } undef` then extracted pointers are `undef`. This is why it appears after the SplitAggregateValue pass. A better solution might be to turn `undef` constants into calls to an undef initializer like what the sea-dsa devs mentioned.

PR #689, opened 2021-02-19
Improved Cargo support

keram88:
This pull request allows Smack to remove the definitions for Rust's panic functions. These are not well supported by Smack because they use calls to function pointers and otherwise handle exceptional control flow.

Additionally, the `RustFixes` pass is converted into a `FunctionPass` because this is the proper pass type for its current functionality.


shaobo-he (approved):
LGTM

zvonimir (approved):
Looks good to me.

PR #692, opened 2021-02-25
Made casts from llvm::StringRef to std::string explicit

### LINKED ISSUE #691: Upgrading SMACK to LLVM 11
Hey guys,

I just did a quick experiment to upgrade SMACK to LLVM 11. It seems the most challenging part is to handle LLVM's change where implicit cast from `llvm::StringRef` to `std::string` is removed. This motivates us to think about the question of whether we should just use `llvm::StringRef`.

shaobo-he:
This commit also replaces some usages of std::string with
llvm::StringRef.

Partially fixes #691

### DIFF
--- a/include/smack/Naming.h
+++ b/include/smack/Naming.h
@@ -5,6 +5,7 @@
 #ifndef NAMING_H
 #define NAMING_H
 
+#include "llvm/ADT/StringRef.h"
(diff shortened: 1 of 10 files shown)


zvonimir (approved):
Looks good to me.

PR #693, opened 2021-02-28
Use versioned LLVM binaries

shaobo-he:


bin/build.sh
```diff
@@ -480,7 +474,7 @@ if [ ${BUILD_SMACK} -eq 1 ] ; then
 
   mkdir -p ${SMACK_DIR}/build
   cd ${SMACK_DIR}/build
-  cmake ${CMAKE_INSTALL_PREFIX} -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_BUILD_TYPE=Debug .. -G Ninja
+  cmake ${CMAKE_INSTALL_PREFIX} -DCMAKE_BUILD_TYPE=Debug .. -G Ninja
```
zvonimir, line 477: Is this going to cause for build to be done with gcc?

format/run-clang-format.py
```diff
@@ -203,12 +203,19 @@ def print_trouble(prog, message, use_colors):
 
 
 def main():
+
+    with open(os.path.join(os.path.dirname(os.path.dirname(
```
zvonimir, line 207: Why don't you just import it like we do in `frontend.py`?

zvonimir (changes requested):

zvonimir:
Could you add nice error reporting if you run SMACK and e.g. `clang-10` is not present? How will the error look like in that case?

bin/build.sh
shaobo-he: If gcc is the default compiler the host OS uses, yes.

format/run-clang-format.py
shaobo-he: Because formatting should be doable even before SMACK is installed. If so, we haven't installed `bin/versions` as a Python module.

shaobo-he:
> Could you add nice error reporting if you run SMACK and e.g. `clang-10` is not present? How will the error look like in that case?

If `clang-10` is not available, then the compilation command should fail with the message the host prints forwarded.

shaobo-he:
> Could you add nice error reporting if you run SMACK and e.g. `clang-10` is not present? How will the error look like in that case?

Follow up on this question: we get error messages like this

```
SMACK program verifier version 2.6.3

Error invoking command:
clang-9 -c -emit-llvm -O0 -g -gcolumn-info -Xclang -disable-O0-optnone -I/usr/local/share/smack/include -DMEMORY_MODEL_NO_REUSE_IMPLS -fcolor-diagnostics -o /home/shaobo/smack-project/smack/simple-bx28p0us.bc test/c/basic/simple.c
[Errno 13] Permission denied: 'clang-9'
```

Does it look non-intuitive? I can check if the binaries exist and dump a better message if they don't.

zvonimir (approved):
Looks good to me.

PR #696, opened 2021-03-04
Add Cargo examples to regressions

shaobo-he:
I had to run them one by one otherwise concurrency issues of Cargo will be triggered, which leads to incorrect results.


zvonimir (approved):
Looks good to me.

PR #697, opened 2021-03-04
Enable support for lib target name specification in Cargo.toml

shaobo-he:


zvonimir:
Do examples we have here: https://github.com/smackers/smack/tree/master/examples/rust-cargo
still work?

shaobo-he:
> Do examples we have here: https://github.com/smackers/smack/tree/master/examples/rust-cargo
> still work?

Yes, it still works.

keram88:
This looks good to me. Maybe we'll have more issues like this, but deal with them as they come up.

keram88 (approved):

zvonimir (approved):
Looks good to me.

PR #698, opened 2021-03-10
Update SMACK to support LLVM 11

shaobo-he:


CMakeLists.txt
```diff
@@ -176,7 +176,11 @@ if (Boost_FOUND)
 endif ()
 # We have to import LLVM's cmake definitions to build sea-dsa
 # Borrowed from sea-dsa's CMakeLists.txt
-find_package (LLVM ${LLVM_SHORT_VERSION} CONFIG)
+# Borrowed from https://github.com/banach-space/llvm-tutor/commit/72cb20d058b9b3f51717c7a17607f7a4c7398642
+find_package (LLVM 11.1 REQUIRED CONFIG)
```
zvonimir, line 180: Does this have to be hard-coded now? That kind of sucks.

zvonimir (changes requested):
Quick question. SVCOMP run finished and it looks good.

CMakeLists.txt
shaobo-he: Maybe we can add minor version to it and specify 11.1.0 as the full llvm version in bin/versions.

zvonimir: I don't get it. Currently we have 3 versions floating around: 11, 11.0.1, and this weird 11.1. What is your suggestion exactly?

shaobo-he: We take the first 2 segments instead of 1 from `LLVM_FULL_VERSION` here. So if `LLVM_FULL_VERSION` is 11.1.0, we use `11.1` here whereas if it's 10.0.1, we use 10.0.

shaobo-he: I just pushed a new version. Could you take a look at it?

CMakeLists.txt
```diff
@@ -8,7 +8,11 @@ project(smack)
 if(NOT WIN32 OR MSYS OR CYGWIN)
 
   file(STRINGS "bin/versions" LLVM_VERSION_STR
-    REGEX "LLVM_SHORT_VERSION=\"[0-9]+\"")
+    REGEX "LLVM_FULL_VERSION=\"[0-9]+\.[0-9]+\.[0-9]+")
```
zvonimir, line 11: LLVM_FULL_VERSION is not being used anywhere in this file. Is that expected?

zvonimir (changes requested):

CMakeLists.txt
shaobo-he: Yes. Part of this variable is in `LLVM_EXTENDED_VERSION`.

zvonimir (approved):
Looks good to me.

PR #700, opened 2021-03-11
Update Docker stuff

shaobo-he:


.github/workflows/smack-ci.yaml
```diff
@@ -48,8 +48,74 @@ jobs:
...
+
       - name: compile and test
         env:
           TRAVIS_ENV: ${{ matrix.travis_env }}
         run: ./bin/build.sh
+
+  # Shaobo: I had to create a per-branch job because matrix doesn't work. The
+  # reason is that `if` expression is evaluated before matrix is expanded.
+  # See:
+  # https://github.community/t/matrix-cannot-be-used-in-jobs-level-if/17177/2
```
michael-emmi, line 60: Does this really justify the duplication?

I don’t see how matrix would have been the best solution anyways. This script will be run on each branch as is, since each branch is a separate workflow.

It seems like all you need here is to introduce a variable to hold the right Docker image tag, and to read that variable in the `docker_build` step. This variable would be set to `stable` or `latest`, depending on whether the branch is `master` or `develop`. Github actions seems capable of defining variables which depend on branch name, e.g., combining `set-env` or `set-output` with some shell command that generates the tag based on branch name.

https://stackoverflow.com/questions/58033366/how-to-get-current-branch-within-github-actions/58035262

https://www.techiediaries.com/get-current-branch-name-github-actions-github_ref-bash/

michael-emmi (changes requested):

.github/workflows/smack-ci.yaml
shaobo-he: Thank you for the suggestion. I've pushed an implementation based on `set-output` (`set-env` is deprecated.)

.github/workflows/smack-ci.yaml
```diff
@@ -48,8 +48,52 @@ jobs:
...
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Extract branch name
```
shaobo-he, line 75: cite the stackoverflow link

michael-emmi: lgtm

michael-emmi (approved):

docs/installation.md
```diff
@@ -57,11 +57,22 @@ vagrant destroy
 
 ### Quick Setup 2: Docker
 SMACK can also be run in a [Docker][] container. We tested the Dockerfile on
-the following configurations:
+the following configuration:
```
zvonimir, line 60: This should be plural, no? Since you do have two configurations.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 MAINTAINER Shaobo He <shaobo@cs.utah.edu>
```
zvonimir, line 2: I guess you maybe want to change your email address here.

docs/installation.md
```diff
@@ -57,11 +57,22 @@ vagrant destroy
...
-* Ubuntu 16.04, docker-ce version 18.09.7
-* OS X 10.14.5, Docker Desktop with Docker engine version 18.09.2
-* Windows 10, Docker Desktop with Docker engine version 18.09.2
+* Ubuntu 18.04, Docker version 19.03.6
+* Windows WSL Ubuntu 20.04, Docker Desktop with Docker engine version 20.10.2
+
+
+#### Docker Hub
+
+SMACK's Docker container images can be pulled from Docker Hub directly:
```
zvonimir, line 68: I think this should say:
The SMACK's Docker container images can be pulled from the Docker Hub directly:

zvonimir (changes requested):

docs/installation.md
keram88: "The SMACK's ..." reads strangely.

zvonimir: Ok, you are a native speaker :). If what @shaobo-he wrote looks good to you, let's role with that.

docs/installation.md
shaobo-he: Done.

zvonimir (approved):
Looks good to me.

PR #702, opened 2021-03-16
Add models for Rust's memory functions

keram88:
This adds models for Rust's memory allocations, including functions in the `std::alloc` crate and the more basic functions that these functions use, e.g., `__rust_alloc`. Calls to these functions are replaced with these models to improve verification times.


shaobo-he (approved):
LGTM. That being said, it would be much better if we could do it on the Rust level.

zvonimir (approved):
Great update!

PR #707, opened 2021-03-17
Removed the occurrences of `TRAVIS` in bin/build.sh

shaobo-he:
They are no longer needed.

### DIFF
--- a/.github/workflows/smack-ci.yaml
+++ b/.github/workflows/smack-ci.yaml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        travis_env:
+        regtest_env:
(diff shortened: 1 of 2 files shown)


zvonimir (changes requested):
You forgot one more place.

zvonimir (changes requested):
There is also TRAVIS_ENV. Maybe we can remove that one too.

shaobo-he:
> You forgot one more place.

Good catch. Fixed.

shaobo-he:
> There is also TRAVIS_ENV. Maybe we can remove that one too.

This variable is actually used by GitHub Actions. So I renamed it.

zvonimir (approved):
Looks good to me.

PR #712, opened 2021-03-21
Fix name matching for Rust functions

keram88:
This improves matching for Rust function names by requiring they match the syntax of Rust hashed function names or requiring an exact match.

### DIFF
--- a/lib/smack/RustFixes.cpp
+++ b/lib/smack/RustFixes.cpp
@@ -20,7 +20,7 @@ using namespace llvm;
 
 bool replaceRustMemoryFunctions(Function &f) {
   bool changed = false;
-  static const std::map<std::string, std::string> alloc_fns = {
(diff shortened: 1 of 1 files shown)


shaobo-he (approved):
LGTM.

lib/smack/RustFixes.cpp
```diff
@@ -18,6 +18,17 @@ namespace smack {
 
 using namespace llvm;
 
+bool isRustNameMatch(StringRef search, StringRef name) {
```
shaobo-he, line 21: Actually this can be written as `return a || b`.

keram88: I thought it would be less readable, but I think this is an improvement.

zvonimir (approved):
Looks brittle, but let's merge it in.

zvonimir:
@keram88 This has conflicts with develop. Could you please resolve those? Thanks!

Issue #713, opened 2021-03-24
Allow targetted memory-safety checking


keram88:
Hello,

SMACK's memory-safety checking currently applies to every function which accesses memory. However, this behavior is not always necessary. For example, some languages such as Rust have have "safe" subsets that can be assumed to be memory-safe. In this circumstance, it may be desirable to only check memory safety on functions which are determined to be potentially unsafe.

Because it may be of some use for other programming languages where only a few functions may be determined to be potentially unsafe, this could be implemented by:
1) Adding a new option for indicating to SMACK which functions are interesting for checking memory safety (like the current `--entry-points` options).
2) Change the `MemorySafety` pass to only annotate the functions from 1.

Beyond implementation details, it is not clear what soundness issues could arise even if only truly unsafe functions were checked, i.e., could unsafe and unsafe code interact through global memory?

zvonimir:
This is a great feature to implement, and we've seen the need for this in the past as well. I think we should maybe briefly meet to discuss this.

shaobo-he:
I think this also enables some optimizations such as if a pointer points to a statically known size of memory.

zvonimir:
I don't understand how this feature would enable that?

zvonimir:
This is done. Closing.

PR #715, opened 2021-04-06
Add an option to enable targeted checking of specific functions

keram88:
This adds the ability to selectively check certain functions in a program, specifically:
* This adds the option `--checked-functions' which is a list of functions
  on which to do checking
* Not specifying this option enables all checks in every function
This partially addresses #713

### DIFF
--- a/.github/workflows/smack-ci.yaml
+++ b/.github/workflows/smack-ci.yaml
@@ -23,6 +23,7 @@ jobs:
             "--exhaustive --folder=c/pthread_extras",
             "--exhaustive --folder=c/strings",
             "--exhaustive --folder=c/special",
+            "--exhaustive --folder=c/targeted-checks",
             "--exhaustive --folder=rust/array --languages=rust",
             "--exhaustive --folder=rust/basic --languages=rust",
             "--exhaustive --folder=rust/box --languages=rust",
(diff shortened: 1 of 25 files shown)


shaobo-he (commented):
Quick question: can we use a list of  regex as the argument instead of a list of strings?

shaobo-he:
The reason I asked this question is we can, though not ideal, disable checking on certain functions via regex negative look ahead

keram88:
This actually might be very useful for e.g. matching mangled functions.

michael-emmi:
+1 for regex.

You might also (a) consider using regex disjunction instead of a list — whatever is more convenient — and (b) consider whether matching happens at the assembly level — i.e. with "flat" and potentially mangled names — or the source or IR level — i.e. with potentially structured names like "MyModule::myFunction".

keram88:
I think I like leaving it as a list of space separated list to remain consistent with the `--entry-points` flag and allow it to be a list of regex patters. A potential concern is that `$` and `.` are valid characters in identifiers, e.g., used by Kotlin and Rust respectively.

I will have to think more about matching on unmangled names since languages have different mangling schemes (Rust doesn't quite demangle with `c++filt` for example.)

zvonimir (approved):
Looks good to me.

keram88:
I implemented specifying the functions to check using regular expressions. I made it so that the entire function name must be matched in order to avoid surprises should someone what just one function. However, I'm not sure this is necessarily the best way to do it.

zvonimir (approved):
Looks good to me.

shaobo-he:
LGTM.

shaobo-he (approved):

PR #725, opened 2021-04-27
Fix a undefined behavior in `ExtractContracts.cpp`

### LINKED ISSUE #724: Undefined behavior in `ExtractContracts.cpp`
See:

https://github.com/smackers/smack/blob/465280b2433d3b61d1ebfc86f4ffb39ad0c047f3/lib/smack/ExtractContracts.cpp#L95

`LI.getLoopFor(B)` could be a nullptr if `B` is not in `LI`. We haven't hit a crash probably because it's a method call that doesn't access `this`.

shaobo-he:
Fixes #724

### DIFF
--- a/lib/smack/ExtractContracts.cpp
+++ b/lib/smack/ExtractContracts.cpp
@@ -92,10 +92,12 @@ std::tuple<BlockList, LoopMap> splitContractBlocks(Function &F, LoopInfo &LI) {
                           << "\n");
             BasicBlock *B = CI->getParent();
             auto NewBB = B->splitBasicBlock(++CI->getIterator());
-            LI.getLoopFor(B)->addBasicBlockToLoop(NewBB, LI);
+            auto L = LI.getLoopFor(B);
+            if (L)
+              L->addBasicBlockToLoop(NewBB, LI);
             blocks.push_front(NewBB);
 
-            if (auto L = LI[B])
+            if (L)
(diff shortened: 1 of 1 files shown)


lib/smack/ExtractContracts.cpp
```diff
@@ -92,10 +92,12 @@ std::tuple<BlockList, LoopMap> splitContractBlocks(Function &F, LoopInfo &LI) {
                           << "\n");
             BasicBlock *B = CI->getParent();
             auto NewBB = B->splitBasicBlock(++CI->getIterator());
-            LI.getLoopFor(B)->addBasicBlockToLoop(NewBB, LI);
+            auto L = LI.getLoopFor(B);
+            if (L)
```
zvonimir, line 96: Should this be an assertion or an if condition? Is the code going to work properly if L is null?

zvonimir (changes requested):

lib/smack/ExtractContracts.cpp
shaobo-he: It should be an if condition because `B` may not be in any loop.

@michael-emmi could you confirm this?

zvonimir: I double-checked this. @shaobo-he is right and I think this is the correct fix.

zvonimir (approved):
Looks good to me.

PR #726, opened 2021-04-27
Upgrade SMACK to support LLVM 12

shaobo-he:


keram88 (approved):
LGTM

zvonimir:
@shaobo-he Could you please update `installation.md` to reflect the updated LLVM and clang versions?
Please do that so that we can merge this. Thanks!

shaobo-he:
> @shaobo-he Could you please update `installation.md` to reflect the updated LLVM and clang versions?
> Please do that so that we can merge this. Thanks!

Done.

zvonimir:
@shaobo-he could we keep demangling and introduce a command line option for it?
I guess you removed it since it was messing up SVCOMP. Is that right?
To me this seems like a nice functionality to have when manually debugging traces. No?
Thx!

shaobo-he:
> @shaobo-he could we keep demangling and introduce a command line option for it? I guess you removed it since it was messing up SVCOMP. Is that right? To me this seems like a nice functionality to have when manually debugging traces. No? Thx!

I removed it because I was kind of dumb to run demangler for each token (i.e., function name and variable name) :D 
We can run demangler just once and it shouldn't cause too much performance degration. If it looks reasonable, then I can bring this functionality back.

zvonimir:
Got it. Sure, it would be nice to have it I think.

PR #728, opened 2021-04-30
Enable control of build.sh installation instructions via the environment

keram88:
This PR does the following:
* All installation flags can now be provided through the environment, enabling control over what is installed
* Added the `INSTALL_LLVM` flag (default enabled) to control whether we install LLVM. This is for situations where a user already has an appropriate installation
* Fix the `PATH` environment variable when doing a prefix build so that it doesn't copy the path from the build environment.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -22,21 +22,22 @@
 ################################################################################
 
 # Set these flags to control various installation options
-INSTALL_DEPENDENCIES=1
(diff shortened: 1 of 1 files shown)


shaobo-he (approved):
LGTM.

zvonimir (approved):
Looks good to me.

PR #733, opened 2021-05-04
Add allocation size attribute for global variables

### LINKED ISSUE #628: Global structures
Whenever there is a global array, the Boogie code has something like `const {:count X}` where `X` is the size of the array.
This works perfectly for me because I can use it as a hint that I should allocate that space in memory without having to use SMACK memory model.

Unfortunately the same is not true for structures. For example, the following code
```
struct {
int x;
int y;
} data;
```
would result in a constant 
```
const data: ref;
axiom (data == $sub.ref(0, 1032));
```
SMACK memory model would take care of allocating enough memory for `x` and `y`, but since I'm not using SMACKs memory model, I'm having problems with examples having global structures.

I was wondering if it would be possible to add the `{:count X}` tag (actually any kind of tag would work) for variables coming from global structures. I assume that this should not have any impact in SMACK analysis, but I cannot tell for sure.

shaobo-he:
Fixes #628

### DIFF
--- a/include/smack/SmackRep.h
+++ b/include/smack/SmackRep.h
@@ -22,6 +22,7 @@ class ProcDecl;
 class Stmt;
 class Expr;
 class Regions;
+class Attr;
 
 using llvm::Regex;
 using llvm::SmallVector;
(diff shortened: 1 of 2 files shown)


zvonimir (approved):
Looks good to me. Sorry about the delay!

PR #741, opened 2021-06-15
Generate precise error type when Boogie is used

shaobo-he:
Previously, SMACK doesn't report the error type such as ``invalid pointer
dereference'' when Boogie is used. This commit fixes it. Furthermore, the
original error type generation for Corral is also improved.

### DIFF
--- a/share/smack/top.py
+++ b/share/smack/top.py
@@ -853,18 +853,36 @@ def verification_result(verifier_output):
         return VResult.VERIFIED
     elif re.search((r'\d* verified, [1-9]\d* errors?|can fail|'
                     r'ERRORS_NO_TIMEOUT'), verifier_output):
-        for p in (VProperty.mem_safe_subprops() + [VProperty.INTEGER_OVERFLOW]
-                  + [VProperty.RUST_PANICS]):
-            if re.search(r'ASSERTION FAILS assert {:%s}' % p.boogie_attr(),
-                         verifier_output):
-                return p.result()
+        attr = None
+        attr_pat = r'assert {:(.+)}'
 
-        listCall = re.findall(r'\(CALL .+\)', verifier_output)
-        if len(listCall) > 0 and re.search(
-                r'free_', listCall[len(listCall) - 1]):
-            return VResult.INVALID_FREE
-        else:
-            return VResult.ASSERTION_FAILURE
+        corral_af_msg = re.search(r'ASSERTION FAILS %s' % attr_pat,
+                                  verifier_output)
+        if corral_af_msg:
+            attr = corral_af_msg.group(1)
+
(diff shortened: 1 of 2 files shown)


keram88 (approved):

share/smack/top.py
```diff
@@ -853,18 +853,36 @@ def verification_result(verifier_output):
...
-                r'free_', listCall[len(listCall) - 1]):
-            return VResult.INVALID_FREE
-        else:
-            return VResult.ASSERTION_FAILURE
+        attr = None
+        attr_pat = r'assert {:(.+)}'
+
+        corral_af_msg = re.search(r'ASSERTION FAILS %s' % attr_pat,
+                                  verifier_output)
+        if corral_af_msg:
```
keram88, line 861: Could we guard this with something like `if args.verifier == 'corral'` to ensure we are picking up the expected text? I'm envisioning something like having an if/elif/else on the verifiers and let the else branch emit a warning about a new verifier not having an explicit pattern for its message.

zvonimir: Yep, I like this comment too. I think this would be cleaner than figuring out which verifier is used based on matching of their outputs. Other than that, this looks good!

shaobo-he: I was thinking about it and realized that sometimes it was unclear which verifier is used (e.g., when the verifier is svcomp).

keram88: Perhaps we should make svcomp its own if-branch. We can keep the regex patterns outside the if/elif/else blocks and use them as needed.
I expect if a new verifier to be integrated before SMACK is considered for svcomp.

zvonimir: I agree with @keram88. When I am adding a new verifier, it is great to be able to easily see which parts of the code (Python scripts) I have to touch. Having such an if condition would help and I think make the code easier to understand.

shaobo-he: Thank you, guys. I made some changes, please take a look at them.

zvonimir (approved):
Looks good to me.

keram88 (approved):

share/smack/top.py
zvonimir: Looks good.

share/smack/top.py
```diff
@@ -843,7 +843,7 @@ def transform_out(args, old):
     return out
 
 
-def verification_result(verifier_output):
+def verification_result(verifier_output, verifier):
```
keram88, line 846: Is this new argument not used? It looks like `args.verifier` is used in the body.

shaobo-he: Good catch.

keram88: Looks good to me.

PR #752, opened 2021-08-13
Rewrite memcpy of structures/arrays into load-then-store sequences

### LINKED ISSUE #749: `memcpy` Merges Regions Unnecessarily
Consider the following program,

```
typedef struct {
  int* x;
  int y;
} ST;

int main(void) {
  ST a, b;
  b.x = (int*)malloc(sizeof(int));
  *b.x = 1;
  memcpy(&a, &b, sizeof(ST));
  //a.x = b.x;
  //a.y = b.y;
  return *a.x;
}
```

`a` and `b` point to the same node. Converting them into assignments avoids this effect. This is because sea-dsa unifies the nodes pointed by `memcpy`'s source and destination arguments if the element types contain pointers.

SeaHorn has a `memcpy` rewritten pass (https://github.com/seahorn/seahorn/blob/dev10/lib/Transforms/Scalar/PromoteMemcpy.cc), which we can borrow to avoid this issue.

shaobo-he:
Fixes #749

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,6 +123,7 @@ add_library(smackTranslator STATIC
   include/smack/SplitAggregateValue.h
   include/smack/Prelude.h
   include/smack/SmackWarnings.h
+  include/smack/PromoteMemcpy.h
   lib/smack/AddTiming.cpp
   lib/smack/BoogieAst.cpp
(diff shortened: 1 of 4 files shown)


zvonimir (changes requested):
Would it be possible to add a simple regression that checks that the pass indeed works?

PR #753, opened 2021-08-17
Externalize entry points that are marked internal

keram88:
This adds a utility `extern-statics` that transforms input LLVM files to have functions which are entry-points to always be externally linkable. SMACK's front-end is changed to process input bit-codes before running `llvm-link` on them in order to keep these functions from being erased during linking.

The usage of `extern-statics` is:
* -in <input_file>
* -out <output_file>
* -entry-points <list,of,procedures>


shaobo-he (approved):
LGTM.

PR #758, opened 2021-10-05
Initial draft of Corral threading

jmcmillan1:
Right now a command line call like:
`smack --verifier=portfolio ./file.c`
should run 4 threads of corral verification on that "file.c" each with a combination of the args (eager_threshold=100, smt.arith.solver=2)

It is likely multiple print-outs will show verification results as that has not been adapted yet, but the first terminating result is held and described


zvonimir:
@jmcmillan1 I just realized you created your pull request on the master branch, and not the develop branch. No big deal, no way you could have known that all development happens in the develop branch. You can read here about our branching model: https://nvie.com/posts/a-successful-git-branching-model/

Anyhow, please close this pull request a open a new one on top of the develop branch.
Oh, and now `verify_bpl` returns a value. So that is fixed too.

PR #761, opened 2021-10-08
Cleans up JSON output generation

### LINKED ISSUE #759: Cleaning up smackd stuff
It seems that initially many years ago `smackd` option (and `smackdOutput` function) were implemented to generate json output for the Eclipse SMACK plugin (that we named smackd). However, this plugin has not been maintained in years, and there is no point in keeping smackd functionality around.
Having said that, after I explored this for a bit, I noticed that `smackdOutput` was being used by the SVCOMP script. Why is that so?
I am trying to start a discussion on how to properly clean this up.

zvonimir:
We deprecated smackd output since it has not been used for years.
Renamed it as json output. Also, fixed a bug related to json
output generation.

Closes #759

### DIFF
--- a/share/smack/errtrace.py
+++ b/share/smack/errtrace.py
@@ -88,7 +88,7 @@ def error_trace(verifier_output, verifier):
     return output
 
 
-def smackdOutput(result, output, verifier):
+def json_output_str(result, output, verifier):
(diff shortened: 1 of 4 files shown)


shaobo-he (approved):
LGTM.

PR #762, opened 2021-10-09
Portfolio solving draft

jmcmillan1:
Right now a command line call like:
smack --verifier=portfolio ./file.c
should run 4 threads of corral verification on that "file.c" each with a combination of the args (eager_threshold=100, smt.arith.solver=2)

It is likely multiple print-outs will show verification results for processes which terminate close together in time, but the first terminating result is held and described


zvonimir:
Ok, so something is weird with this pull request since it is trying to update the sea-dsa submodule of SMACK. My hunch is that you did not update submodules once you switch to the develop branch, and you added you commits to the develop branch with the old submodule. Could you please take a look at this?
@keram88 can probably help you out here.

jmcmillan1:
You are right, the sea-dsa submodule was not updated after bringing in develop's changes, should be updated now

shaobo-he:
Is this PR ready to be reviewed? I saw that it's still a draft.

shaobo-he:
One tiny suggestion: we may want the parallel runs to be configurable. Namely, the user can supply a configuration file that specifies the options for each run. Currently, they are hardcoded.

The configuration file can be, e.g., a YAML file, like the following,

```yaml
- verifier: corral
  options: /bopt:proverOpt:O:smt.qi.eager_threshold=100

- verifier: boogie
  options: /proverOpt:O:smt.arith.solver=2
```

Our portfolio solver can read this file and run each configuration in the list in parallel.

By default, we can generate a simple file that just contains verifier boogie and corral without any options.

share/smack/top.py
```diff
@@ -83,7 +86,7 @@ def message(self, args):
         if self is VResult.VERIFIED:
             return ('SMACK found no errors'
                     + ('' if args.modular else ' with unroll bound %s'
-                        % args.unroll) + '.')
+                                               % args.unroll) + '.')
```
zvonimir: The added space here is weird. Am I missing something?

share/smack/top.py
```diff
@@ -528,15 +532,24 @@ def arguments():
         choices=[
             'boogie',
             'corral',
+            'portfolio',
             'symbooglix',
             'svcomp'],
         default='corral',
-        help='back-end verification engine')
+        help='back-end verification engine')  # portfolio added as an option
```
zvonimir: Remove the comment.

share/smack/top.py
```diff
@@ -846,8 +859,8 @@ def transform_out(args, old):
 
 def verification_result(verifier_output, verifier):
     if re.search(
-        r'[1-9]\d* time out|Z3 ran out of resources|timed out|ERRORS_TIMEOUT',
-            verifier_output):
+            r'[1-9]\d* time out|Z3 ran out of resources|timed out|'
```
zvonimir: I am again confused by added space. Is that what the formatter required?

share/smack/top.py
```diff
@@ -891,16 +904,82 @@ def verification_result(verifier_output, verifier):
...
 
+    # ugly way of adding additional params to the verifier options
+    if isinstance(args, tuple):
+        args, commands_to_add = args
+        args.verifier = commands_to_add['verifier']
+        args.verifier_options = commands_to_add['verifier-options']
+
+    if args.verifier == 'portfolio':
+        if args.portfolio_config is None:
+            yaml_info = {'thread1':
```
zvonimir: Let's rename this into `portfolio_config`.

share/smack/top.py
```diff
@@ -891,16 +904,82 @@ def verification_result(verifier_output, verifier):
...
     """Verify the Boogie source file with a back-end verifier."""
 
+    # ugly way of adding additional params to the verifier options
+    if isinstance(args, tuple):
+        args, commands_to_add = args
+        args.verifier = commands_to_add['verifier']
+        args.verifier_options = commands_to_add['verifier-options']
+
+    if args.verifier == 'portfolio':
+        if args.portfolio_config is None:
```
zvonimir: I think there is way to rename this into `portfolio_config_file` without renaming the actual command line option.

share/smack/top.py
```diff
@@ -891,16 +904,82 @@ def verification_result(verifier_output, verifier):
...
+                             'verifier-options':
+                                 "/bopt:proverOpt:O:smt.qi.eager_threshold=100"},
+                         'thread4':
+                            {'verifier': 'corral',
+                             'verifier-options':
+                                 ("/bopt:proverOpt:O:smt.qi.eager_threshold=100"
+                                  " /bopt:proverOpt:O:smt.arith.solver=2")}}
+        else:
+            yaml_info = yaml.load(open(args.portfolio_config, 'r'))
+        p = multiprocessing.Pool()
```
zvonimir: Add new line before this line.

share/smack/top.py
```diff
@@ -891,16 +904,82 @@ def verification_result(verifier_output, verifier):
...
+        thread_args = copy.deepcopy(args)
+
+        results = {}  # map of result -> thread name
+        for thread in list(yaml_info.keys()):
```
zvonimir: Rename `thread` into `thread_name`

share/smack/top.py
```diff
@@ -891,16 +904,82 @@ def verification_result(verifier_output, verifier):
...
+                            {'verifier': 'corral',
+                             'verifier-options':
+                                 ("/bopt:proverOpt:O:smt.qi.eager_threshold=100"
+                                  " /bopt:proverOpt:O:smt.arith.solver=2")}}
+        else:
+            yaml_info = yaml.load(open(args.portfolio_config, 'r'))
+        p = multiprocessing.Pool()
+        thread_args = copy.deepcopy(args)
+
+        results = {}  # map of result -> thread name
```
zvonimir: Wait. I find this weird. Why are you mapping result to thread, and not the other way around - thread to result?

share/smack/top.py
zvonimir: Also, do thread names from yaml have to be unique? Or can you have two threads with the same name?

share/smack/top.py
```diff
@@ -891,16 +904,82 @@ def verification_result(verifier_output, verifier):
...
+                    return term
+
     if args.verifier == 'svcomp':
         verify_bpl_svcomp(args)
         return
 
     elif args.verifier == 'boogie' or args.modular:
         command = ["boogie"]
         command += [args.bpl_file]
+        #  command += ["/nologo", "/doModSetAnalysis"]
```
zvonimir: Why is this commented out?

zvonimir (changes requested):

share/smack/top.py
jmcmillan1: Currently if two threads have the same name, then the first occurrence will be overwritten by the next one  just by the default yaml.read() reading into a dictionary

share/smack/top.py
jmcmillan1: The item that I am checking on to see if "is done" is the AsyncResult object (result). Once I get the finished result I can then lookup its name by using the object itself as the key. I could flip this if that is more intuitive, it will just mean I will need to do a lookup in the dictionary every time I want to check "is done?".

share/smack/top.py
jmcmillan1: I think a leftover from an old version of SMACK awhile ago

share/smack/top.py
```diff
@@ -537,6 +542,14 @@ def arguments():
                                 choices=['z3', 'cvc4', "yices2"], default='z3',
                                 help='back-end SMT solver')
 
+    verifier_group.add_argument(
+        '--portfolio-config',
+        metavar='FILE',
+        default=None,
+        action=FileAction,
+        type=str,
+        help='read portfolio thread specs from a .yaml')
```
zvonimir: Change this help string to `read portfolio configuration in YAML format from FILE`

share/smack/top.py
```diff
@@ -846,8 +859,8 @@ def transform_out(args, old):
 
 def verification_result(verifier_output, verifier):
     if re.search(
-        r'[1-9]\d* time out|Z3 ran out of resources|timed out|ERRORS_TIMEOUT',
-            verifier_output):
+            r'[1-9]\d* time out|Z3 ran out of resources|timed out|ERRORS_TIMEOUT',
```
zvonimir: Why did you introduce this additional indentation here? Was that required for formatting to go through?

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
+                            'thread4':
+                               {'verifier': 'corral',
+                                'verifier-options':
+                                   ("/bopt:proverOpt:O:smt.qi.eager_threshold=100"
+                                    " /bopt:proverOpt:O:smt.arith.solver=2")}}
+    else:
+        portfolio_config = yaml.load(open(args.portfolio_config, 'r'))
+
+    p = multiprocessing.Pool()
+    thread_args = copy.deepcopy(args)
```
zvonimir: How come you don't have to create a deep copy of args for each thread?
Aren't `thread_args` going to get messed up if you keep overwriting them for each thread?

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
-        command = ["corral"]
-        command += [args.bpl_file]
-        command += ["/tryCTrace", "/noTraceOnDisk", "/printDataValues:1"]
-        command += ["/k:%d" % args.context_bound]
-        command += ["/useProverEvaluate"]
-        command += ["/timeLimit:%s" % args.time_limit]
-        command += ["/cex:%s" % args.max_violations]
-        command += ["/maxStaticLoopBound:%d" % args.loop_limit]
-        command += ["/recursionBound:%d" % args.unroll]
-        command += ["/bopt:proverOpt:O:smt.qi.eager_threshold=100"]
```
zvonimir, line 928: You lost this line!

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
-        command += ["/bopt:proverOpt:O:smt.arith.solver=2"]
```
zvonimir, line 929: You lost this line!

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
-    result = verification_result(verifier_output, args.verifier)
+    return process_verifier_output(args, verifier_output)
 
-    if args.json_file:
-        with open(args.json_file, 'w') as f:
-            f.write(json_output_str(result, verifier_output, args.verifier))
 
-    if result in VResult.ERROR:
-        error = error_trace(verifier_output, args.verifier)
+def get_result(result):
```
zvonimir: What is the purpose of this callback function you introduced?
It seems to me that it doesn't really do anything, but I must be missing something here.

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
+    return result
 
-        if args.error_file:
-            with open(args.error_file, 'w') as f:
-                f.write(error)
 
-        if not args.quiet:
-            print(error)
+def thread_verify_bpl(args, commands_to_add):
+    args.modular = False
```
zvonimir: Why is this hard-coded here?

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
+def get_result(result):
+    return result
 
-        if args.error_file:
-            with open(args.error_file, 'w') as f:
-                f.write(error)
 
-        if not args.quiet:
-            print(error)
+def thread_verify_bpl(args, commands_to_add):
```
zvonimir: Rename `commands_to_add` into `args_to_add`.

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
+    args.modular = False
 
-        if args.replay:
-            replay_error_trace(verifier_output, args)
-    print(result.message(args))
-    return result.return_code()
+    if 'verifier' in commands_to_add:
+        args.verifier = commands_to_add['verifier']
+    else:
+        args.verifier = 'corral'
```
zvonimir: Why are you selecting corral here?
I think if verifier is not in commands_to_add, we should just leave whichever verifier was selected.

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
+
+    if 'modular' in commands_to_add:
+        args.modular = commands_to_add['modular']
+
+    if 'verifier-options' in commands_to_add:
+        args.verifier_options = commands_to_add['verifier-options']
+
+    if 'solver' in commands_to_add:
```
zvonimir: I think that for all of these we should issue a warning saying something like: `WARNING: SMACK is overwriting command line argument blah with the one from the choses portfolio configuration`

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
+    elif args.verifier == 'corral':
+        command = invoke_corral(args)
+
+    elif args.verifier == 'symbooglix':
+        command = invoke_symbooglix(args)
+
+    if args.verifier_options:
+        command += args.verifier_options.split()
+
+    return args, try_command(command, timeout=args.time_limit)
```
zvonimir: This will work, but it is weird to be invoking `try command` from the return statement. So maybe just return `command` here and then run it on the other side.

zvonimir: Sorry, I just noticed that you have to run try_command here. So then run it in a separate statement just before the return statement.

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
+                              callback=get_result)] = thread
+
+    while True:  # sleep?
+        for result in list(results.keys()):
+            if result.ready():
+                p.terminate()
+                args, verifier_output = result.get()
+                verifier_output = process_verifier_output(args, verifier_output)
+
+                codes = {0 : 'verified',
```
zvonimir: Why do you need these `codes`? Is this just for debugging?

share/smack/top.py
```diff
@@ -899,85 +979,133 @@ def verify_bpl(args):
...
+    p = multiprocessing.Pool()
+    thread_args = copy.deepcopy(args)
+    results = {}  # map of process -> thread name
+
+    for thread in list(portfolio_config.keys()):
+        results[p.apply_async(thread_verify_bpl,
+                              args=(thread_args, portfolio_config[thread]),
+                              callback=get_result)] = thread
+
+    while True:  # sleep?
```
zvonimir: Yeah, there is probably a better way with sleeping, but this is good for now.

zvonimir (changes requested):

share/smack/top.py
jmcmillan1: This was just to interpret the return code of trying a command with smack, but if the singular printout from process_verifier_output() is sufficient then I can remove these codes and the additional SMACK PORTFOLIO printout

share/smack/top.py
jmcmillan1: If verifier is not provided in commands_to_add then the selected verifier will be 'portfolio' from the initial call. I chose corral to be the default if nothing specific was provided, I figured that is how normal SMACK behaves with no provided --verifier=...

share/smack/top.py
jmcmillan1: I had the same thought but it did not happen. I will just change this though because it will be easier to understand going forwards if everyone that looks at it has that thought

share/smack/top.py
jmcmillan1: Let me know if I should delete these codes. Then the expected output would become something like
`SMACK found no errors with unroll bound 1.
SMACK PORTFOLIO: thread1 returned result code 0`
rather than
`SMACK found no errors with unroll bound 1.
SMACK PORTFOLIO: thread1 returned result code 0, corresponding to: verified`

share/smack/top.py
jmcmillan1: This disables modular from the command line, if modular is then included in the portfolio_config it has the same behavior as SMACK which is to default to that modular over whatever verifier might be provided

share/smack/top.py
zvonimir: I think we can remove them since SMACK printout will already capture the status. We should say though which thread returned first. No need to add result code. So just something like "SMACK portfolio thread1 terminated".

share/smack/top.py
zvonimir: Ah, got it. To more I think about it, let's make `verifier` a required option in the portfolio config file. I think that silently setting it will cause more confusion. So we return an error and exit if verifier is not set in the portfolio file.

share/smack/top.py
zvonimir: Hmmm, that's weird. I wonder if that is because of how the configurations that you tested with looked like. In many scenarios you would not notice this problem since overwriting will still do the right thing.
Otherwise, I have no reasonable explanation for why the current implementation would work properly.

share/smack/top.py
zvonimir: I see. I am (again) worried about silently overwriting things. I think the default behavior should be as follows.
If something is present in the portfolio file setting for a particular thread, and *not* present in the command line arguments, use the option from the portfolio file.
If something is *not* present in the portfolio file setting for a particular thread, and is present in the command line arguments, use the option from the command line arguments.
If something is present in the portfolio file setting for a particular thread, and is also present in the command line arguments, report an error that it is set in both places and exit.
To me this seems like a reasonable logic to go with.
Thoughts?

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
 
-    if result in VResult.ERROR:
-        error = error_trace(verifier_output, args.verifier)
+def thread_verify_bpl(args, args_to_add):
 
-        if args.error_file:
-            with open(args.error_file, 'w') as f:
-                f.write(error)
+    if 'verifier' in args_to_add:
+        if args_to_add['verifier'] is 'portfolio':
```
zvonimir: this should be `if args_to_add['verifier'] == 'portfolio':`

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
+            raise RuntimeError("portfolio is not a valid verifier"
+                               " specification within the portfolio configuration")
```
zvonimir: This line is longer than 80 characters. Make it shorter.

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
+    if 'modular' in args_to_add:
+        if args.modular:
+            raise RuntimeError("argument modular specified in both command"
+                               " line and portfolio configuration")
+        else:
+            print("Warning: SMACK is using argument modular from"
+                  " the chosen portfolio configuration")
+            args.modular = args_to_add['modular']
+
+    if (args.verifier is not 'boogie') and args.modular:
```
zvonimir: This should be `args.verifier != 'boogie'`

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
+        raise RuntimeError("Incompatible arguments modular"
+                           " and non-boogie verifier were specified")
+
+    if 'verifier-options' in args_to_add:
+        if args.verifier_options is not '':
```
zvonimir: This should be `if args.verifier_options:`

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
+            raise RuntimeError("argument verifier-options specified in both"
+                               " command line and portfolio configuration")
+        else:
+            print("Warning: SMACK is using argument verifier-options from"
+                  " the chosen portfolio configuration")
+            args.verifier_options = args_to_add['verifier-options']
+
+    if 'solver' in args_to_add:
+        if args.solver is not 'z3': # is there a better check than not default?
```
zvonimir: This should be `if args.solver != 'z3':`

zvonimir: Also, there should be two space before your comment. So '#' should be preceded by two spaces.

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
+                               {'verifier': 'corral',
+                                'verifier-options': ""},
+                            'thread2':
+                               {'verifier': 'corral',
+                                'verifier-options':
+                                    "/bopt:proverOpt:O:smt.arith.solver=2"},
+                            'thread3':
+                               {'verifier': 'corral',
+                                'verifier-options':
+                                    "/bopt:proverOpt:O:smt.qi.eager_threshold=100"},
```
zvonimir: Line over 80 characters. Break it or reformat somehow.

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
+                            'thread4':
+                               {'verifier': 'corral',
+                                'verifier-options':
+                                   ("/bopt:proverOpt:O:smt.qi.eager_threshold=100"
```
zvonimir: Line over 80 characters. Break it or reformat somehow.

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
+                              args=(copy.deepcopy(args),
+                              portfolio_config[thread]))] = thread
+
+    #TODO: revisit this loop to improve efficiency
+    while True:
+        for result in list(results.keys()):
+            if result.ready():
+                p.terminate()
+                args, verifier_output = result.get()
+                verifier_output = process_verifier_output(args, verifier_output)
```
zvonimir: Line over 80 characters. Break it or reformat somehow.

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
+
+    p = multiprocessing.Pool()
+    results = {}  # map of process -> thread name
+
+    for thread in list(portfolio_config.keys()):
+        results[p.apply_async(thread_verify_bpl,
+                              args=(copy.deepcopy(args),
+                              portfolio_config[thread]))] = thread
+
+    #TODO: revisit this loop to improve efficiency
```
zvonimir: You need space after `#`.

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
+                                   ("/bopt:proverOpt:O:smt.qi.eager_threshold=100"
+                                    " /bopt:proverOpt:O:smt.arith.solver=2")}}
+    else:
+        portfolio_config = yaml.load(open(args.portfolio_config, 'r'))
+
+    p = multiprocessing.Pool()
+    results = {}  # map of process -> thread name
+
+    for thread in list(portfolio_config.keys()):
+        results[p.apply_async(thread_verify_bpl,
```
zvonimir: This has to be reformatted. Do the following:
```
async_result = p.apply_async(...)
results[async_result] = thread
```

share/smack/top.py
```diff
@@ -899,85 +978,145 @@ def verify_bpl(args):
...
+    if args.verifier_options:
+        command += args.verifier_options.split()
+
+    verifier_output = try_command(command, timeout=args.time_limit)
+    return args, verifier_output
+
+
+def verify_bpl_portfolio(args):
+    if args.portfolio_config is None:
+        portfolio_config = {'thread1':
```
zvonimir: I actually think now we should be reading this in from a file. Call it `default-portfolio.yaml` and put it in the same folder where this script is.
Note that you have to add the pattern "*.yaml" here: https://github.com/smackers/smack/blob/main/CMakeLists.txt#L222

zvonimir (changes requested):

zvonimir (approved):
There is still some simple formatting stuff that is failing. I will approve this and merge it, and fix it later on.

PR #763, opened 2021-10-18
Remove calls to Rust's panic functions

keram88:
This removes the calls to panic functions called by Rust programs. Removing the calls enables dead code elimination to remove the arguments flowing into the panic function call.

### DIFF
--- a/lib/smack/RustFixes.cpp
+++ b/lib/smack/RustFixes.cpp
@@ -27,7 +27,9 @@ bool isRustNameMatch(StringRef search, StringRef name) {
   return hashed_match || exact_match;
 }
 
-bool replaceRustMemoryFunctions(Function &f) {
+bool replaceSpecialRustFunctions(Function &f) {
+  std::vector<Instruction *> to_remove;
+
   bool changed = false;
   static const std::map<StringRef, StringRef> alloc_fns = {
       {"_ZN5alloc5alloc5alloc17h", "__smack_rust_std_alloc"},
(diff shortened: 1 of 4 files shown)


shaobo-he:
This PR looks good to me but I was wondering if there's any approach that works on the Rust level. For example, can we use macros or set up empty hooks?

zvonimir (approved):
Looks good to me.

keram88:
> 
> 
> This PR looks good to me but I was wondering if there's any approach that works on the Rust level. For example, can we use macros or set up empty hooks?

We can use macros, but this is meant to deal with panics arising within the standard library.

shaobo-he:
> > This PR looks good to me but I was wondering if there's any approach that works on the Rust level. For example, can we use macros or set up empty hooks?
> 
> We can use macros, but this is meant to deal with panics arising within the standard library.

Ok. What about empty hooks? For example: https://gist.github.com/rust-play/2dcc399b5af55b7657221bec2442b605

lib/smack/RustFixes.cpp
```diff
@@ -64,6 +64,7 @@ bool replaceSpecialRustFunctions(Function &f) {
           FunctionCallee marker = m->getOrInsertFunction(
               Naming::RUST_PANIC_MARKER, Type::getVoidTy(m->getContext()));
           CallInst *panic_marker = CallInst::Create(marker);
+	  panic_marker->setDebugLoc(ci->getDebugLoc());
```
zvonimir: Please fix formatting here.

shaobo-he (approved):
LGTM.

PR #764, opened 2021-10-19
Added unroll folder to regressions

zvonimir:
Also, we should only use Corral on the regressions in the unroll folder.
The reason is that Boogie's unrolling of nested loops is not intuitive.
For example, nested loops of 5 iterations each need 25 or so unrolls
in Boogie to be fully unrolled.


shaobo-he (approved):
LGTM

PR #770, opened 2021-11-30
update file reading to yaml.safe_load

jmcmillan1:
Now use safe_load() instead of load() for reading yaml files as is consistent with PyYaml 6.0


zvonimir (approved):
Looks good to me.

PR #772, opened 2021-12-10
Blocklist for Rust Functions

keram88:
This names some functions that cause crashes in SMACK. Because these functions deal with errors that will be caught by another check in SMACK, we choose to remove the bodies of these functions.


shaobo-he (approved):
This PR looks good to me. Just one minor suggestion, could you add a comment that explains why these functions are removed? Like what crashes they lead to.

zvonimir (approved):
Looks good to me.

PR #778, opened 2022-02-16
Removed unnecessary compiler selection in `build.rs`

shaobo-he:
It shouldn't matter what the C compiler that the `cc` crate uses. So, I removed this function call such that we don't see cargo build failure in an environment such as our Docker image, where `clang` is not in the path.


zvonimir (approved):
Looks good to me.

PR #779, opened 2022-02-28
Fix svcomp assert

zvonimir:
Moving dealing with SVCOMP asserts into the main flow so that we can experiment with SVCOMP benchmarks without having to set verifier to SVCOMP.


shaobo-he:
LGTM.

PR #782, opened 2022-03-01
Set Rust's panic behavior as aborting instead of unwinding

shaobo-he:
Stack unwinding logic is very complicated and even though the
functions associated with it are not called, they stay in the
program. This commit lets Rust programs to abort when panics
occur, which produces much simpler programs.


keram88 (approved):
Looks good to me.

zvonimir (approved):
Looks good to me.

PR #797, opened 2023-03-23
Updated build script to support fedora/centos

rakamaric:


shaobo-he (approved):
LGTM.

bin/build.sh
```diff
@@ -204,6 +206,13 @@ linux-@(ubuntu|neon)-@(16|18|20)*)
   fi
   ;;
 
+linux---x86_64)
+  BUILD_LLVM=1
+  BUILD_Z3=1
```
shaobo-he, line 210: I think the Z3 binary provided by its GitHub repo releases is general enough to run on AL2.

bin/build.sh
```diff
@@ -277,6 +286,27 @@ if [ ${INSTALL_DEPENDENCIES} -eq 1 ] ; then
     sudo apt-get install -y ${DEPENDENCIES}
     ;;
 
+
+  linux---x86_64)
+    sudo yum -y install ninja-build
+    sudo rpm -U https://packages.microsoft.com/config/centos/7/packages-microsoft-prod.rpm
+    sudo yum -y install dotnet-sdk-5.0
+    sudo pip3 install pyyaml psutil toml
```
shaobo-he, line 293: We can do `yum install python3-yaml` stuff like we do `apt-get install python3-yaml` on Debian OSes.

bin/build.sh
```diff
@@ -277,6 +286,27 @@ if [ ${INSTALL_DEPENDENCIES} -eq 1 ] ; then
...
+
+    mkdir -p ${DEPS_DIR}
+    cd ${DEPS_DIR}
+    ${WGET} https://github.com/Kitware/CMake/releases/download/v3.26.0/cmake-3.26.0-linux-x86_64.sh
```
shaobo-he, line 297: `cmake` can be installed via `yum`.

rakamaric: Yes, but not good enough version.

bin/build.sh
rakamaric: I did not manage to do this on the Linux version I was testing this on. In particular, python3-yaml package is not there.

bin/build.sh
rakamaric: I did not manage to get this to work. The version of GMP (or some other dependency, I forgot) that Z3 requires cannot be easily installed on AL2. So I could not get it to work. So I am building it from source so that it builds against the available older version of GMP.

PR #799, opened 2023-04-18
Upgrade CVC4 to CVC5

keram88:
The version of Boogie we use no longer supports CVC4. This pull request addresses that by updating the installer and SMACK to use CVC5 instead.


shaobo-he (approved):
LGTM.

PR #808, opened 2026-03-14
Add integer sign inference analysis pass

shaobo-he:
Introduce SignAnalysis, a module pass that infers the sign (signed/unsigned) of integer SSA values via bidirectional dataflow over a four-point lattice. The analysis propagates constraints from sign-carrying operations (sdiv, zext, signed comparisons, nsw/nuw flags), through memory via sea-dsa alias info, and across function boundaries via call/return edges. SmackRep::lit() now queries SignAnalysis to decide constant rendering, falling back to the previous heuristic when the analysis returns Unknown or Conflict.


shaobo-he:
Closed this PR as the analysis has become too complicated.

